PR #1, opened 2019-01-31
Enh/h5dataio to array

oruebel:
## Motivation

np.isfinite was not working correctly for H5DataIO when wrapping a Python List. See https://github.com/NeurodataWithoutBorders/pynwb/issues/807

## Change

To fix the issue I added the ``__array__`` function similar to numpy to the DataIO class. This allows numpy to cast H5DataIO to an array when a list is being wrapped (as well as when a numpy array is being wrapped). 

## How to test the behavior?
```
import numpy as np
from hdmf.data_utils import H5DataIO
data = H5DataIO(np.array([10., 20.]))
np.isfinite(data)  # Already worked because of the forwarding of attributes in DataIO.__getattr__
data = H5DataIO([10., 20.])
np.isfinite(data)  # Used to fail
```

## Checklist

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -474,7 +474,23 @@ class DataIO(with_metaclass(ABCMeta, object)):
         return len(self.__data)
 
     def __getattr__(self, attr):
+        """Delegate attribute lookup to data object"""
         return getattr(self.data, attr)
+
+    def __array__(self):
+        """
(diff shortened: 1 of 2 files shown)


ajtritt (approved):
This seems to be the best approach. It's unfortunate that numpy duplicates lists for these types of comparisons, but there's not much we can do about that :)

PR #15, opened 2019-03-27
add support for uint as datatype

### LINKED ISSUE #254: uint missing from dtypes
https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype

The dtypes table includes `int`, which is equal to `int32`, but no `uint`. Is there a reason for this? Could we add support for `uint`?

bendichter:
fix https://github.com/NeurodataWithoutBorders/nwb-schema/issues/254

---
name: Pull request
title: ''
labels: ''
assignees: ''

---

## Motivation

`int` maps to `int32`, so it stands to reason that `uint` should map to `uint32`.

## How to test the behavior?
testing this is pretty elaborate, as it would require you to create an extension that uses uint, then construct a file, then write it.

## Checklist

### DIFF
--- a/src/hdmf/build/map.py
+++ b/src/hdmf/build/map.py
@@ -366,6 +366,7 @@ class ObjectMapper(with_metaclass(ExtenderMeta, object)):
         "uint32": np.uint32,
         "uint16": np.uint16,
         "uint8": np.uint8,
+        "uint": np.uint32
     }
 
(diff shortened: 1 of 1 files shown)


oruebel (approved):
Assuming tests pass, this looks fine.

bendichter:
@ajtritt can you merge for me?

PR #17, opened 2019-03-28
external links to groups now use relative path

### LINKED ISSUE #16: relative paths for external links
Briefly describe the needed feature as well as the reasoning behind it

I was trying to share two files, one of which has an external link to another, and I realized that this link used the absolute path to second file, not the relative path, so when I moved the two files the link broke, even if the relative position of the files stayed the same (they were in the same directory). This makes it pretty much impossible to share files that have links between them.

I would like to be able to use relative links instead, so that I can move both of the files at once and the link continues to work.

**Additional context**
I am currently using external links to hold subject-specific information. I am working with ECoG data that contains the a cortical surface mesh object. My current solution is to save this mesh object in every session file, but sometimes there are 100+ session files for a single subject and this surface mesh object never changes for a subject, so it would be better to store this as a separate file and link to if from each session file.

## Checklist

bendichter:
---
name: Pull request
title: ''
labels: ''
assignees: ''

---

(I'm submitting as a draft to test coverage of change.)

## Motivation

fix #16 

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -588,7 +588,10 @@ class HDF5IO(HDMFIO):
         if parent.file.filename == target_builder.source:
             link_obj = SoftLink(path)
         elif target_builder.source is not None:
-            link_obj = ExternalLink(target_builder.source, path)
+            target_filename = os.path.abspath(target_builder.source)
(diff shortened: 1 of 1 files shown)


oruebel (approved):
This looks good to me. If the tests pass I think this can be merged.

bendichter:
@oruebel thanks. I think the tests are jammed though

PR #49, opened 2019-05-07
Fixes extension subtypes

### LINKED ISSUE #892: get_class not working for extension subtypes
We would like to use type inheritance within our extensions.  Generating pynwb classes with ```get_class``` is failing for subtypes that inherit from another of our extension types, throwing a RecursionError. In contrast, generating the classes by writing them ourselves works fine.

```python
# ---------------
# Create an extension with an inheritance hierarchy
# ---------------
import pynwb
from pynwb import load_namespaces, register_class, get_class
from pynwb.file import NWBContainer, NWBDataInterface
from pynwb.spec import NWBNamespaceBuilder, NWBGroupSpec, NWBAttributeSpec
from hdmf.utils import docval

name = 'my_extension'
ns_path = name + ".namespace.yaml"
ext_source = name + ".extensions.yaml"

potato = NWBGroupSpec(neurodata_type_def='Potato',
                                   neurodata_type_inc='NWBDataInterface',
                                   doc='a generic potato',
                                   quantity='*')

colorful_potato = NWBGroupSpec(neurodata_type_def='ColorfulPotato',
                                   neurodata_type_inc='Potato',
                                   doc='potato with a color',
                                   attributes=[NWBAttributeSpec(name='color', doc='color of this potato', 
                                                                dtype='text', required=True)])

ns_builder = NWBNamespaceBuilder(name + ' extensions', name)
ns_builder.add_spec(ext_source, potato)
ns_builder.add_spec(ext_source, colorful_potato)
ns_builder.export(ns_path)


# ---------------
# Create the pynwb classes
# ---------------
load_namespaces(ns_path)
Potato = get_class('Potato', 'my_extension')
ColorfulPotato = get_class('ColorfulPotato', 'my_extension')
```

We can instantiate the top-level class.
```python
generic_potato = Potato(name='A good-ol potato')
```

But it fails on the subclass
``` python
red_potato = ColorfulPotato(name='A red potato', color='red')

>> RecursionError: maximum recursion depth exceeded while calling a Python object
```

However, we have no problem if we write the python classes ourselves.
```python
@register_class('Potato', 'my_extension')
class Potato(NWBDataInterface):
    '''A generic potato.'''

    __nwbfields__ = ('name',)
        
    @docval({'name': 'name', 'type': str, 'doc': 'name of this potato'})
    def __init__(self, **kwargs):
        super(Potato, self).__init__(name=kwargs['name'])
        
        
@register_class('ColorfulPotato', 'my_extension')
class ColorfulPotato(Potato):
    '''An colorful potato.'''
    
    __nwbfields__ = ('name', 'color')

    @docval({'name': 'name', 'type': str, 'doc': 'name of this colorful potato'},
            {'name': 'color', 'type': str, 'doc': 'color of this colorful potato'})
    def __init__(self, **kwargs):
        super(ColorfulPotato, self).__init__(name=kwargs['name'])
        self.color=kwargs['color']

generic_potato = Potato(name='A good-ol potato')
red_potato = ColorfulPotato(name='A red potato', color='red')

print(red_potato)

>> Red Potato <class '__main__.ColorfulPotato'>
>> Fields:
>>   color: red
```

### Environment
Please describe your environment according to the following bullet points.
- **Python Executable:** Conda
- **Python Version:** Python 3.6.7
- **Operating System:** macOS
- **Pynwb Version:** 1.0.2
- **hdmf Version:** 1.0.2

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues/) ?
- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?

rly:
Fixes https://github.com/NeurodataWithoutBorders/pynwb/issues/892

See 
https://stackoverflow.com/questions/47397998/python-decorator-that-calls-super-init 
for how to make decorators work on super.init

### DIFF
--- a/src/hdmf/build/map.py
+++ b/src/hdmf/build/map.py
@@ -1404,7 +1404,7 @@ class TypeMap(object):
         @docval(*docval_args)
         def __init__(self, **kwargs):
             pargs, pkwargs = fmt_docval_args(base.__init__, kwargs)
-            super(base, self).__init__(*pargs, **pkwargs)
+            base.__init__(self, *pargs, **pkwargs) # call function like a method
             for f in new_args:
                 setattr(self, f, kwargs.get(f, None))
 
--- a/src/hdmf/build/map.py
+++ b/src/hdmf/build/map.py
@@ -1404,7 +1404,7 @@ class TypeMap(object):
         @docval(*docval_args)
         def __init__(self, **kwargs):
             pargs, pkwargs = fmt_docval_args(base.__init__, kwargs)
-            super(type(self), self).__init__(*pargs, **pkwargs)
+            super(base, self).__init__(*pargs, **pkwargs)
             for f in new_args:
                 setattr(self, f, kwargs.get(f, None))
 
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -264,10 +264,9 @@ def fmt_docval_args(func, kwargs):
     if func_docval:
         for arg in func_docval[__docval_args_loc]:
             val = kwargs_copy.pop(arg['name'], None)
-            if 'default' in arg:
-                if val is not None:
-                    ret_kwargs[arg['name']] = val
-            else:
+            if val is not None:
+                ret_kwargs[arg['name']] = val
+            elif 'default' not in arg:
                 ret_args.append(val)
         if func_docval['allow_extra']:
             ret_kwargs.update(kwargs_copy)


src/hdmf/build/map.py
```diff
@@ -1404,7 +1404,7 @@ def __get_cls_dict(self, base, addl_fields):
         @docval(*docval_args)
         def __init__(self, **kwargs):
             pargs, pkwargs = fmt_docval_args(base.__init__, kwargs)
-            super(type(self), self).__init__(*pargs, **pkwargs)
+            super(base, self).__init__(*pargs, **pkwargs)
```
bendichter:
```suggestion
            base.__init__(*pargs, **pkwargs)
```

rly:
@ajtritt I changed a pretty key function `fmt_docval_args` to return the keyword arg pairs when there is no default. It seems like the function should work this way, but it fails this test in `unit/utils_test/test_core.py`:
```python
class MyTestClass(object):
    @docval({'name': 'arg1', 'type': str, 'doc': 'argument1 is a str'},
            {'name': 'arg2', 'type': 'int', 'doc': 'argument2 is a int'},
            {'name': 'arg3', 'type': bool, 'doc': 'argument3 is a bool. it defaults to False', 'default': False})
    def basic_add2_kw(self, **kwargs):
        return kwargs

class TestDocValidator(unittest.TestCase):

    def setUp(self):
        self.test_obj = MyTestClass()

    def test_fmt_docval_args(self):
        """ Test that fmt_docval_args works """
        test_kwargs = {
            'arg1': 'a string',
            'arg2': 1,
            'arg3': True,
        }
        rec_args, rec_kwargs = fmt_docval_args(self.test_obj.basic_add2_kw, test_kwargs)
        exp_args = ['a string', 1]
        self.assertListEqual(rec_args, exp_args)
```

Since the input `test_kwargs` is a dictionary with the correctly named arguments, I think the test should be that `rec_args == []` and `rec_kwargs == {'arg1': 'a string', 'arg2': 1, 'arg3': True}`.

But maybe I am missing something about the meaning of "default." Does this look right to you? Should the test be changed?

bendichter:
@rly I think the idea was to have docval items with no default be returned as args and the docval items with default to be treated as kwargs. I don't think there's anything inherently wrong with making them all kwargs, but is this necessary for fixing #892?

rly:
I lost my mental debugging stack trace from yesterday, but I think changing how `fmt_docval_args` worked was necessary for assigning values passed in a subclass constructor to required argument fields of a parent class.

bendichter:
@rly I incorporated your `base` change in my PR #51, which seems to fix https://github.com/NeurodataWithoutBorders/pynwb/issues/892 without needing to change docval

bendichter:
@rly can we close this?

rly:
Yes, let's close. I'm looking into whether there is a use case where `fmt_docval_args` breaks, but so far, it seems like the code magically manages it just fine.

PR #51, opened 2019-05-07
fix get_class with supplied name

bendichter:
if name is in spec fields, remove from `docval` and add directly to args in `__init__`

## Motivation

fix #48, fix https://github.com/NeurodataWithoutBorders/pynwb/issues/892

## How to test the behavior?
```python
from pynwb import NWBHDF5IO, NWBFile, load_namespaces, get_class
from datetime import datetime
import numpy as np


nwbfile = NWBFile('aa', 'aa', datetime.now().astimezone())

load_namespaces('/Users/bendichter/dev/to_nwb/to_nwb/ndx_single_dataset/spec/single_dataset.namespace.yaml')

sim_meta_data = get_class('SimulationMetaData', 'single_dataset')

meta_data = sim_meta_data(pin=np.random.randn(1000, 12))
nwbfile.add_lab_meta_data(meta_data)

with NWBHDF5IO('test_a.nwb', 'w') as io:
    io.write(nwbfile, cache_spec=True)


with NWBHDF5IO('test_a.nwb', 'r', load_namespaces=True) as io:
    nwb2 = io.read()
    print(nwb2.lab_meta_data)
```
works and 
```python
meta_data = sim_meta_data(name='bad_name_here', pin=np.random.randn(1000, 12))
``` 
correctly throws an error

## Checklist

- [x] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR description clearly describes problem and the solution?
- [x] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/hdmf-dev/hdmf/pulls) for the same change?
- [x] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ? By including "Fix #XXX" you allow GitHub to close the corresponding issue.

PR #54, opened 2019-05-08
Add tests for #51 and fix

rly:
Inserting name as a positional arg in the first location can sometimes 
mess up other positional args because the docval args no longer list 
name as a positional arg


bendichter:
looks good except I'm not sure if names are allowed to contain spaces

oruebel:
I believe the main restriction with names is that they cannot contain "/" since that is used in paths in HDF5. Spaces I believe should be fine (although in general I would encourage to use "_" or "-" instead of spaces in names)

bendichter (approved):

PR #57, opened 2019-05-08
#handle zero-length lists and tuples in convert_dtype

### LINKED ISSUE #56: ObjectMapper.convert_dtype fails on length 0 lists and tuples
see here:
https://github.com/hdmf-dev/hdmf/blob/d60de334e98cfe88edae0663af3c1c66d7d00ed5/src/hdmf/build/map.py#L439

tmp_dtype is only defined if at least one iteration of the above loop has occurred.

NileGraddis:
Fix #56

### DIFF
--- a/src/hdmf/build/map.py
+++ b/src/hdmf/build/map.py
@@ -431,6 +431,8 @@ class ObjectMapper(with_metaclass(ExtenderMeta, object)):
                 ret = np.asarray(value).astype(dtype_func)
                 ret_dtype = ret.dtype.type
         elif isinstance(value, (tuple, list)):
+            if len(value) == 0:
+                return value, spec['dtype']
             ret = list()
             for elem in value:
                 tmp, tmp_dtype = cls.convert_dtype(spec, elem)
--- a/src/hdmf/build/map.py
+++ b/src/hdmf/build/map.py
@@ -432,7 +432,7 @@ class ObjectMapper(with_metaclass(ExtenderMeta, object)):
                 ret_dtype = ret.dtype.type
         elif isinstance(value, (tuple, list)):
             if len(value) == 0:
-                return value, spec['dtype']
+                return value, spec_dtype
             ret = list()
             for elem in value:
                 tmp, tmp_dtype = cls.convert_dtype(spec, elem)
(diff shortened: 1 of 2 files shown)


NileGraddis:
@ajtritt would you mind taking a look at this? Our allensdk tests are breaking on the latest pynwb/hdmf without these changes.

src/hdmf/build/map.py
```diff
@@ -431,6 +431,8 @@ def convert_dtype(cls, spec, value):
                 ret = np.asarray(value).astype(dtype_func)
                 ret_dtype = ret.dtype.type
         elif isinstance(value, (tuple, list)):
+            if len(value) == 0:
+                return value, spec['dtype']
```
ajtritt: Why is this not `return value, cls.__spec[spec_dtype]` ? This method is meant to return the `numpy.dtype` object. The way this line is written will return a `str`

ajtritt (changes requested):
see comment

src/hdmf/build/map.py
NileGraddis: that makes a good amount of sense. Will fix

NileGraddis: hmm, ObjectMapper seems not to have a `__spec` attribute. It looks like `__resolve_dtype` works on an object and a type spec to produce a numpy dtype. Since this is handling the degenerate case where we only have a spec, does it work for this line to just read:
```Python
return value, np.dtype(spec_dtype)
```
?

NileGraddis: @ajtritt one other thing - when testing pynwb against latest hdmf (with these changes or not) I notice that there are about 5 test failures. Are these known & tracked? What is the method for testing pynwb against the latest hdmf?

ajtritt: My bad, I meant.

```python
return value, spec_dtype
```
the `str` -> `np.dtype` gets handled on 421, no need to do it yourself.

ajtritt: I have not experience the errors in NWB with the latest HDMF, but lets take that conversation elsewhere

NileGraddis:
@ajtritt could you take a look at the changes I've made to this pr?

oruebel (approved):

PR #63, opened 2019-05-16
change cache_spec default value to True

bendichter:
fix https://github.com/NeurodataWithoutBorders/pynwb/pull/710

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -140,7 +140,10 @@ class H5SpecWriter(SpecWriter):
 
     def __write(self, d, name):
         data = self.stringify(d)
-        dset = self.__group.create_dataset(name, data=data, dtype=self.__str_type)
+        if name in self.__group:  # handle append mode
+            dset = self.__group[name]
+        else:
+            dset = self.__group.create_dataset(name, data=data, dtype=self.__str_type)
         return dset
 
     def write_spec(self, spec, path):
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -188,7 +188,7 @@ class HDF5IO(HDMFIO):
         dest_file.close()
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
-            {'name': 'cache_spec', 'type': bool, 'doc': 'cache specification to file', 'default': False},
+            {'name': 'cache_spec', 'type': bool, 'doc': 'cache specification to file', 'default': True},
             {'name': 'link_data', 'type': bool,
              'doc': 'If not specified otherwise link (True) or copy (False) HDF5 Datasets', 'default': True})
     def write(self, **kwargs):


oruebel (dismissed):

t-b:
@bendichter Are the pynwb tests run against this version of hdmf? They were not passing last time I tried that.

bendichter:
@t-b hmm let me check on this

bendichter:
There are two issues we need to resolve before merging.

1) As @t-b pointed out, this currently causes an error in pynwb.
`python test.py -v`
```python
======================================================================
ERROR: test_append (test_io.TestAppend)
----------------------------------------------------------------------
  File "/Users/bendichter/dev/pynwb/tests/integration/test_io.py", line 274, in test_append
    io.write(nwb)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 388, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/hdmf/src/hdmf/backends/hdf5/h5tools.py", line 216, in write
    ns_builder.export('namespace', writer=writer)
  File "/Users/bendichter/dev/hdmf/src/hdmf/utils.py", line 388, in func_call
    return func(self, **parsed['args'])
  File "/Users/bendichter/dev/hdmf/src/hdmf/spec/write.py", line 198, in export
    writer.write_spec(out, path)
  File "/Users/bendichter/dev/hdmf/src/hdmf/backends/hdf5/h5_utils.py", line 147, in write_spec
    return self.__write(spec, path)
  File "/Users/bendichter/dev/hdmf/src/hdmf/backends/hdf5/h5_utils.py", line 143, in __write
    dset = self.__group.create_dataset(name, data=data, dtype=self.__str_type)
  File "/anaconda3/lib/python3.6/site-packages/h5py/_hl/group.py", line 139, in create_dataset
    self[name] = dset
  File "/anaconda3/lib/python3.6/site-packages/h5py/_hl/group.py", line 371, in __setitem__
    h5o.link(obj.id, self.id, name, lcpl=lcpl, lapl=self._lapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5o.pyx", line 202, in h5py.h5o.link
RuntimeError: Unable to create link (name already exists)\
```
This error does not appear on the dev branch and I don't see why this chance would create this error, so some investigation is needed. 

2) Integration tests are now much slower because it needs to cache the spec each time. I think it would be best if we changed the pynwb integration test code to override the default for cache_spec and set it to `False` in the majority of cases, and then turn it on for a few tests where we test `cache_spec` specifically.

bendichter:
OK 1 was handled by checking if a spec has already been written before writing it. 2 was handled by setting the default for integration tests to cache_spec=False: https://github.com/NeurodataWithoutBorders/pynwb/pull/945

bendichter:
for full test coverage, we need to add a test for appending to a file with cache_spec=True in the original and second write

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -140,7 +140,10 @@ def stringify(spec):
 
     def __write(self, d, name):
         data = self.stringify(d)
-        dset = self.__group.create_dataset(name, data=data, dtype=self.__str_type)
+        if name in self.__group:  # handle append mode
```
ajtritt, line 143: This behavior is handled by [require_dataset](http://docs.h5py.org/en/stable/high/group.html?highlight=require_dataset#Group.require_dataset)

ajtritt (changes requested):

rly:
See PR #84 for revised PR

Issue #65, opened 2019-05-21
Overwriting file with object already in file creates broken link and failure to write


rly:
**Bug**

Pseudocode:
- create container C1
- add object O to container C1
- write container C1 to file F
- create container C2
- add same object O to container C2
- write container C2 to file F

This results in a broken link to object O, and file F no longer has object O.

This example uses pynwb but it also applies to custom containers. 

```python
from datetime import datetime
from dateutil.tz import tzlocal
from pynwb import NWBFile, TimeSeries, NWBHDF5IO, get_manager
import numpy as np

start_time = datetime(2017, 4, 3, 11, tzinfo=tzlocal())

nwbfile = NWBFile(session_description='demonstrate NWBFile basics',  # required
                  identifier='NWB123',  # required
                  session_start_time=start_time)  # required

data = list(range(100, 200, 10))
timestamps = list(range(10))
test_ts = TimeSeries(name='test_timeseries1', data=data, unit='m', timestamps=timestamps)
nwbfile.add_acquisition(test_ts)

with NWBHDF5IO('example_file_path.nwb', manager=get_manager(), mode='w') as io:
    io.write(nwbfile)

# do it again
nwbfile = NWBFile(session_description='demonstrate NWBFile basics',  # required
                  identifier='NWB123',  # required
                  session_start_time=start_time)  # required

# add object that already exists and was added to the file
nwbfile.add_acquisition(test_ts)

# add new object
test_ts2 = TimeSeries(name='test_timeseries2', data=data, unit='m', timestamps=timestamps)
nwbfile.add_acquisition(test_ts2)

with NWBHDF5IO('example_file_path.nwb', manager=get_manager(), mode='w') as io:
    io.write(nwbfile)  # broken link warnings

with NWBHDF5IO('example_file_path.nwb', manager=get_manager(), mode='r') as io:
    read_file = io.read()
    print(read_file)  # missing test_timeseries1

```

**Environment**

Please describe your environment according to the following bullet points.

    Python Executable: Conda
    Python Version: Python 3.7
    Operating System: Windows
    HDMF Version: latest dev

rly:
Finally got around to investigating this issue. The core issue is that under this procedure:
1. create container C1
2. add object O to container C1
3. write container C1 to file F
4. create container C2
5. add same object O to container C2
6. write container C2 to file F

[...]

PR #66, opened 2019-05-21
Create IO warnings and unit tests

rly:
## Motivation

Prevent user errors of doing invalid operations under certain file open modes:
1. cannot init io object in r, r+ mode if file doesn't exist
2. cannot init io object in w mode if file exists
3. cannot write to file in r mode
4. cannot read file in w, w- mode
5. cannot read data if there are no values in file

One test is commented out until #65 is fixed.


ajtritt (changes requested):
- Make sure we are consistent about whether or not we accept 'x' mode.
- refactor `FooFile` so it does not take a `name` argument

ajtritt (approved):

Issue #68, opened 2019-05-23
pynwb validation broken since a16cc63e


t-b:
The validation in pynwb is broken since a16cc63e (passively read datasets that are vectors of strings, 2019-05-03), see #46 and https://github.com/NeurodataWithoutBorders/pynwb/issues/911.

```
thomas@thomas-win7-x64 MINGW64 /e/projekte/pynwb (better-validation)
$ python docs/gallery/domain/ophys.py

thomas@thomas-win7-x64 MINGW64 /e/projekte/pynwb (better-validation)
$ python -m pynwb.validate ophys_example.nwb; echo $?
Validating ophys_example.nwb against core namespace
 - found the following errors:
root/file_create_date (file_create_date): incorrect type - expected 'isodatetime', got 'object'
1
```

rly:
This should be working now.

PR #69, opened 2019-05-24
Fixes for getting validation against cached namespace information working

t-b:
These changes here are required for https://github.com/NeurodataWithoutBorders/pynwb/pull/944.
Before merging we need to find a solution for 428ec63, reverting was the only way for me to get validation working.

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -76,21 +76,20 @@ class HDF5IO(HDMFIO):
         '''
         Load cached namespaces from a file.
         '''
-        f = File(path, 'r')
-        if SPEC_LOC_ATTR not in f.attrs:
-            msg = "No cached namespaces found in %s" % path
-            warnings.warn(msg)
-        else:
-            spec_group = f[f.attrs[SPEC_LOC_ATTR]]
-            if namespaces is None:
-                namespaces = list(spec_group.keys())
-            for ns in namespaces:
-                ns_group = spec_group[ns]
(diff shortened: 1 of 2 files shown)


bendichter:
why did we need to revert a16cc63e9065b92673d175fd9e55a86c6eb46d54? What was the issue with the validator that requires you to read all vectors of strings on `read`?

t-b:
@bendichter Because of #68, I don't want to revert a16cc63, I just needed a hack to get validation working.

bendichter:
I see.. so maybe we could allow time vectors to be objects?

t-b:
@bendichter Or just fix the parts in pynwb which reads them. But I have not found these.

bendichter:
The issue is you wouldn't be able to know it's a date until you read the string.

oruebel:
> The issue is you wouldn't be able to know it's a date until you read the string

Do we have the schema accessible at that time? If so, the isodatetime type should be specified there. 

>  I just needed a hack to get validation working.

Could the validator not simply read the string when it validates date times?

ajtritt (dismissed):

t-b:
@ajtritt Thanks for the review. Just to avoid a mess I've removed the reverting commit. Once #74 is merged this PR is from POV good to go.

rly (approved):

rly:
@t-b is this good to merge? if so, please merge. thanks

t-b:
@rly Ready to merge. But I'm not allowed to merge!

rly:
Oops, my bad!

PR #78, opened 2019-05-30
Read HDF5 datasets with SIMPLE dataspaces as well

t-b:
- [ ] Needs to keep arrays with just a single entry as some constructors expect these.


src/hdmf/build/map.py
```diff
@@ -1202,12 +1203,16 @@ def construct(self, **kwargs):
                 val = const_args[argname]
             else:
                 continue
+            # reduce single entry arrays to scalars
+            if isinstance(val, (Dataset, np.ndarray)):
+                if val.shape == (1,):
+                    val = val[0]
```
oruebel: I'm not 100% sure whether just checking for single-length shape is sufficient. I think you may need to look at the docval constructor or the spec to see if an array is expected or if a scalar is fine.

rly: There are places in pynwb where an array is expected and a scalar is not, e.g.:
https://github.com/NeurodataWithoutBorders/pynwb/blob/50f934c9e7dd8e9dccada38500c3790b6aca3dc1/src/pynwb/file.py#L201

`'array_data`' allows only array-like objects. You would have to change that to allow the scalar types as well. 
https://github.com/hdmf-dev/hdmf/blob/8ddc8574123853215a4d6c43d27bd2aec908017d/src/hdmf/utils.py#L13

t-b: @oruebel @rly Thanks both. I'll come up with some docval magic.

t-b: @oruebel @rly I found a different approach which does not require me to understand docval :duck:.

t-b: Ditch that, this is BS.

oruebel:
@ajtritt I think this is good to merge

t-b:
@oruebel Nope, sorry this PR is not good. I'll  revise it once I find time.

t-b:
@rly @oruebel Any hints how some docval magic would look like to find out if the expected argument is a scalar or on array_like type?

oruebel:
> hints how some docval magic would look like to find out if the expected argument is a scalar or on array_like type?

Just to clarify, is the goal to (1) enforce that a parameter is a scalar or (2) to check if a parameter value passed in is a scalar?

For (1) I think you would just list "scalar_data" as the type of the parameter in docval (or int, float, str if you are expecting a particular type of scalar). 

For (2) you could use https://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html

t-b:
@oruebel I need to find out if the argument to the constructor is a scalar or array and precondition the HDF5 dataset according to that. So I *think* I need to query the type of the argument using docval.

ajtritt:
@t-b 
> Any hints how some docval magic would look like to find out if the expected argument is a scalar or on array_like type?

No, docval does not do anything like that. It **DOC**uments and **VAL**idates function arguments. You will have to check inside the function to determine scalar/array-like.

oruebel:
>  I need to find out if the argument to the constructor is a scalar or array

For this, checking the value of the parameter with [np.isscalar ](https://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html) is probably the way to go.

t-b:
Let's step back and have an example ;)

The attached file was acquired with a beta version of NWBv2 support in MIES.

[MIES-26da4118-NWBv2.zip](https://github.com/hdmf-dev/hdmf/files/4078431/MIES-26da4118-NWBv2.zip)

```
$ pip show pynwb hdmf
Name: pynwb
Version: 1.1.2.post0.dev27
Summary: Package for working with Neurodata stored in the NWB format
Home-page: https://github.com/NeurodataWithoutBorders/pynwb
Author: Andrew Tritt
Author-email: ajtritt@lbl.gov
License: BSD
Location: e:\projekte\pynwb\src
Requires: h5py, hdmf, numpy, pandas, python-dateutil
Required-by:
---
Name: hdmf
Version: 1.5.3
Summary: A package for standardizing hierarchical object data
Home-page: https://github.com/hdmf-dev/hdmf
Author: Andrew Tritt
Author-email: ajtritt@lbl.gov
License: BSD
Location: c:\users\thomas\anaconda3\envs\shreejoy-nwb-python\lib\site-packages
Requires: pandas, h5py, numpy, scipy, ruamel.yaml
Required-by: pynwb
```

Reading the file with 

```python
from pynwb import NWBHDF5IO
import numpy as np

with NWBHDF5IO('MIES-26da4118-NWBv2.nwb', 'r') as io:
    nwbfile = io.read()

    for key in nwb.acquisition:
        acquisition = nwb.get_acquisition(key)
        print(acquisition)

    for key in nwb.stimulus:
        stimulus = nwb.get_stimulus(key)
        print(stimulus)
```

gives the following exceptions [output.txt](https://github.com/hdmf-dev/hdmf/files/4078438/output.txt).

Remember this is plain pynwb and hdmf without any patches on top.

Now the question is what is the appropriate fix for this? My approach here was to fix the container creation method. I think that touching `IntracellularElectrode` here is the wrong appraoch as I would need to do that for every class. My feeling is that the fix should be more inside the builder as that should have all the knowledge about hdf5 but not the upper layers.

t-b:
@rly @oruebel I've updated my above post. Any ideas?

rly:
FYI @t-b your test file has a dataset in `/root/source_script/` that is 5x2 when it should be a scalar text field. Even with the upcoming fixes, that file will still not load.

t-b:
> FYI @t-b your test file has a dataset in /root/source_script/ that is 5x2 when it should be a scalar text field. Even with the upcoming fixes, that file will still not load.

@ukos-git Thanks for pointing that out. We'll fix that.

@rly Do you know why the validator did not find that?

rly:
@t-b. I don't think the validator validates shape in any way. We would have to be careful adding that into the validator, because then technically, all files prior to https://github.com/hdmf-dev/hdmf-common-schema/pull/17 with a `DynamicTable` will fail validation, as `VectorData` would be read as a scalar.

rly:
Superseded by #269.

PR #80, opened 2019-06-06
H5SpecReader: Support SIMPLE HDF5 dataspaces as well

t-b:
Some programs like Igor Pro always write SIMPLE dataspaces instead of
SCALAR ones. As we should not care about the exact type, we dig now
deeper if we find an array with just one element.


src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -160,6 +160,11 @@ def __init__(self, **kwargs):
 
     def __read(self, path):
         s = self.__group[path][()]
+
```
ajtritt: @t-b have you run into issues with this?

ajtritt: A better fix might be to make H5SpecWriter.__write always write JSON schema strings as scalars.

t-b:
@rly Is that change here good? Or do you want a complete solution together with #78?

rly:
@t-b I think this is fine, but I'm a little confused. Are you using Igor Pro to write specs? Does Igor Pro not read scalars?

Also, do you mean "scalar dataspace" instead of "simple dataspace"?
>  A scalar dataspace is a special case of the simple data space and is defined to be a 0-dimensional single data point in size.
- https://support.hdfgroup.org/HDF5/doc1.6/Dataspaces.html

rly:
Also, are you using the latest h5py?

> When a scalar dataset was accessed with the syntax dataset[()], h5py incorrectly returned an ndarray. H5py now correctly returns an array scalar. Using dataset[...] on a scalar dataset still returns an ndarray.
- http://docs.h5py.org/en/stable/whatsnew/2.0.html#changes-to-scalar-slicing-code

oruebel:
@ajtritt I think this is ready to merge

ajtritt (approved):
I'm failing to see the connection to Igor Pro and dataspaces (I am sure its there, but some more clarification would be good), but this modification seems okay to me.

PR #82, opened 2019-06-08
Fix #81 - Write lists of strings with their original encoding

### LINKED ISSUE #81: List of strings encoded in ASCII, not Unicode
An attribute that is a list of strings is encoded as ASCII instead of Unicode. All other strings are encoded as Unicode. 

image

rly:


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -509,9 +508,9 @@ def set_attributes(self, **kwargs):
             if isinstance(value, (set, list, tuple)):
                 tmp = tuple(value)
                 if len(tmp) > 0:
-                    # a list of strings will need a special type
+                    # store list of strings in unicode, not ascii
                     if isinstance(tmp[0], (text_type, binary_type)):
-                        value = [np.string_(s) for s in tmp]
+                        value = [np.unicode_(s) for s in tmp]
```
ajtritt, line 546: Just delete this, so the values get written in the format that they are passed in as.

ajtritt (changes requested):

rly:
@ajtritt this should be good to go now

ajtritt (approved):

PR #83, opened 2019-06-19
Update order of keys in YAML spec

### LINKED ISSUE #274: Reorder YAML keys to be consistent and more human-readable
The ordering of keys for many of the types in the YAML files could be improved to be consistent across definitions/files and more human readable. 

For example, if `neurodata_type_def` is present, it should always be first; else, `name` should be first. `dims` should be next to `shape`. `default_value` should be next to `value`. Description of the size/shape/quantity of the dataset should go before any attributes, groups, datasets, or links contained within it. 

See also https://github.com/NeurodataWithoutBorders/nwb-schema/issues/139 where the order: ```['neurodata_type_def', 'neurodata_type_inc', 'name', 'dtype', 'doc']``` was decided on.

I propose this more full ordering below. As was discussed previously, this ordering should not be enforced, but we should try to follow these guidelines when modifying the core and in the script that generates extensions. 

I also propose having a blank line before each new type definition `neurodata_type_def`

```
neurodata_type_def
neurodata_type_inc
name
default_name
dtype
target_type
dims
shape
value
default_value
doc
required
quantity
attributes
datasets
groups
links
```

rly:
Fixes NeurodataWithoutBorders/nwb-schema#274 for writing schema YAML files. See also https://github.com/NeurodataWithoutBorders/pynwb/pull/978

Ordering of keys in YAML files is now:
```
neurodata_type_def
neurodata_type_inc
name
default_name
dtype
target_type
dims
shape
value
default_value
doc
required
quantity
attributes
datasets
groups
links
```
except where `neurodata_type_def` is not used, in which case name goes before `neurodata_type_inc`.

### DIFF
--- a/src/hdmf/spec/write.py
+++ b/src/hdmf/spec/write.py
@@ -70,8 +70,8 @@ class YAMLSpecWriter(SpecWriter):
         yaml.representer.RoundTripRepresenter.add_representer(type(None), my_represent_none)
 
         order = ['neurodata_type_def', 'neurodata_type_inc', 'name', 'default_name',
-             'dtype', 'target_type', 'dims', 'shape', 'default_value', 'value', 'doc',
-             'required', 'quantity', 'attributes', 'datasets', 'groups', 'links']
+            'dtype', 'target_type', 'dims', 'shape', 'default_value', 'value', 'doc',
+            'required', 'quantity', 'attributes', 'datasets', 'groups', 'links']
         if isinstance(obj, dict):
             keys = list(obj.keys())
             for k in order[::-1]:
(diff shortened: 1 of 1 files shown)


src/hdmf/spec/write.py
```diff
@@ -69,16 +69,20 @@ def my_represent_none(self, data):
...
             for k in order[::-1]:
                 if k in keys:
                     keys.remove(k)
                     keys.insert(0, k)
+            if 'neurodata_type_def' not in keys and 'name' in keys:
+                keys.remove('name')
+                keys.insert(0, 'name')
             return yaml.comments.CommentedMap(
-                yaml.compat.ordereddict([(k, self.sort_keys(obj[k])) for k in keys])
+                yaml.compat.ordereddict([(k,self.sort_keys(obj[k])) for k in keys])
```
bendichter:
```suggestion
                yaml.compat.ordereddict([(k, self.sort_keys(obj[k])) for k in keys])
```

bendichter:
otherwise, looks good

ajtritt (approved):

PR #84, opened 2019-06-20
Cache the spec by default when writing file

### LINKED ISSUE #76: Always cache spec
Is there a reason why we would not want to cache the spec when writing a file?

Advantages of always caching spec: fewer complex use cases and interactions to handle and tests to make.

rly:
Fixes #76 

Also, prevent overwriting of spec with same name
e.g. if the group '/specifications/core/2.0.2/' already contains the group 'nwb.base', do not try to create it again when caching the spec.


ajtritt (dismissed):

rly:
I forgot that @bendichter had already made a PR #63 on this. @ajtritt I now use `require_dataset` based on your comment there.

ajtritt (approved):

PR #85, opened 2019-06-21
add bool type to _type_map dict

bendichter:
---
name: Pull request
title: ''
labels: ''
assignees: ''

---

## Motivation

previously get_class would break if given a neurodata_type with an attribute that is of type bool. This fixes is.

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/build/map.py
+++ b/src/hdmf/build/map.py
@@ -1333,6 +1333,7 @@ class TypeMap(object):
         'float64': float,
         'int': int,
         'int32': int,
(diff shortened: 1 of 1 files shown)


rly (approved):

rly:
Looks good, though I am confused as to why we have both `TypeMap.__type_map` and a `ObjectMapper.__dtypes`. They both map from a `spec.dtype` value to something else, so presumably they should have the same keys.

PR #88, opened 2019-06-25
Ignore setting parent when parent adds child with existing parent

### LINKED ISSUE #87: Cannot write external links
With the most recent 1.0.4 version of hdmf, parent-child relationships are enforced such that a child can have only one parent, which cannot be reassigned. This breaks linking of external files. 

```python
from pynwb import get_manager

manager = get_manager()

io = NWBHDF5IO(filename, 'r', manager=manager)
nwbfile1 = io.read()
ts = nwbfile1.get_acquisition('ts')

nwbfile2 = NWBFile(session_description='',
                   identifier='id',
                   session_start_time=start_time)
nwbfile2.add_acquisition(ts)
```

`ts` already has a parent `nwbfile1`. Therefore, `nwbfile2` cannot add `ts` as a child. 

Options:
1. Allow `Container.parent` to be reset. This could lead to strange behavior and bugs down the road (or bugs that return). 
2. Linking is currently handled automagically during `NWBHDF5IO.write`. Instead of determining on write whether a link exists, ask the user to make all links explicit, e.g. `nwbfile3.add_acquisition(NWBLink(ts))`. This is more akin to a filesystem -- you can't have the same file in two different folders, but you can make a link/shortcut to the file and move that link around.

rly:
Hotfix to fix #87. Also fixes the issue for making soft (internal) links.

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -45,12 +45,12 @@ class Container(with_metaclass(ExtenderMeta, object)):
     def add_child(self, **kwargs):
         child = getargs('child', kwargs)
         if child is not None:
-            self.__children.append(child)
-            self.set_modified()
-
             # if child.parent is a Container, then the mismatch between child.parent and parent
             # is used to make a soft/external link from the parent to a child elsewhere
+            # if child.parent is not a Container, it is either None or a Proxy and should be set to self
             if not isinstance(child.parent, Container):
+                self.__children.append(child)
+                self.set_modified()
                 child.parent = self
         else:
             warn('Cannot add None as child to a container %s' % self.name)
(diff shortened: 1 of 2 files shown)


src/hdmf/container.py
```diff
@@ -51,7 +51,11 @@ def add_child(self, **kwargs):
         if child is not None:
             self.__children.append(child)
```
ajtritt: I think you want to move this under the conditional as well

ajtritt: I think you want to move this under the conditional as well

ajtritt (changes requested):

src/hdmf/container.py
rly: This would no longer be true:
```python
parent_obj = Container('obj1')
child_obj = Container('obj2', parent_obj)
self.assertIs(parent_obj.children[0], child_obj)
```
because `child_obj.parent` gets set before `child_obj.parent.add_child(child_obj)` gets called in lines 20-22 above. 
I had added lines 21-22 because I thought the parent should know the objects that call it its parent, but given how the code works, this doesn't seem to be important, so just in case that breaks something as well, I'll revert it.

ajtritt (approved):

PR #89, opened 2019-06-27
Improve unit test runner output to make debugging easier

rly:
This changes `test.py` to output colored, tabular test results when in verbose mode (`python test.py -v`) which should make figuring out the source of a warning or error easier.


rly:
Unit test output in verbose mode now looks like this (screenshot from CircleCI), which is so much more readable.

image

ajtritt (changes requested):
can you move coloredtestrunner.py into tests/?

ajtritt (approved):

PR #98, opened 2019-07-15
WIP zarr backend

donghekang:
## Motivation

Add support for Zarr DirectoryStore I/O

## How to test the behavior?

See the included tests

### Dependencies
- [X] Update sources to latest HDMF dev
- [X] Remove dependency on six in Zarr backend
- [X] Update requirements to the current Zarr and numcodecs version
- [X] Make the import of Zarr optional to make sure HDMF work even without Zarr installed, i.e., only the explicit import of the zarr backend classes should fail if Zarr is not installed. Update:  In ``src/hdmf/utils.py`` ,``src/hdmf/io/__init__.py`` , ``tests/unit/test_io_zarr.py`` etc.
- [ ] https://github.com/hdmf-dev/hdmf/pull/446 adds a ``get_docval_macro`` function. We should remove the ``__get_docval_macro`` that has been added in this PR and update the code to use the new one from HDMF.

### Features
- [X] Implement core Zarr I/O backend and utilities 
- [X] Builder.written flag has been removed and built status should be tracked in the I/O backend
- [X] Remove logic to support export to Zarr from HDF5 and use new export functions instead
- [X] Add support to allow the use of Zarr-supported compressors and filters for datasets
- [X] Raise NotImplemented error for RegionReferences and mention in Docs
- [X]  Add support for AbstractDataChunkIterator (also ``exhaust_dci`` flag has been moved to the HDF5 backend and removed from HDMFIO)
- [ ] Add support for for RegionReferences (Optional)

### Testing and Documentation
- [X] Add unit tests for iterative data write
- [X] Update CHANGELOG.md
- [X] Check that we can convert an NWB file from HDF5 --> Zarr --> HDF5
- [x] Add tests for export between HDF5 and Zarr and vice versa
- [ ] Tests for export from ZarrToZarr are in TestExportZarrToZarr. Export of links seems to fail. Re-enable and debug the tests.
- [ ] Check https://github.com/hdmf-dev/hdmf/pull/179/ to see if we need to update reference handling
- [ ] Resolve TODO items listed in the code
- [ ] Add example for HDF5 to Zarr conversion (and vice versa) to the export docs.

## Checklist


oruebel:
https://github.com/NeurodataWithoutBorders/pynwb/issues/300
https://github.com/NeurodataWithoutBorders/pynwb/issues/230

bendichter:
@kangdh thanks for getting this going! A bunch of us are really excited about incorporating alternative backends!

bendichter:
@kangdh I think for requirements it might make more sense to have Zarr and other optional backends be included as [extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)

donghekang:
> @kangdh I think for requirements it might make more sense to have Zarr and other optional backends be included as [extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)

Because we import zarr in the utils.py, I think zarr should be installed by default.

bendichter:
@kangdh it looks like you developed this in python 2. Could you change to python 2/3 syntax? e.g. changing `print a` to `print(a)`. Also we use flake8 for managing code style, which is pretty strict. Check the results there to see where it has issues.

oruebel:
I'm working on the fixes for Python 3

oruebel:
@kangdh the following PR https://github.com/kangdh/hdmf/pull/1 to your branch should fix the test failures for Python 2/3 and flake8. Can you please review and merge the PR with your branch.

bendichter:
What is the plan for integrating this? We have groups asking for it and I am concerned that if it sits in PRs it will accumulate merge conflicts

oruebel:
> We have groups asking for it and I am concerned that if it sits in PRs it will accumulate merge conflicts

I agree that I need to get back to this. I think the main issue is accumulation of changes in HDMF rather than merge conflicts, because this PR mainly adds new classes, but they'll also need to be updated to work with the changes in HDMF. 

Realistically, the plan is for me to finish this up during the hackathon at Allen. The main open issue are:

-  I need to resolve the problem with converting between different backends. Currently this is handled a bit hackish in that I am checking Zarr whether the builder was written to HDF5 and ignore the flags on the builder if that is the case. 
- I'll need to deal with is the dimension scales that Ryan is adding

bendichter:
What's the plan for this? The Allen Institute is really pushing for this to be done.

oruebel:
* This is waiting for https://github.com/hdmf-dev/hdmf/pull/326 so that we can implement a proper conversion between HDF5 and Zarr (and vice versa). The current version has some hacks in the Zarr backend implementation to allow conversion of HDF5 to Zarr but this should be cleaned up before merge.
* The PR needs to be updated to the latest version of HDMF (and PyNWB)

oruebel:
TODO:

- [ ] Update sources to HDMF dev
- [ ] Builder.written flag has been removed and built status should be tracked in the I/O backend
- [ ] Remove logic to support export to Zarr from HDF5 and use new export functions instead
- [ ] Check ``exhaust_dci`` flag has been moved to the HDF5 backend and removed from HDMFIO
- [ ] Add tests for export
- [ ]

bendichter:
@oruebel Since Zarr only supports chunked datasets, how does this backend handles writing unchunked datasets? Does it write the dataset as one big chunk?

oruebel:
In that case the ``chunk`` parameter will be set to ``False`` when calling the ``require_dataset`` function of ``zarr.hiearchy.Group`` I believe this means that Zarr will store the array in a single block, but I'd need to double check what Zarr actually does in this case when the file is written to disk.

bendichter:
That could be a problem for large datasets because a user would not be able to read a section of it without reading the entire dataset into memory

oruebel:
The default option for chunking is set to True currently in ZarrIO. We can run some test with some NWB files. The convert should be fairly simple:

```
from pynwb import NWBHDF5IO, NWBZarrIO
import os
infile = "H19.28.012.11.05-2.nwb"
outfile = "test_zarr_" + os.path.basename(infile)
h5r = NWBHDF5IO(infile , 'r', load_namespaces=False)
f = h5r.read()
zw = NWBZarrIO(outfile, 
               mode='w', 
               manager=h5r.manager, 
               chunking=True)
zw.write(f, cache_spec=True)
zw.close()
h5r.close()
```

oruebel:
Closing in favor of the new PR https://github.com/hdmf-dev/hdmf/pull/696

PR #104, opened 2019-07-16
Fix checking of docval shape on types without len

### LINKED ISSUE #103: Docval shape cannot be checked on types without len
When shape is defined for a docval argument and type can be an object that does not have a `__len__`, e.g. `ElectricalSeries.data`
```python
class ElectricalSeries(TimeSeries):
    @docval(...
            {'name': 'data', 'type': ('array_data', 'data', TimeSeries),
             'shape': ((None, ), (None, None), (None, None, None)),
             'doc': 'The data this TimeSeries dataset stores. Can also store binary data e.g. image frames'},
             ...)
```
then doing a direct link between data of one ElectricalSeries and another:
```python
ts1 = ElectricalSeries('test_ts1', [0, 1, 2, 3, 4, 5], region, timestamps=[0.0, 0.1, 0.2, 0.3, 0.4, 0.5])
ts2 = ElectricalSeries('test_ts2', ts1, region, timestamps=[0.0, 0.1, 0.2, 0.3, 0.4, 0.5])
```
results in the error: 
```
  File "C:\Users\Ryan\Documents\NWB\temp\pynwb\tests\unit\test_ecephys.py", line 42, in test_timestamps_link
    ts2 = ElectricalSeries('test_ts2', ts1, region, timestamps=[0.0, 0.1, 0.2, 0.3, 0.4, 0.5])
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 379, in func_call
    allow_extra=allow_extra)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 176, in __parse_args
    if not __shape_okay_multi(argval, arg['shape']):
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 71, in __shape_okay_multi
    return any(__shape_okay(value, a) for a in argshape)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 71, in <genexpr>
    return any(__shape_okay(value, a) for a in argshape)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 78, in __shape_okay
    if not len(valshape) == len(argshape):
TypeError: object of type 'NoneType' has no len()
```

rly:
Fixes #103

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -16,6 +16,8 @@ __macros = {
 
 
 def docval_macro(macro):
+    """Class decorator to add the class to a list of types associated with the key macro in the __macros dict
+    """
     def _dec(cls):
         if macro not in __macros:
             __macros[macro] = list()
@@ -172,7 +174,8 @@ def __parse_args(validator, args, kwargs, enforce_type=True, enforce_shape=True,
                         if not __type_okay(argval, arg['type']):
                             fmt_val = (argname, type(argval).__name__, __format_type(arg['type']))
                             type_errors.append("incorrect type for '%s' (got '%s', expected '%s')" % fmt_val)
-                    if enforce_shape and 'shape' in arg:
+                    if enforce_shape and 'shape' in arg and hasattr(argval, '__len__'):
+                        # note: shape only enforced on args that have len(), because type can be a class without len()
                         if not __shape_okay_multi(argval, arg['shape']):
                             fmt_val = (argname, get_data_shape(argval), arg['shape'])
                             value_errors.append("incorrect shape for '%s' (got '%s, expected '%s')" % fmt_val)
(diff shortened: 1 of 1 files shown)


bendichter:
This will allow data in that does not match the spec. Another option would be to do something like:

```python
if isinstance(argval, hdmf.Container):
    argval = getattr(argval, argname)
```

before the shape checks

bendichter:
Actually I guess we can assume that if it was written once it was write the first time so it will also be right this time.

rly:
Good idea. 

But I don't think we can assume the passed-in objects have been written to file. You can still create in-memory objects and pass them around. The only places I have seen where `shape` is specified for a docval argument with a Container `type` is for data and timestamps, which have this unpacking/chaining ability. I updated the code to reflect that. Note that we cannot import Container because it would be a circular import.

bendichter:
Yeah they haven't been written to file but they have gone through docval at least once to be written to memory. Ultimately I am fine with either way. This looks good to me

bendichter (approved):

PR #108, opened 2019-07-24
Add container deepcopy

### LINKED ISSUE #65: Overwriting file with object already in file creates broken link and failure to write

rly:
Fix #65 and adds generally useful functionality

This adds a `Container.__deepcopy__ function` which:
1) resets the `parent` to None
2) resets the `container_source` to None
3) sets a new `object_id`
4) sets `modified` to True
5) sets all children's `parent` to itself

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -3,7 +3,6 @@ from uuid import uuid4
 from six import with_metaclass
 from .utils import docval, getargs, ExtenderMeta
 from .query import HDMFDataset
-from .data_utils import DataIO
 from warnings import warn
 from copy import deepcopy
 
@@ -145,10 +144,6 @@ class Container(with_metaclass(ExtenderMeta, object)):
 
         for child in cp.children:
             child.parent = cp
-
-        for k, v in cp.__dict__.items():
-            if isinstance(v, DataIO):
-                setattr(cp, k, v.data)
 
         # resolve HDMFDataset after and separately because DataIO can wrap an HDMFDataset
         for k, v in cp.__dict__.items():
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -2,6 +2,8 @@ import abc
 from uuid import uuid4
 from six import with_metaclass
 from .utils import docval, getargs, ExtenderMeta
+from .query import HDMFDataset
+from .data_utils import DataIO
 from warnings import warn
 from copy import deepcopy
 
@@ -143,6 +145,15 @@ class Container(with_metaclass(ExtenderMeta, object)):
 
         for child in cp.children:
             child.parent = cp
+
+        for k, v in cp.__dict__.items():
+            if isinstance(v, DataIO):
+                setattr(cp, k, v.data)
+
+        # resolve HDMFDataset after and separately because DataIO can wrap an HDMFDataset
+        for k, v in cp.__dict__.items():
+            if isinstance(v, HDMFDataset):
+                setattr(cp, k, v.dataset)
 
         return cp
 
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -139,6 +139,7 @@ class Container(with_metaclass(ExtenderMeta, object)):
         cp.__parent = None
         cp.__container_source = None
         cp.__object_id = str(uuid4())
+        cp.__modified = True
 
         for child in cp.children:
             child.parent = cp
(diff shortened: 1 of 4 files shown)


src/hdmf/container.py
```diff
@@ -126,6 +128,30 @@ def parent(self, parent_container):
...
+        cp.__object_id = str(uuid4())
+        cp.__modified = True
+
+        for child in cp.children:
+            child.parent = cp
+
+        # resolve HDMFDataset after and separately because DataIO can wrap an HDMFDataset
+        for k, v in cp.__dict__.items():
+            if isinstance(v, HDMFDataset):
+                setattr(cp, k, v.dataset)
```
oruebel, line 151: On read ``v.dataset`` may be and ``h5py.Dataset`` with references. I think the resolution for this would be something like

```
if len(v) > 0 and isinstance(v[0], Container): 
    setattr(cp, k, [find_copied_container(c) for c in v])
else:
    setattr(cp, k, v.dataset)
```
where ``find_copied_container`` would need to replace the reference with the new copy. Not sure, what would happen if you tried to call deepcopy on, e.g., a single DynamicTable that contains references, as in this case those containers may not have been copied yet by deepcopy. 

Also, I'm not sure what happens when the container that is being copied was loaded from a file. In this case you may have h5py.Dataset objects that would need to be replaced as well (or are those all wrapped?).

francois-ecm:
Hello,
In the laboratory where I am working (INMED) we are trying to copy a NWB file and modify/add meta-data in the copy, for example to correct a mistake. However it is currently quite hard to do that, and this deepcopy function could be a great help !

oruebel:
@francois-ecm thanks for the feedback. I agree, the deepcopy function would be a good way for doing this. We are still working through some cases on the deepcopy functionality. Given other ongoing items I'd expect that it may take a few more weeks for this PR.

For appending to a file PyNWB currently supports: 1) adding new containers (e.g,. TimeSeries), and 2) updating existing fields stored in h5py.Datasets. If that is what you need to do, then the following may be a possible solution for your case in the meantime:

```
from shutil import copyfile
from pynwb import NWBHDF5IO

# Copy the file
src_file = "path/myfile"
dst_file = "path/mycopy"
copyfile(src_file, dst_file)

# Open the copy and update data
f = NWBHDF5IO(dst_file, 'a')
nf = f.read()
....  # Update dataset or add containers
f.write(nf)
```

Updating of attributes is currently not yet supported. I'm not sure how hard it would be to make this work, but if that is what you need to do, then I'd suggest to open an issue describing your use-case so that we can discuss this.

PR #122, opened 2019-07-31
Fix enforce shape in docval when argument does not have len

rly:
This fixes changes made in #104. 

Enforcing shape on an argument that could be unpacked (e.g. `TimeSeries.timestamps` could be a `TimeSeries`) and could also be an object that does not have `__len__` (e.g. a `DataChunkIterator`) would break and cause an infinite loop. This change fixes the infinite loop and uses the right check for shape (`get_data_shape`).

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -78,6 +78,8 @@ def __shape_okay_multi(value, argshape):
 
 
 def __shape_okay(value, argshape):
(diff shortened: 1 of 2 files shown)


oruebel:
Could you add some tests for these cases

ajtritt (dismissed):

ajtritt (approved):

PR #100, opened 2019-07-16
move methods from NWBBaseType to Container

bendichter:
---
name: Pull request
about: Create a pull request to help us improve
title: ''
labels: ''
assignees: ''

---

## Motivation

fix #60 

## How to test the behavior?
```
Show here how to reproduce the new behavior (can be a bug fix or a new feature)
```

## Checklist

- [ ] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [ ] Have you ensured the PR description clearly describes problem and the solution?
- [ ] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [ ] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/hdmf-dev/hdmf/pulls) for the same change?
- [ ] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ? By including "Fix #XXX" you allow GitHub to close the corresponding issue.

Issue #60, opened 2019-05-09
Move NWBBaseType functionality to Container


rly:
Some of the functionality in NWB core.py is general-purpose and should eventually be extracted into HDMF, including all of NWBBaseType. 

Aside: tracking modified data, if we want to go down that route, should happen in _setter.

PR #125, opened 2019-08-01
Add enhanced LabelledDict class and tests

rly:
This PR moves LabelledDict from PyNWB over to HDMF, enhances (/slightly changes) its functionality, adds error checking, and adds unit tests.

See #60. One step to doing #100.

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -636,3 +636,105 @@ def pystr(s):
         return s.decode('utf-8')
     else:
         return s
+
+
+class LabelledDict(dict):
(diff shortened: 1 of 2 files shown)


rly:
LabelledDict is currently unused in HDMF but will be used when #100 is revised and merged. Note that unlike LabelledDict in PyNWB:
- Only objects with the key attribute can be added to the LabelledDict
- If the syntax my_dict[other_attr == val] is used, a set is returned instead of a list. Dicts are unordered, so a set is more appropriate.

ajtritt (approved):

PR #132, opened 2019-08-07
[WIP] Enh. DataChunkIterator, support: 1) h5 dataset, 2) skipping of blocks, 3) arbit. iter dim

oruebel:
## Motivation

* When wrapping an h5py.Dataset object with DataChunkIterator, DataChunkIterator currently uses the iterator from the dataset to read the dataset one-slice-at-a-time. In particular when slices are buffered into a chunk, this is inefficient as we can instead just read the next block in one read, rather than iteratively. This helps to significantly reduce the I/O operatiosn needed and accoringly helps to improve performance significantly. 
* It is currently not possible to skip empty blocks in DataChunkIterator. This PR allows iterators to return None for empty chunks and in turn DataChunkIterator will ignore those chunks and fast-forward to the next or alterantively, if an empty chunk is encountered while buffering, it will stop the chunk and return the data is has accumulated so far. Skipping chunks is useful when we have sparse data as it avoids allocating empty data blocks.
* DataChunkIterator currently only supports iteration over the first dimension of an nD array. This PR enhances DataChunkIterator to support iterator over an arbitrary user-defined dimension.

**Status** Need to add tests to make sure this is working. 
 
## How to test the behavior?
```
```

## Checklist


src/hdmf/data_utils.py
```diff
@@ -200,13 +221,19 @@ def _read_next_chunk(self):
             else:
                 self.__next_chunk.data = np.asarray(curr_next_chunk)
                 if self.__next_chunk.selection is None:
-                    self.__next_chunk.selection = slice(0, next_chunk_size)
+                    self.__next_chunk.selection = slice(start=curr_chunk_offset,
+                                                        stop=curr_chunk_offset+next_chunk_size)
```
oruebel: @rly the changes in line 204 to 225 should hopefully do the trick to skip blocks

src/hdmf/data_utils.py
```diff
@@ -200,13 +221,19 @@ def _read_next_chunk(self):
             else:
                 self.__next_chunk.data = np.asarray(curr_next_chunk)
                 if self.__next_chunk.selection is None:
-                    self.__next_chunk.selection = slice(0, next_chunk_size)
+                    self.__next_chunk.selection = slice(start=curr_chunk_offset,
```
rly: FYI I got an error here: `TypeError: slice() takes no keyword arguments`

rly:
If the first value returned by the iterator is None, then this error is generated:
```
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 875, in __chunked_iter_fill__
    dset = parent.create_dataset(name, **io_settings)
  File "C:\Users\Ryan\Miniconda3\envs\pynwb-hdmf-dev\lib\site-packages\h5py-2.9.0-py3.7-win-amd64.egg\h5py\_hl\group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "C:\Users\Ryan\Miniconda3\envs\pynwb-hdmf-dev\lib\site-packages\h5py-2.9.0-py3.7-win-amd64.egg\h5py\_hl\dataset.py", line 89, in make_new_dset
    raise TypeError("One of data, shape or dtype must be specified")
TypeError: One of data, shape or dtype must be specified
```

If we can't get around this, it would be useful to raise a warning in the DataChunkIterator.

rly:
TODO before review:

- [ ] write tests for reading h5 datasets
- [ ] write tests for writing data with datachunkiterator on a different dimension

@oruebel could you take a look at my changes so far? it looks like everything works but i messed with the internal logic of `_read_next_chunk` quite a bit.

rly:
@oruebel this is finally ready for review

rly (approved):

PR #143, opened 2019-09-25
Infer dtype from data when spec dtype is none or numeric

### LINKED ISSUE #140: DataChunkIterator converting data to float
## Description

`pynwb==1.1.0.post0.dev1` and `hdmf==1.2.0.post0.dev1`
I’m trying to store `int16` data using DataChunkIterator but it seems to be converting dtype to `float`. Here’s an example:

```python
from datetime import datetime
from pynwb import NWBFile
from pynwb import NWBHDF5IO
from pynwb.ecephys import ElectricalSeries
from hdmf.data_utils import DataChunkIterator
import numpy as np
import os

nwb = NWBFile(session_description='session', identifier='1', session_start_time=datetime.now(tzlocal()))

# Add electrode groups and channels
nChannels = 10
dev0 = nwb.create_device(name='dev0')
elecs_group = nwb.create_electrode_group(name='electrodes', description='', location='ctx', device=dev0)
for i in np.arange(nChannels):
    ii = float(i)
    nwb.add_electrode(x=ii, y=ii, z=ii, imp=ii, location='', filtering='', group=elecs_group)

#Add signal
elecs_region = nwb.electrodes.create_region(name='electrodes',
                                            region=np.arange(nChannels).tolist(),
                                            description='')

def data_generator(nChannels):
    X_data = np.zeros((nChannels,1000)).astype('int16')
    for id in range(0, nChannels):
        data = X_data[id,:]
        print(data.dtype)
        yield data
data = data_generator(nChannels=10)
iter_data = DataChunkIterator(data=data, iter_axis=1)

signal = ElectricalSeries(name='ElectricalSeries', 
                          data=iter_data, 
                          electrodes=elecs_region,
                          starting_time=0.,
                          rate=1.)
nwb.add_acquisition(signal)

#Write file
with NWBHDF5IO('file_1.nwb', mode='w') as io:
    io.write(nwb)
    
# Read file
io = NWBHDF5IO('file_1.nwb', 'r+')
nwbfile = io.read()
print(nwbfile.acquisition['ElectricalSeries'].data[:,0].dtype)
io.close()
```

If I pass the data directly, it works fine:
```python
# iter_data = DataChunkIterator(data=data, iter_axis=1)
iter_data = np.zeros((nChannels,1000)).astype('int16')
```
## Checklist

- [ x] Have you ensured the feature or change was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?
- [ x] Have you included a brief and descriptive title?
- [ x] Have you included a clear description of the problem you are trying to solve?
- [ x] Have you included a minimal code snippet that reproduces the issue you are encountering?
- [ x] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?

### LINKED ISSUE #142: 'numeric' dtype in spec acts as None dtype
Related to #140 

For most subtypes of TimeSeries, the "data" dataset has the spec 'dtype: numeric'. In fact, that is the only place in the NWB schema where 'numeric' is used. 

As far as I can see, the only times that 'numeric' is actually used in HDMF are in the validator and in the following code:
https://github.com/hdmf-dev/hdmf/blob/5a9a9cb5f371dcb66c1f08ab84cef4c8506f4e69/src/hdmf/build/map.py#L474-L477

where 'numeric' serves the same purpose as not specifying a dtype at all. 

When `value, None` is returned above, this None is used as the dtype for the DatasetBuilder, and this messes with code downstream where 'dtype=None' is used when writing an AbstractDataChunkIterator (#140).

The above code logic seems like it would also cause other problems that have not yet come up. For example, I can write a list of strings for 'data' of an ElectricalSeries:
```python
signal = ElectricalSeries(name='ElectricalSeries',
                          data=['hello'],
                          electrodes=elecs_region,
                          starting_time=0.,
                          rate=1.)
```
Writing and reading this works just fine, but this should not be allowed. I think the check for that should be somewhere in HDMF, though I have not yet found it. Help in tackling this would be appreciated.

rly:
Fixes #140 and fixes #142.

If an AbstractDataChunkIterator is used for a dataset that has the dtype 'numeric' or no dtype in the spec, then the dtype used to write the dataset will be inferred from the `dtype` of the AbstractDataChunkIterator. Same for all numpy types, lists and tuples, and Python primitives. Previously, the dtype used to write the dataset was set to None, and so I believe h5py used a default dtype of float32.


rly:
I still need to add tests, but just wanted to put the proposed solution out there.

luiztauffer:
I tested on my code and this PR solved the problem, related to #140 , thanks!

ajtritt (approved):

oruebel (approved):
Looks good to me.

PR #148, opened 2019-09-26
Attempt to add PyNWB test to CircleCI

### LINKED ISSUE #147: Add PyNWB tests to CI
Changes in HDMF may break PyNWB. It would help the developers to setup CI such that PyNWB tests are run automatically on each PR for HDMF. These tests could just be run on Linux with Python 3.7. These tests will not always succeed since changes here may be needed before corresponding changes in PyNWB can be merged into the dev branch.

rly:
Fix #147

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -275,7 +275,7 @@ jobs:
             pip check
             python setup.py develop
             cd ..
-            pip uninstall -y hdmf==1.2.0
+            pip uninstall -y hdmf
             python setup.py develop
             pip show hdmf
           # Install is expected to be quick. Increase timeout in case there are some network issues.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -256,6 +256,38 @@ jobs:
...
+          command: |
+            . ../venv/bin/activate
+            git clone https://github.com/NeurodataWithoutBorders/pynwb.git
+            cd pynwb
+            pip install -U -rrequirements-dev.txt -rrequirements.txt
+            pip check
+            python setup.py develop
+            cd ..
+            pip uninstall hdmf==1.2.0
+            python setup.py develop
```
oruebel, line 281: ``pip uninstall hdmf==1.2.0`` uninstalling the pinned version is dangerous in that it means you will need to update the test any time PyNWB changes the version of HDMF it is using

rly: Good point.

oruebel (approved):

ajtritt (approved):

PR #150, opened 2019-09-26
Bug/bwd comp

ajtritt:
Fix something that broke PyNWB.

### DIFF
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -272,9 +272,6 @@ class NamespaceCatalog(object):
             {'name': 'namespace', 'type': SpecNamespace, 'doc': 'the SpecNamespace object'})
     def add_namespace(self, **kwargs):
         """Add a namespace to this catalog"""
-        msg = ("NamespaceCatalog.add_namespace has been deprecated. "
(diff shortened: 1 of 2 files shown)


src/hdmf/spec/namespace.py
```diff
@@ -272,9 +272,6 @@ def spec_namespace_cls(self):
             {'name': 'namespace', 'type': SpecNamespace, 'doc': 'the SpecNamespace object'})
     def add_namespace(self, **kwargs):
         """Add a namespace to this catalog"""
-        msg = ("NamespaceCatalog.add_namespace has been deprecated. "
-               "SpecNamespaces should be added with load_namespaces.")
-        warn(msg, DeprecationWarning)
```
rly, line 277: Why un-deprecate this function?
Why was it deprecated in the first place?

ajtritt: I deprecated it because I thought you should only be able to add namespaces to a catalog using the load_namespaces functionality. That was more of a hunch than anything real. Eventually, I realized the utility in the add_namespaces functionality, so I'm undeprecating it.

rly (approved):

PR #157, opened 2019-09-29
Throw specific exception when no data type specified for root group

rly:
This is helpful to address https://github.com/NeurodataWithoutBorders/pynwb/issues/1077 where an incompatible file is read.

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -285,6 +285,7 @@ class HDF5IO(HDMFIO):
                 writer = H5SpecWriter(ns_group)
                 ns_builder.export('namespace', writer=writer)
 
+    @docval(returns='the Container object that was read in', rtype=Container)
     def read(self, **kwargs):
         if self.__mode == 'w' or self.__mode == 'w-' or self.__mode == 'x':
             raise UnsupportedOperation("Cannot read from file %s in mode '%s'. Please use mode 'r', 'r+', or 'a'."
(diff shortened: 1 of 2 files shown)


t-b:
@rly How about detecting NWBv1 files as well and give the user a good error message. If written and found something at https://github.com/AllenInstitute/ipfx/blob/664f9f3578fe9a34f27c80b742b2af91459521c7/ipfx/nwb_reader.py#L606.

rly:
@t-b thanks for the code snippet. I had started to write something similar in pynwb which relies on catching the above exception, but maybe the NWB version should be checked before triggering the whole builder parsing logic. I'll put up a PR in pynwb soon for feedback.

oruebel (approved):

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1043,3 +1049,7 @@ def __rec_get_ref(self, l):
             else:
                 ret.append(elem)
         return ret
+
+
+class MissingRootException(ValueError):
```
ajtritt, line 1054: I don't think this should ever happen. Root always exists by definition.

ajtritt (changes requested):
I'm not sure what the best solution is, but this seems like it could be a bandage on a larger problem. `root` could not have a `data_type` if you just added some objects (with `data_type`) to the root group, which seems like a perfectly okay thing to do.

rly:
Rethinking this, I agree with @ajtritt that this is not the best approach to the problem.

PR #159, opened 2019-10-01
Add ability to apply DataIO to Data after construction

ajtritt:
## Motivation

This feature will be necessary to address https://github.com/NeurodataWithoutBorders/pynwb/issues/1067

## How to test the behavior?
See test


bendichter:
@ajtritt Would I apply this feature with something like this?

in `pynwb/io/core.py`:
```python
@register_map(DynamicTableMap)
class NWBDynamicTableMap(DynamicTableMap):

    @docval(*get_docval(DynamicTableMap.get_attr_value, 'spec', 'container', 'manager'),
            returns='the value of the attribute')
    def get_attr_value(self, **kwargs):
        spec, container, manager = getargs('spec', 'container', 'manager', kwargs)
        attr_value = super(NWBDynamicTableMap, self).get_attr_value(spec, container, manager)
        attr_value.set_dataio(H5DataIO(chunks=True))

        return attr_value
```

I am getting the following error:
```python
  ImportError: Failed to import test module: test_io
    File "/Users/bendichter/dev/pynwb/.tox/py37/lib/python3.7/site-packages/unittest2/loader.py", line 456, in _find_test_path
      module = self._get_module_from_name(name)
    File "/Users/bendichter/dev/pynwb/.tox/py37/lib/python3.7/site-packages/unittest2/loader.py", line 395, in _get_module_from_name
      __import__(name)
    File "/Users/bendichter/dev/pynwb/tests/integration/test_io.py", line 7, in <module>
      from pynwb import NWBFile, TimeSeries, get_manager, NWBHDF5IO, validate
    File "/Users/bendichter/dev/pynwb/src/pynwb/__init__.py", line 229, in <module>
      from . import io as __io  # noqa: F401,E402
    File "/Users/bendichter/dev/pynwb/src/pynwb/io/__init__.py", line 1, in <module>
      from . import base as __base
    File "/Users/bendichter/dev/pynwb/src/pynwb/io/base.py", line 1, in <module>
      from .core import NWBContainerMapper
    File "/Users/bendichter/dev/pynwb/src/pynwb/io/core.py", line 24, in <module>
      class NWBContainerMapper(NWBBaseTypeMapper):
    File "/Users/bendichter/dev/pynwb/src/pynwb/__init__.py", line 152, in _dec
      __TYPE_MAP.register_map(container_cls, cls)
    File "/Users/bendichter/dev/pynwb/.tox/py37/lib/python3.7/site-packages/hdmf/utils.py", line 438, in func_call
      return func(self, **parsed['args'])
    File "/Users/bendichter/dev/pynwb/.tox/py37/lib/python3.7/site-packages/hdmf/build/map.py", line 1669, in register_map
      raise ValueError('cannot register map for type %s - no data_type found' % container_cls)
  ValueError: cannot register map for type <class 'pynwb.core.NWBContainer'> - no data_type found
```

but I may not be applying the method right

rly:
Please also add tests where data is None and data is an h5py.Dataset. Otherwise, this looks good.

tests/unit/test_data_utils.py
```diff
@@ -0,0 +1,19 @@
...
+    def test_set_dataio(self):
+        """
+        Test that Data.set_dataio works as intended
+        """
+        dataio = H5DataIO(chunks=(1, 1, 3))
+        data = np.arange(30).reshape(5, 2, 3)
+        container = Data('wrapped_data', data)
+        container.set_dataio(dataio)
+        self.assertIs(dataio.data, data)
+        self.assertIs(dataio, container.data)
```
oruebel: I think this test should be added here https://github.com/hdmf-dev/hdmf/blob/dev/tests/unit/utils_test/test_core_DataIO.py to make sure we don't have DataIO tests in multiple places.

src/hdmf/region.py
```diff
@@ -0,0 +1,91 @@
+from abc import ABCMeta, abstractmethod
```
oruebel, line 1: Not sure if we already have specific tests for RegionSlicer and ListSlicer but if so, they should probably also move to a corresponding file for testing this module.

ajtritt:
@bendichter

```python
@register_map(DynamicTableMap)
class NWBDynamicTableMap(DynamicTableMap):
```
should be 

```python
@register_map(DynamicTable)
class NWBDynamicTableMap(DynamicTableMap):
```

src/hdmf/region.py
ajtritt: I don't think we do

ajtritt:
@rly 

> Please also add tests where data is None and data is an h5py.Dataset

This functionality is independent of h5py, and all new lines have been hit by the tests, so I'm not sure what this use case adds. Please clarify.

oruebel (approved):

bendichter:
Ryan and I successfully tested this for our application with chunked data in `DynamicTables`.

rly:
> @rly
> 
> > Please also add tests where data is None and data is an h5py.Dataset
> 
> This functionality is independent of h5py, and all new lines have been hit by the tests, so I'm not sure what this use case adds. Please clarify.

@ajtritt Never mind. Those tests are not needed. Separately, it may be worth adding `'default': None` to the docval of `H5DataIO` though, so that the new `set_dataio` function could work with `H5DataIO`.

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -215,7 +216,8 @@ class H5DataIO(DataIO):
     @docval({'name': 'data',
              'type': (np.ndarray, list, tuple, Dataset, Iterable),
              'doc': 'the data to be written. NOTE: If an h5py.Dataset is used, all other settings but link_data' +
-                    ' will be ignored as the dataset will either be linked to or copied as is in H5DataIO.'},
+                    ' will be ignored as the dataset will either be linked to or copied as is in H5DataIO.',
```
ajtritt, line 219: @rly is this what you mean?

rly: yes, somehow i missed this. my bad!

rly (approved):

PR #161, opened 2019-10-02
add append for Datasets in Data objects

bendichter:
change table to use the newly defined append method

needed for https://github.com/NeurodataWithoutBorders/pynwb/issues/1067

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -429,7 +429,7 @@ class Data(AbstractContainer):
             self.data.extend(arg)
         elif isinstance(self.data, np.ndarray):
             self.__data = np.append(self.__data, [arg])
-        elif isinstance(h5py.Dataset):
+        elif isinstance(self.data, h5py.Dataset):
             shape = list(self.__data.shape)
             shape[0] += len(arg)
             self.__data.resize(shape)
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -5,6 +5,7 @@ from six import with_metaclass
 from .utils import docval, get_docval, call_docval_func, getargs, ExtenderMeta
 from .data_utils import DataIO
 from warnings import warn
+import h5py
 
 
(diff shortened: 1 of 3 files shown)


src/hdmf/container.py
```diff
@@ -423,6 +429,11 @@ def extend(self, arg):
             self.data.extend(arg)
         elif isinstance(self.data, np.ndarray):
             self.__data = np.append(self.__data, [arg])
+        elif isinstance(h5py.Dataset):
```
bendichter:
```suggestion
        elif isinstance(self.data, h5py.Dataset):
```

oruebel (changes requested):
It would be nice to have tests for these new cases for append. Other than that this looks good to me.

rly:
Suggestion: on line 58 of src/hdmf/common/table.py: change `self.data.append` to `self.append`
Otherwise, we will have a similar problem that we had before where `self.data` is an `h5py.Dataset` and `h5py.Dataset` does not have the function `append`.

bendichter:
@oruebel yeah, we should write some tests in hdmf. I just wanted to get this in the queue so the whole team could be on board with these changes.

bendichter:
I added the tests to `test_io_map_data.py` as this was the only place in the tests I found an instantiated `Data`.

rly (approved):

ajtritt (approved):

bendichter:
@oruebel is this OK to merge?

oruebel (approved):
@bendichter looks good. Just FYI, this change introduces an issue in that with this ``Data`` is now dependent on the HDF5 backend. This means., if we add new backends the changes are no longer contained to a new HDMFIO class, but we'd need to update ``Data`` as well. However, this is a non-trivial issue. For now, lets merge this and we'll worry about fixing this as we work on getting the Zarr backend ready for merge.

bendichter:
@oruebel Yes, I agree it would be better to define these methods in a way that is supportable within the different backend modules once we are supporting other backends.

PR #165, opened 2019-10-04
RF: git submodule to subtree for src/hdmf/common/hdmf-common-schema

yarikoptic:
As hinted in https://github.com/NeurodataWithoutBorders/pynwb/issues/1076#issuecomment-536796896 , IMHO git subtree mechanism should be a better fit than submodule for inclusion of `src/hdmf/common/hdmf-common-schema`.

Benefits
- no need for any additional `git submodule` manipulations (so no need for custom procedures to get perspective contributors to read the docs)
- switching between different states/branches of the project with different state of that subtree would not require any additional command

I have used the https://gist.githubusercontent.com/Nikita240/0c98cea8f53a15e69699cd8bc40657c4/raw/84c4020fe9384a0eafd79bf853745af0ded902ed/subtrees.sh helper mentioned on [SO](https://stackoverflow.com/a/44407172) to do that, and then only adjusted CI/docs.

Remaining TODOs (in case you decide to go this route):
- [ ] check if any existing branch needs the same dance and do it there as well
- [ ] extend DOCs on how to update that subtree whenever needed

Feel free to ignore / close.  I just thought to propose  ;)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,6 @@ references:
   ci-steps: &ci-steps
     steps:
       - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
       - run:
           <<: *initialize-venv
       - run:
@@ -67,8 +65,6 @@ references:
   conda-steps: &conda-steps
     steps:
       - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
       - run:
           name: Configure conda
           command: |
@@ -93,8 +89,6 @@ references:
   gallery-steps: &gallery-steps
     steps:
       - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
       - restore_cache:
           keys:
             - ophys-data-cache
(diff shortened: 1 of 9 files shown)


t-b:
I would prefer sticking with submodules.

The major feature of git submodules is that you have completely separate histories and that is not by-design the case for git subtree.

Quoting https://www.atlassian.com/git/tutorials/git-subtree:

> The responsibility of not mixing super and sub-project code in commits lies with you.

oruebel:
Another question is how this will interact with PyNWB and NWB-Schema. I'd prefer to have a consistent strategy fro managing these dependencies across the repos. @ajtritt toughts?

yarikoptic:
> Another question is how this will interact with PyNWB and NWB-Schema

my idea was to introduce similar replacement in PyNWB later if decision here was made to adopt subtree.

yarikoptic:
> I would prefer sticking with submodules.
> 
> The major feature of git submodules is that you have completely separate histories and that is not by-design the case for git subtree.
> 
> Quoting https://www.atlassian.com/git/tutorials/git-subtree:
> 
> > The responsibility of not mixing super and sub-project code in commits lies with you.

FWIW: That is correct. submodules indeed help a bit to effectively avoid creating a "fork" of the submodule/subtree.

rly:
We decided to delay changes in the software process until there is a need.

PR #167, opened 2019-10-08
Allow SZIP compression options 

### LINKED ISSUE #166: szip compression parameters type checking fails in H5DATAIO
Currently szip can only be called without compression options.

Compression options for szip should be in the form of a tuple ('ec'|'nn', even integer 0-32). If a tuple, e.g. ```('ec',16)``` is passed to H5DATAIO:

```
>>> TypeError: incorrect type for 'compression_opts' (got 'tuple', expected 'int')
```

for

```
type(data) # numpy array
H5DataIO(data, chunks=(3,4,5,6)_A, compression="szip", compression_opts=('ec',16), shuffle=False)
```

oruebel:
## Motivation

Fix #166 
Enhance parameter checking in H5DataIO
Add tests for parameter checking

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -5,6 +5,7 @@ except ImportError:
     from collections import Iterable  # Python 2.7
 from six import binary_type, text_type
 from h5py import Group, Dataset, RegionReference, Reference, special_dtype
+from h5py import filters as h5py_filters
 import json
 import numpy as np
 import warnings
@@ -233,7 +234,7 @@ class H5DataIO(DataIO):
                     'http://docs.h5py.org/en/latest/high/dataset.html#dataset-compression',
              'default': None},
             {'name': 'compression_opts',
-             'type': int,
+             'type': (int, tuple),
              'doc': 'Parameter for compression filter',
              'default': None},
             {'name': 'fillvalue',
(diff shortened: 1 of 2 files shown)


ajtritt (dismissed):

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -150,6 +151,35 @@ def test_write_dataset_list_compress(self):
...
+    def test_write_dataset_list_compress_szip(self):
+        a = H5DataIO(np.arange(30).reshape(5, 2, 3),
+                     compression='szip',
+                     compression_opts=('ec', 16),
+                     shuffle=True,
+                     fletcher32=True)
+        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', a, attributes={}))
+        dset = self.f['test_dataset']
+        self.assertTrue(np.all(dset[:] == a.data))
+        self.assertEqual(dset.compression, 'lzf')
```
bendichter:
```suggestion
        self.assertEqual(dset.compression, 'szip')
```

bendichter: otherwise looks good!

oruebel: @bendichter good catch. This should be fixed now. I don't have SZIP installed so I unfortunately could not run that test case.

oruebel:
@rly @bendichter @ajtritt could you check and approve. I had to dismiss your approval to fix a bug that Ben had found.

ajtritt (approved):

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -277,25 +278,66 @@ def __init__(self, **kwargs):
...
+                        raise ValueError("GZIP compression_opts setting must be an integer from 0-9, " +
+                                         "not " + str(self.__iosettings['compression_opts']))
+                elif self.__iosettings['compression'] == 'lzf':
+                    if self.__iosettings['compression_opts'] is not None:
+                        raise ValueError("LZF compression filter accepts no compression_opts")
+                elif self.__iosettings['compression'] == 'szip':
+                    if not isinstance(self.__iosettings['compression_opts'], tuple) or \
+                                    len(self.__iosettings['compression_opts']) != 2:
+                        raise ValueError("SZIP compression filter compression_opts" +
+                                         " must be a 2-tuple ('ec'|'nn', even integer 0-32")
```
rly, line 322: Super minor, but regarding code style, I'm pretty sure the backslash and plus symbols are not necessary. 

PEP8 says: "The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation."

rly: You need an ending parenthesis inside the last quote marker on line 322

rly: Since we check for gzip compression opts being an int from 0-9, can we also check for the szip compression opts being ('ec'|'nn', even integer 0-32)?

oruebel: Not sure what the issue with 322 is?

> szip compression opts being ('ec'|'nn', even integer 0-32)?

sure

rly: Not a syntax error - but within the string, the "(" before "'ec'" needs a corresponding ")".

PR #173, opened 2019-10-16
pass through DataIO params after formatting strings

### LINKED ISSUE #162: H5DataIO.chunks arg ignored for strings
For https://github.com/NeurodataWithoutBorders/pynwb/issues/1067 we need to have string vectors that are chunked. When I try to chunk string objects in pynwb (using hdmf), they result in a dataset that is not chunked. When I try chunking them in h5py there is no problem and they are chunked as expected.

```python
from datetime import datetime
from pynwb import NWBFile, NWBHDF5IO, TimeSeries
from pynwb.misc import AnnotationSeries
from hdmf.backends.hdf5 import H5DataIO


tso = TimeSeries(name='test_ts', data=H5DataIO(data=[1, 2, 3], chunks=True), timestamps=[1, 2, 3])
aso = AnnotationSeries(name='test_AS', data=H5DataIO(data=('aa', 'bb'), chunks=True), timestamps=(1., 2.))
nwb = NWBFile(session_description='session', identifier='1', session_start_time=datetime.now(tzlocal()))
nwb.add_acquisition(aso)
nwb.add_acquisition(tso)

with NWBHDF5IO('test_file.nwb', 'w') as io:
    io.write(nwb)

with NWBHDF5IO('test_file.nwb', 'r') as io:
    nwb = io.read()
    print('TimeSeries:')
    print('{}, {}'.format(nwb.acquisition['test_ts'].data.dtype, nwb.acquisition['test_ts'].data.chunks))
    print('')
    print('AnnotationSeries:')
    print('{}, {}'.format(nwb.acquisition['test_AS'].data.dtype, nwb.acquisition['test_AS'].data.chunks))


from h5py import File, special_dtype
import numpy as np

file = File('test.h5', 'w')
aa = file.create_dataset(name='test_ds', data=np.string_((u'aa', u'bb')), dtype=special_dtype(vlen=str),
                         chunks=True)
print('')
print('vlen-str in h5py:')
print('{}, {}'.format(aa.dtype, aa.chunks))
```

```
TimeSeries:
int64, (3,)

AnnotationSeries:
object, None

vlen-str in h5py:
object, (2,)
```

ajtritt:
## Motivation

fix #162

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -599,6 +599,13 @@ class DataIO(with_metaclass(ABCMeta, object)):
         data = popargs('data', kwargs)
         self.__data = data
 
+    @abstractmethod
+    def get_io_params(self):
+        """
+        Returns a dict with the I/O parameters specifiedin in this DataIO.
+        """
+        pass
+
     @property
     def data(self):
         """Get the wrapped data object"""
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -589,7 +589,7 @@ class ShapeValidatorResult(object):
 
 
 @docval_macro('data')
-class DataIO(with_metaclass(ABCMeta, object)):
+class DataIO(object):
     """
     Base class for wrapping data arrays for I/O. Derived classes of DataIO are typically
     used to pass dataset-specific I/O parameters to the particular HDMFIO backend.
@@ -599,12 +599,11 @@ class DataIO(with_metaclass(ABCMeta, object)):
         data = popargs('data', kwargs)
         self.__data = data
 
-    @abstractmethod
     def get_io_params(self):
         """
         Returns a dict with the I/O parameters specifiedin in this DataIO.
         """
-        pass
+        return dict()
 
     @property
     def data(self):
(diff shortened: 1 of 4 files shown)


src/hdmf/data_utils.py
```diff
@@ -599,6 +599,12 @@ def __init__(self, **kwargs):
         data = popargs('data', kwargs)
         self.__data = data
 
+    def get_io_params(self):
+        """
+        Returns a dict with the I/O parameters specifiedin in this DataIO.
+        """
+        return dict()
```
bendichter, line 606: Is this right? Just return an empty `dict`?

bendichter (commented):
It would be nice to see a full IO-to-disk test that shows the resulting HDF dataset is in fact chunked

oruebel (dismissed):
Looks good. Thanks for chasing down this bug.

rly (approved):

rly:
> It would be nice to see a full IO-to-disk test that shows the resulting HDF dataset is in fact chunked

Agreed

src/hdmf/data_utils.py
ajtritt: The base DataIO class offers no ability to specify parameters, so there's not really anything to return

PR #179, opened 2019-10-18
resolve references properly

ajtritt:
## Motivation

Reference resolution was _always_ returning Containers, even if Containers had not been built. This PR adds code to resolve to Containers and Builders in Container and Builder contexts, respectively.

This should help the problem underlying the error seen [here](https://github.com/NeurodataWithoutBorders/pynwb/pull/1094#issuecomment-543118777) on https://github.com/NeurodataWithoutBorders/pynwb/pull/1094

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -46,17 +46,23 @@ class H5Dataset(HDMFDataset):
 
 
 class DatasetOfReferences(with_metaclass(ABCMeta, H5Dataset, ReferenceResolver)):
+    """
+    An extension of the base ReferenceResolver class to add more abstract methods for
+    subclasses that will read HDF5 references
+    """
 
     @abstractmethod
     def get_object(self, h5obj):
-        pass
-
-    @classmethod
-    @abstractmethod
-    def get_inverse_class(cls):
+        """
+        A class that maps an HDF5 object to a Builder or Container
+        """
         pass
 
     def invert(self):
+        """
+        Return an object that defers reference resolution
+        but in the opposite direction.
+        """
         if not hasattr(self, '__inverted'):
             cls = self.get_inverse_class()
             docval = get_docval(cls.__init__)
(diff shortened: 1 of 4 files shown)


src/hdmf/query.py
```diff
@@ -160,3 +161,30 @@ def __next__(self):
...
+
+    pass
+
+
+class ContainerResolver(ReferenceResolver):
+    """
+    A reference resolver that resolves references to Containers
+    """
+
+    pass
```
oruebel: On the base resolver classes, would it make sense to have 

```
def get_object(self, inobj):
     raise NotImplementedError
```

to define the functions that the subclasses need to implement?

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -44,15 +45,45 @@ def shape(self):
...
+
+    def get_object(self, h5obj):
+        return self.io.get_builder(h5obj)
+
+
+class ContainerResolverMixin(ContainerResolver):
+
+    def get_object(self, h5obj):
+        return self.io.get_container(h5obj)
+
```
oruebel: Can you clarify why we need these Mixin classes in addition to the ContainerResolver and BuildResolver classes? It looks like the Mixin classes is the only place where these get used.

src/hdmf/query.py
ajtritt: I don’t think so, since there’s no guarantee that a backend supports a 1-1 object mapping.

src/hdmf/query.py
```diff
@@ -160,3 +161,58 @@ def __next__(self):
...
+
+
+class ContainerResolver(ReferenceResolver):
+    """
+    A reference resolver that resolves references to Containers
+
+    Subclasses should implement invert and the calssmethod get_inverse_class
```
oruebel: Spelling error: "calssmethod " --> "classmethod". The same is in the docstring of BuilderResolver and ContainerResolver

oruebel: Shouldn't subtypes also implement the invert function? If so, this should be mentioned in the docstring as well.

src/hdmf/build/map.py
```diff
@@ -1127,9 +1130,18 @@ def __get_subspec_values(self, builder, spec, manager):
         elif isinstance(spec, DatasetSpec):
             if not isinstance(builder, DatasetBuilder):
                 raise ValueError("__get_subspec_values - must pass DatasetBuilder with DatasetSpec")
-            ret[spec] = builder.data
+            ret[spec] = self.__check_ref_resolver(builder.data)
         return ret
 
+    @staticmethod
+    def __check_ref_resolver(data):
```
oruebel, line 1137: I'd rename this to ``__check_and_invert_ref_resolver`` it's a bit verbose, but I think will help with readability.

oruebel (dismissed):

oruebel (approved):

PR #182, opened 2019-10-19
write from DataChunkIterators all at once

ajtritt:
## Motivation

DataChunkIterators get read/written one at a time. This modification causes DataChunkIterators to be exhausted in a round-robin fashion, allowing DataChunkIterators to be read/written "in parallel"

## How to test the behavior?
```
```

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -69,7 +69,7 @@ class HDF5IO(HDMFIO):
         self.__built = dict()       # keep track of which files have been read
         self.__read = dict()        # keep track of each builder for each dataset/group/link
         self.__ref_queue = deque()  # a queue of the references that need to be added
-        self.__dci_queue = dequeu()       # a queue of DataChunkIterators that need to be exhausted
+        self.__dci_queue = deque()       # a queue of DataChunkIterators that need to be exhausted
         ObjectMapper.no_convert(Dataset)
 
     @property
@@ -500,9 +500,9 @@ class HDF5IO(HDMFIO):
     def write_builder(self, **kwargs):
         f_builder, link_data = getargs('builder', 'link_data', kwargs)
         for name, gbldr in f_builder.groups.items():
-            self.write_group(self.__file, gbldr)
+            self.write_group(self.__file, gbldr, exhaust_dci=False)
         for name, dbldr in f_builder.datasets.items():
-            self.write_dataset(self.__file, dbldr, link_data)
+            self.write_dataset(self.__file, dbldr, link_data, exhaust_dci=False)
         for name, lbldr in f_builder.links.items():
             self.write_link(self.__file, lbldr)
         self.set_attributes(self.__file, f_builder.attributes)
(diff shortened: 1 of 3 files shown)


oruebel (commented):
Yes, I believe that should do the trick

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -69,6 +69,7 @@ def __init__(self, **kwargs):
         self.__built = dict()       # keep track of which files have been read
         self.__read = dict()        # keep track of each builder for each dataset/group/link
         self.__ref_queue = deque()  # a queue of the references that need to be added
+        self.__dci_queue = dequeu()       # a queue of DataChunkIterators that need to be exhausted
```
```suggestion
        self.__dci_queue = deque()       # a queue of DataChunkIterators that need to be exhausted
```

rly (changes requested):
Can you add a test of this?

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -943,6 +974,29 @@ def __chunked_iter_fill__(cls, parent, name, data, options=None):
                 dset.resize(new_shape)
             # Process and write the data
             dset[chunk_i.selection] = chunk_i.data
+        except StopIteration:
+            return False
+        return True
+
+    @classmethod
+    def __chunked_iter_fill__(cls, parent, name, data, options=None):
```
rly, line 989: Should this function be called somewhere? As far as I can see, it is not called anywhere, and could just be removed. But perhaps it should be called if exhaust_dci is false (line 875)?

ajtritt: I originally took it out, but put it back for backward compatibility.

rly: Ok. Can you add a comment saying this is currently unused? Or remove it? Since it is a private function, I don't think it is needed for backwards compatibility. All of the tests of this function would have to be change though.

ajtritt: It’s not private. Double underscores on one side makes it private. Double underscores on both sides makes it public. I want to keep it around so we still have a function for exhausting a DataChunkIterator

oruebel: Ultimately, I think it may be useful to have this configurable, i.e., be able to switch behavior via a flag to either do it round-robin or exhaust one-at-a-time. If you need to merge this, then I think it's fine to do that in a separate issue/pr.

ajtritt: That was the intention of the exhaust_dci argument of write_dataset

rly (dismissed):

oruebel (dismissed):

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -265,6 +265,22 @@ def test_write_dataset_iterable_multidimensional_array(self):
...
+        b = np.arange(30, 60).reshape(5, 2, 3)
+        aiter = iter(a)
+        biter = iter(b)
+        daiter1 = DataChunkIterator.from_iterable(aiter, buffer_size=2)
+        daiter2 = DataChunkIterator.from_iterable(biter, buffer_size=2)
+        builder = GroupBuilder("root")
+        builder.add_dataset('test_dataset1', daiter1, attributes={})
+        builder.add_dataset('test_dataset2', daiter2, attributes={})
+        self.io.write_builder(builder)
+        dset1 = self.f['test_dataset1']
```
oruebel, line 301: Should have this test be run twice, i.e., once fro  exhaust_dci=False and once with exhaust_dci=True?

ajtritt: Sure. Just added.

oruebel (approved):

PR #185, opened 2019-10-24
Support writing DynamicTables with DataChunkIterators

ajtritt:
## Motivation

Fix issues with writing to DynamicTable with DataChunkIterator

## How to test the behavior?
```
```


oruebel (dismissed):

ajtritt:
addressed by #390

PR #186, opened 2019-10-24
Allow enforcing unique ids on DynamicTable

oruebel:
## Motivation

Allow us to enforce the requirement for unique row ids in the DynamicTable when adding rows. This is related to https://github.com/oruebel/ndx-icephys-meta/issues/8


ajtritt (approved):

bendichter:
@oruebel I like this. Any reason you didn't set default to true?

oruebel:
> Any reason you didn't set default to true

The main reason I did not make it the default was: a) enforcing this requires additional compute for  search on the ids and b) would change the current default behavior. I am open to making this the default, though. I just wanted to avoid potential trouble by making it the default.

bendichter:
@oruebel yeah that makes sense. Let's see how it fairs as an option first and then think about transitioning it to default

PR #187, opened 2019-10-25
Allow search of ids in ElementIdentifiers

oruebel:
## Motivation

This is to allow us to address https://github.com/oruebel/ndx-icephys-meta/issues/9  When creating references to DynamicTable we often don't know the indices of the rows we need but only the list of ids. With this we can now do:

```
dst = DynamicTable(.....)
rows_with_ids = dst[ dst.id == [0,10] ]
````

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -85,6 +85,27 @@ class ElementIdentifiers(Data):
     def __init__(self, **kwargs):
         call_docval_func(super(ElementIdentifiers, self).__init__, kwargs)
 
+    @docval({'name': 'other', 'type': (Data, np.ndarray, list, tuple, int),
+             'doc': 'List of ids to search for in this ElementIdentifer object'},
+            rtype=np.ndarray,
+            returns='Array with the list of indices where the elements in the list where found.'
+                    'Note, the elements in the returned list are ordered in increasing index'
+                    'of the found elements, rather than in the order in which the elements'
+                    'where given for the search. Also the length of the result may be different from the length'
+                    'of the input array. E.g., if our ids are [1,2,3] and we are search for [3,1,5] the '
+                    'result would be [0,2] and NOT [2,0,None]')
+    def __eq__(self, other):
+        """
+        Given a list of ids return the indices in the ElementIdentifiers array where the
+        indices are found.
+        """
+        # Determin the ids we want to find
+        search_ids = other if not isinstance(other, Data) else other.data
+        if isinstance(search_ids, int):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/hdmf/common/table.py [hunk 1]
```diff
@@ -85,6 +85,27 @@ class ElementIdentifiers(Data):
...
 
+    @docval({'name': 'other', 'type': (Data, np.ndarray, list, tuple, int),
+             'doc': 'List of ids to search for in this ElementIdentifer object'},
+            rtype=np.ndarray,
+            returns='Array with the list of indices where the elements in the list where found.'
+                    'Note, the elements in the returned list are ordered in increasing index'
+                    'of the found elements, rather than in the order in which the elements'
+                    'where given for the search. Also the length of the result may be different from the length'
+                    'of the input array. E.g., if our ids are [1,2,3] and we are search for [3,1,5] the '
+                    'result would be [0,2] and NOT [2,0,None]')
```
rly, line 96: I think for `docval` to do its intended behavior, the function has to be defined as `def __eq__(self, **kwargs)`. Given that this function must be `def __eq__(self, other)` in order to override the `==` functionality, the docval decorator should be replaced with a docstring. Then it is clear that validation and all the other docval functionality does not run.

oruebel: @rly @ajtritt I added in https://github.com/hdmf-dev/hdmf/pull/187/commits/fe6f124dda449debe2fb85681fb98f900b3b746e a test to check if docval correctly performs the type checking and it seems to work fine.

oruebel: I.e., the test in https://github.com/hdmf-dev/hdmf/pull/187/commits/ed95b583bc007ca1e58d7720b3292ce901d4c55f actually tests this and seems fine too

rly: Interesting. So it does work then. Out of curiousity, does the name need to match the name of the variable? Or is it just doing positional arg matching?

oruebel: When calling ``ElementIdentifiers == 1`` Python will pass as a positional argument, i.e., the error checking in docval happens for this here https://github.com/hdmf-dev/hdmf/blob/8ddae22504cfd07db394a549b586e395c83dc8ad/src/hdmf/utils.py#L423-L436 The name of the docval argument and the argument of the ``__eq__`` function need to match, e.g, if I change the docval argname it ``something`` you would get ``TypeError: __eq__() got an unexpected keyword argument 'something'``

rly (approved):

PR #188, opened 2019-10-25
Updated DynamicTableRegion.__getitem__ to return dict with table name…

oruebel:
## Motivation

Currently, in a DynamicTableRegion column in a DynamicTable the references are resolved directly to the selected rows of the table. For the user, this looses the information about the table the the column points to. This is not a big problem when we only have a single level of referencing, but gets complicated when we have a hierarchy of tables. As such, I propose here to instead of returning the ``rows``, to return a dict with ``{'table_name': rows}``.

For example, in the context of https://github.com/oruebel/ndx-icephys-meta we have a hierarchical set of tables for Conditions, Runs, .... etc. Currently this results in the following for the Conditions table:

Screen Shot 2019-10-25 at 9 24 29 AM

With this change, this will look like this instead

Screen Shot 2019-10-25 at 9 25 41 AM

I.e., instead of a hierarchical list without any information about where each nesting came from, we'd have a hierarchical dict where the key indicates name of the table and the values are the selected rows. I think from a user perspective this would be more usable.

I'm sure this will also require some changes in PyNWB, but I wanted to put this out as a proposal to see what you think @rly @bendichter @ajtritt . I put the PR in draft mode, mainly because it should not be merged until we have a matching PR on PyNWB.

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -618,13 +618,15 @@ class DynamicTableRegion(VectorData):
     def __getitem__(self, key):
         # treat the list of indices as data that can be indexed. then pass the
         # result to the table to get the data
+        re = None
         if isinstance(key, tuple):
             arg1 = key[0]
             arg2 = key[1]
-            return self.table[self.data[arg1], arg2]
+            re = self.table[self.data[arg1], arg2]
         elif isinstance(key, (int, slice)):
             if isinstance(key, int) and key >= len(self.data):
                 raise IndexError('index {} out of bounds for data of length {}'.format(key, len(self.data)))
-            return self.table[self.data[key]]
+            re = self.table[self.data[key]]
         else:
             raise ValueError("unrecognized argument: '%s'" % key)
+        return {self.table.name: re}
--- a/tests/unit/common/test_table.py
+++ b/tests/unit/common/test_table.py
@@ -208,7 +208,7 @@ class TestDynamicTable(unittest.TestCase):
         table = self.with_columns_and_data()
 
         dynamic_table_region = DynamicTableRegion('dtr', [0, 1, 1], 'desc', table=table)
-        fetch_ids = [x[1] for x in dynamic_table_region[:3]]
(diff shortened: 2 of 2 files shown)


ajtritt:
This is going to make building any general purpose tool for traversing tables difficult, since indexing into table elements will now happen with table names, rather than numeric indices. 

The problem you're addressing is introduced when you convert the DynamicTable object to a pandas DataFrame. Table identity should not be lost when indexing directly on DynamicTable objects.

oruebel:
> The problem you're addressing is introduced when you convert the DynamicTable object to a pandas DataFrame

The problem seems to be introduced once we select elements from the DynamicTableRegion, i.e., select a set of rows, rather than being specific to the Pandas conversion.

> This is going to make building any general purpose tool for traversing tables difficult, since indexing into table elements will now happen with table names, rather than numeric indices.

I'm not sure I understand how this change affects this, since the result in the dict from __getitem__  is still be the same. If we don't want the key to be the table name, then maybe we could do {'data': rows, 'name': table_name, 'index': row_index}, which would give you the data from the rows as they are returned right now in the `data` key and in addition the name of the table as well as the indicies of the rows (both of which are currently missing).

oruebel:
Thinking about it some more, this is really not the right solution. The problem is not that I would want to know the name of the table, but also the column names etc. I.e., rather than getting the the resolved data rows, I think I'd want a new DynamicTable that is a subset of the previous one. I'll close the PR for now until I have a better solution.

ajtritt:
> I'm not sure I understand how this change affects this, since the result in the dict from getitem is still be the same.


Indexing into a subelement of the returned rows requires knowing the table name ahead of time. Since table name is a value (something that changes), and you’re storing that value in a key, indexing is not consistent. 

> If we don't want the key to be the table name, then maybe we could do {'data': rows, 'name': table_name, 'index': row_index},


That does solve the issue of storing value in a key, but still does not seem generalizable. 

It seems like the problem originates from trying to pack multiple rows into single rows. Is that correct? Or is it just a problem of ambiguity in the name of a table? 

Could

Issue #136, opened 2019-08-29
Refactor DataChunkIterator to not read first chunk on init


rly:
DataChunkIterator reads first chunk on init in order to determine properties such as maxshape and dtype. This can break certain setups where there is no first chunk, such as acquisition systems streaming to a DataChunkIterator.

rly:
Closed for now as this is partially addressed by `GenericDataChunkIterator` and has not been requested outside of the dev team.

PR #189, opened 2019-10-28
Avoid reading first chunk of DataChunkIterator on init if maxshape and dtype are specified

rly:
If a `DataChunkIterator` is constructed with an iterator for `data`, then the first chunk is read on `init` in order to help determine the maxshape and dtype. This is unnecessary if both maxshape and dtype are provided to `init`, so we should avoid reading the first chunk. This helps with #136. 

It looked like someone had already written in this condition years ago but left it commented out. This PR uncomments that condition.

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -136,7 +136,11 @@ class DataChunkIterator(AbstractDataChunkIterator):
 
     @docval(*__docval_init)
     def __init__(self, **kwargs):
-        """Initialize the DataChunkIterator"""
+        """Initialize the DataChunkIterator.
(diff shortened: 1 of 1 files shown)


ajtritt (dismissed):
Seems like a good idea to bring that check back. Is there a reason why we don't move setting of maxshape to the getter i.e. only attempt to read the next chunk for setting maxshape when it is read ?

rly:
> Is there a reason why we don't move setting of maxshape to the getter i.e. only attempt to read the next chunk for setting maxshape when it is read ?

That seems like a reasonable thing to do. I'll refactor it.

ajtritt (approved):

PR #211, opened 2019-11-23
Clean up tests

### LINKED ISSUE #198: Add assertRaisesWith testing helper
Python's builtin unittest module provides an `assertRaises` and `assertRaisesRegexp`, but not a method to check for an exact match. It would help simplify our test assertions to add an `assertRaisesWith` method and `assertWarnsWith` method to test for an exact match on the exception/warning message, similar to https://gitweb.torproject.org/stem.git/commit/?id=1c030e3

```python
def assertRaisesWith(self, exc_type, exc_msg, func, *args, **kwargs):
    """
    Asserts the given invocation raises the expected exception. This is
    similar to unittest's assertRaises and assertRaisesRegexp, but checks
    for an exact match.
    """

    return self.assertRaisesRegexp(exc_type, '^%s$' % re.escape(exc_msg), func, *args, **kwargs)
```

rly:
- Some tests throw warnings. Catch those in tests. 
- Add some tests for `src/hdmf/utils.py`.  
- Refactor `src/hdmf/utils.py` 
- Remove some python 2 legacy code (still lots). 
- Create `hdmf.testing` module with subclass of `unittest.TestCase`
- Fix #198 by implementing `assertRaisesWith` and `assertWarnsWith`
- Move round-trip testing code `assertContainerEqual` and `TestMapH5RoundTrip` from PyNWB and tests/unit/common to the new testing module


src/hdmf/utils.py
```diff
@@ -55,11 +53,7 @@ def __type_okay(value, argtype, allow_none=False):
             return __is_bool(value)
         return argtype in [cls.__name__ for cls in value.__class__.__mro__]
     elif isinstance(argtype, type):
-        if argtype == six.text_type:
-            return isinstance(value, six.text_type) or isinstance(value, six.string_types)
-        elif argtype == str:
-            return isinstance(value, six.string_types)
```
oruebel, line 61: It looks like this change removes the case of strings alltogether (not just the six.text_type case). Is this intentional?

rly: if argtype == str, then new line 62 will run: `return isinstance(value, argtype)` which will catch the case of strings.

oruebel (approved):

PR #213, opened 2019-11-27
Purge support for python 2 to clean up code, remove six dependency

rly:
This PR is not high priority by any means, but since I had started refactoring, I might as well finish the job. To clean up the code and remove the dependency on the six package, I purged all uses of six, e.g., `text_type`, `binary_type`, `string_types`, `integer_types`, `raise_from`, `with_metaclass`, etc.

Also:
- removed any imports of future
- updated `.gitignore` to use absolute paths which prevents my code editor from ignoring the `src/hdmf/build/` directory. 
- use python 3 style classes (implicit inheritance from `object`) and `super()`


rly:
@ajtritt can we merge this? this will make the diff for my dimension scales PR significantly easier to follow. thanks

ajtritt (approved):

oruebel (approved):

Issue #212, opened 2019-11-25
type checking for dynamically loaded filters, and working with hdf5plugin


d-sot:
Loading external filters in hdf5 works in h5py. I tried it with H5DataIO but it gives a type checking error. For lz4 filter example:

```
TypeError: incorrect type for 'compression' (got 'int', expected 'str or bool')
```

The registered filter ID in hdf5 for external filters is an int passed as the compression parameter. Their compression options can be multiple parameters, With h5py it is passed as a tuple.

I came across this looking to use blosc compressors for their performance. They're available on pip recently https://pypi.org/project/hdf5plugin/ . They're also available on HDF5 website as a separate package on their download page.

for:
```
# data is a numpy array
H5DataIO(data, chunks=(4,4,4,4),compression=32001,compression_opts=(0, 0, 0, 0, 1, True, 1) )
```

oruebel:
Adding ``int`` to the allowed types for the compression parameter on H5DataIO here

https://github.com/hdmf-dev/hdmf/blob/8602c509d8a9e7d31b6ce0ef6d003232744a6897/src/hdmf/backends/hdf5/h5_utils.py#L361-L364

should probably fix this issue.

d-sot:
Thanks. Adding just int, it first gave an error:
```
File "hdmf/backends/hdf5/h5_utils.py", line 296, in __init__
    str(self.__iosettings['compression']))
ValueError: 32001 compression not support by this version of h5py.
```
just changing the line 486 https://github.com/hdmf-dev/hdmf/blob/8602c509d8a9e7d31b6ce0ef6d003232744a6897/src/hdmf/backends/hdf5/h5_utils.py#L485-L488 to 
```
    return filter in h5py_filters.encode or filter in hdf5plugin.FILTERS.values()
```
fixed it, and it works now.

bendichter:
Nice work! If we wanted to integrate this we'd have to add `hdf5plugins` as a
requirement.

rly:
It looks like `h5py_filters.encode` does not include dynamically loaded filters. Perhaps we do not need the function `filter_available`, and we can let the h5py library raise an error when trying to create a dataset with an invalid compression value.

oruebel:
> ... we can let the h5py library raise an error when trying to create a dataset with an invalid compression value

[...]

PR #215, opened 2019-11-29
Allow passing integer filter ID for dynamically loaded filters

d-sot:
This is regarding issue #212. Following @oruebel suggestion, ```H5DataIO``` accepts filter ID integers for ```compression```, and if ```allow_plugin_filters``` is passed as True, ```filter_available``` checks for and confirms that it can both encode and decode. Otherwise filter_available returns False for integer filters.
The ```h5py.filters.h5z interface``` is used to check the filter:
```
h5py_filters.h5z.filter_avail(filter) # returns True if the filter is available
h5py_filters.h5z.FILTER_CONFIG_ENCODE_ENABLED # == 1, constant
h5py_filters.h5z.FILTER_CONFIG_DECODE_ENABLED # == 2, constant
h5py_filters.h5z.get_filter_info(filter) # returns int in [1,2,3] matching the FILTER_CONFIG constants
```

## Motivation

[hdf5plugin](https://pypi.org/project/hdf5plugin/ ) allows easy access to high performance blosc compressors.

## Checklist


bendichter:
@d-sot this looks great! It would be great to have a test for this new feature, but on the other hand I don't see how we would test this without adding hdf5plugins as a dependency. Any idea? @oruebel do you think it would make sense to add hdf5plugins as a testing requirement?

oruebel:
@bendichter I don't think we necessarily need hdf5plugins to test this, as all we need for testing is some custom filter. I agree it would be nice to have a test. Otherwise the code looks good. @d-sot  can you add a test?

oruebel (approved):

PR #216, opened 2019-12-04
allow scalars in Data constructor

ajtritt:
# Motivation

The `Data` class constructor should accept scalars as data.

## How to test the behavior?
```python
from hdmf.container import Data
d  = Data("test_data", "a string scalar")
```

## Checklist


src/hdmf/utils.py
```diff
@@ -9,7 +9,7 @@
 
 __macros = {
     'array_data': [np.ndarray, list, tuple, h5py.Dataset],
-    'scalar_data': [str, int, float],
+    'scalar_data': [str, int, float, bytes],
```
oruebel, line 9: do we also need to allow ``bool`` as a scalar?

src/hdmf/data_utils.py
```diff
@@ -624,7 +628,11 @@ def __len__(self):
         return len(self.data)
 
     def __bool__(self):
```
oruebel, line 626: It would be nice to add a docstring to describe what the logic behind ``__bool__`` is. Is this for checking whether DataIO holds any data?

src/hdmf/common/__init__.py
```diff
@@ -193,3 +194,12 @@ def validate(**kwargs):
     builder = io.read_builder()
     validator = ValidatorMap(io.manager.namespace_catalog.get_namespace(name=namespace))
     return validator.validate(builder)
+
+
+@docval(*get_docval(HDF5IO.__init__), is_method=False)
+def get_hdf5io(**kwargs):
```
oruebel, line 200: What is ``get_hdf5io`` used for?

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1112,3 +1115,9 @@ def __rec_get_ref(self, l):
             else:
                 ret.append(elem)
         return ret
+
+    @classmethod
+    @docval(*get_docval(H5DataIO.__init__))
+    def set_dataio(cls, **kwargs):
```
oruebel, line 1149: Do we have tests for the ``set_dataio`` functions?

ajtritt:
addressed by #390

PR #217, opened 2019-12-06
Improve testing of containers

rly:
`assertContainerEqual` does not check that the name, container_source, parent, or children match - just that the other fields (which does not include name!) matches. This PR fixes that.


oruebel (dismissed):

rly:
A new change:
The unittest framework automatically tries to run abstract `TestCase` classes as tests. These end up getting skipped. Instead, I converted those classes to abstract mix-in classes. Actual test classes inherit from the mix-in class to use the mix-in method definitions, and they will need to override unimplemented methods. See https://stackoverflow.com/questions/4566910/abstract-test-case-using-python-unittest

This design pattern is now used in several places, most notably the change from `TestMapRoundTrip` to `TestH5RoundTripMixin`

oruebel (approved):

PR #220, opened 2019-12-13
ObjectMapper.construct: Output the exception message as well

t-b:
This is much more helpful for debugging.


oruebel (approved):

rly:
@t-b the underlying error message is nested (`raise ex2 from ex1`) and should still be output as well. Is it necessary to copy it into ex2's error message?

t-b:
@rly Yes the error messages can be a bit verbose for the case you are describing. Is that a problem?

rly:
@t-b No, it is fine. I was just curious what the rationale was given that the inner error message should be printed to stderr as well.

PR #192, opened 2019-11-06
[WIP] Add dimension scales/coordinates

rly:
Add dimension scales/coordinates to support direct association of datasets with dimensions of other datasets within a group. See https://github.com/NeurodataWithoutBorders/pynwb/issues/626

Very much a work in progress. I am also taking the opportunity to comment and refactor `src/hdmf/build/map.py`. `ObjectMapper` is now in its own 1000-line file. Unfortunately, the refactoring obscures the changes I made in the `ObjectMapper`.

PR #221, opened 2019-12-18
Split ObjectMapper from map.py

rly:
This PR moves `ObjectMapper` out of map.py and into objectmapper.py. This change was initially in #192 but was suggested to be made its own independent PR so that the changes are more readable. 

It also adds documentation and changes some variable names to be easier to understand.

There is one noteworthy, necessary change though: in order to prevent import loops between map.py and objectmapper.py, I changed the default value for the argument 'mapper_cls' of `TypeMap.__init__` from `ObjectMapper` to `None`. Then, within `__init__`, if `mapper_cls` is None, `ObjectMapper` is imported and `mapper_cls` is set to `ObjectMapper`, so backwards compatibility is maintained. Not the cleanest, but since this was the only use of `ObjectMapper` in the remaining code of map.py, I think moving the `ObjectMapper` class to its own file is still worth it.

### DIFF
--- a/src/hdmf/build/__init__.py
+++ b/src/hdmf/build/__init__.py
@@ -5,6 +5,7 @@ from .builders import ReferenceBuilder
 from .builders import RegionBuilder
 from .builders import LinkBuilder
 
-from .map import ObjectMapper
+from .objectmapper import ObjectMapper
+
 from .map import BuildManager
 from .map import TypeMap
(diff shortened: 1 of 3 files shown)


rly:
PyNWB will have to be updated so that this call:
```python
from hdmf.build.map import ObjectMapper, TypeMap
```
becomes
```python
from hdmf.build import ObjectMapper, TypeMap
```

ajtritt (approved):

PR #223, opened 2019-12-29
Schema validator

### LINKED ISSUE #2: schema yaml validator
We should have a tool that checks the yaml files to make sure that schema specification files are correct. In other words, a meta-schema validator that checks the core and any extension schema files. This will be particularly useful for extension sharing, so we could run submitted extensions through an automated validator to ensure that they are at least syntactically correct.

https://json-schema.org/ would be a good tool for this.

## Checklist

bendichter:
fix #2


bendichter:
running into this issue: https://github.com/Julian/jsonschema/issues/313

bendichter:
fix with this work-around: https://github.com/Julian/jsonschema/issues/313#issuecomment-269361225

bendichter:
I tried using `"allOf"` to factor out some common dependencies between types, but that did not mesh well with the `"additionalProperties": false` constraint, so I unfactored them.

PR #229, opened 2020-01-02
try using numpy to determine shape

### LINKED ISSUE #228: shape of VectorIndex
## Description

When `docval` checks the shape of `VectorIndex`, it returns the shape of the referenced table. It should instead check the shape of `VectorIndex.data`
## Steps to Reproduce

```python
from hdmf.common.table import VectorIndex, VectorData
from hdmf.utils import docval


data = [[1., 1., 1.], [2., 2., 2.], [1., 2., 1.]]
data_vect = VectorData(name='vec', description='desc', data=data)

indexes = [2, 3]
data_ind = VectorIndex(name='vec_index', data=indexes, target=data_vect)

@docval({'name': 'vec_index', 'type': VectorIndex, 'doc': 'test', 'shape': (None,)},
        is_method=False)
def test_func(**kwargs):
    pass

test_func(vec_index=data_ind)
```

```
  File "/Users/bendichter/dev/ndx-point-cloud-table/test_vectorindex.py", line 17, in <module>
    test_func(vec_index=data_ind)
  File "/Users/bendichter/anaconda3/lib/python3.7/site-packages/hdmf/utils.py", line 451, in func_call
    raise_from(ExceptionType(msg), None)
  File "<string>", line 3, in raise_from
ValueError: incorrect shape for 'vec_index' (got '(2, 2, 3)', expected '(None,)')
```

`np.shape(data_ind)` and `np.shape(data_ind.data)` both yield the correct shape of (2,). Presumably this mistaken shape is due to a `[:]` call on `vec_index` within docval, which not only yields the wrong shape, but also reads the data unnecessarily. This may be a contributing factor to the performance issues we are seeing for @oruebel's icephys extension.

## Environment

    Python Executable: Conda
    Python Version: Python 3.7
    Operating System: macOS

## Checklist

bendichter:
## Motivation

`np.shape` determines the shape of an array by reading the data in place
without copying, so it should be more efficient than loading the data
and more robust to cases where no data load is enforced.


src/hdmf/utils.py
```diff
@@ -634,19 +634,19 @@ def __get_shape_helper(local_data):
...
     if hasattr(data, 'maxshape'):
         return data.maxshape
-    elif hasattr(data, 'shape'):
+    if hasattr(data, 'shape'):
         return data.shape
-    elif isinstance(data, dict):
-        return None
-    elif hasattr(data, '__len__') and not isinstance(data, (str, bytes)):
+    if isinstance(data, dict):
+        return
```
rly: Style comment: I would prefer being explicit about the return value
```suggestion
        return None
```

src/hdmf/utils.py
```diff
@@ -634,19 +634,19 @@ def __get_shape_helper(local_data):
...
+    if hasattr(data, '__len__') and not isinstance(data, (str, bytes)):
         if not strict_no_data_load or (isinstance(data, list) or isinstance(data, tuple) or isinstance(data, set)):
             return __get_shape_helper(data)
-        else:
-            return None
-    else:
-        return None
```
rly, line 775: Style comment: I would prefer to be explicit with return values for functions where a return value is expected, so I would add `return None` here.

rly (changes requested):

src/hdmf/utils.py
```diff
@@ -634,19 +634,19 @@ def __get_shape_helper(local_data):
             if len(local_data) and not isinstance(local_data[0], (str, bytes)):
                 shape.extend(__get_shape_helper(local_data[0]))
         return tuple(shape)
+
+    shape = np.shape(data)
+    if shape:
+        return shape
     if hasattr(data, 'maxshape'):
```
rly, line 770: To maintain existing behavior where maxshape is checked before shape, this check should go before `np.shape`

src/hdmf/utils.py
```diff
@@ -634,19 +634,19 @@ def __get_shape_helper(local_data):
...
         return data.maxshape
-    elif hasattr(data, 'shape'):
+    if hasattr(data, 'shape'):
```
rly, line 772: This is redundant with `np.shape`. See https://github.com/numpy/numpy/blob/v1.18.4/numpy/core/fromnumeric.py#L1903-L1946

rly: never mind

src/hdmf/utils.py
```diff
@@ -634,19 +634,19 @@ def __get_shape_helper(local_data):
...
         return data.shape
-    elif isinstance(data, dict):
-        return None
-    elif hasattr(data, '__len__') and not isinstance(data, (str, bytes)):
+    if isinstance(data, dict):
+        return
+    if hasattr(data, '__len__') and not isinstance(data, (str, bytes)):
         if not strict_no_data_load or (isinstance(data, list) or isinstance(data, tuple) or isinstance(data, set)):
```
rly, line 648: Should this have been `strict_no_data_load and (isinstance...`?

rly:
@bendichter I made some changes to your PR, including adding needed tests for `get_data_shape`, but I could not come up with a use case / MWE where the use of `np.shape` is better than what we currently have. Could you give an example?

bendichter:
We decided to remove the changes to the shape check but keep the tests

PR #230, opened 2020-01-03
rmv datetime64 from supported dtype strings

bendichter:
## Motivation

"datetime64" is currently defined as a synonym for "isodatetime" in `hdmf.spec`, but is not a supported type in the schema-language document. We could add it to the schema language, as I am currently proposing for "bytes", "short", "uint64", and "datetime" [here](https://github.com/NeurodataWithoutBorders/nwb-schema/pull/382), but in this case I would rather remove it. `datetime64` is a numpy datatype for representing datetimes. To me it does not make sense to have this be synonymous with the standardized ISO date-time string. As far as I can tell we are not using this type, so it would require no other changes.

### DIFF
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -37,7 +37,7 @@ class DtypeHelper():
             'object': ['object'],
             'region': ['region'],
             'numeric': ['numeric'],
-            'isodatetime': ["isodatetime", "datetime", "datetime64"]
+            'isodatetime': ["isodatetime", "datetime"]
         }
 
     # List of recommended primary dtype strings. These are the keys of primary_dtype_string_synonyms


bendichter:
@ukos-git, it looks like you added this. Do you need it? Would you mind if we remove it?

ukos-git:
The `datetime64` data type of numpy/panda is very limited. I initially thought it would be a good idea to store dates in this type and it would have been also easier to do. We are only using the `isodatetime` type and I think it is good to delete `datetime64` as long as noone requires full numpy compatibility.

bendichter:
Thanks @ukos-git, we came to the same conclusion. It terms of flexibility and timezone support, ISO fits our needs better.

oruebel (approved):

PR #236, opened 2020-01-06
Enh/remove get name func

oruebel:
## Motivation

The classmethod ``HDF5IO.__get_name`` is a one-line function that is used only once. To simplify the code, this PR removes the method in favor of calling the corresponding call directly. 

## Checklist


ajtritt:
This one-line function defines the rule for what a cached specification source file is named, based on the original source. I'd prefer to just add better documentation rather than toss it out.

oruebel:
Sounds good. I'll add a docstring for the function instead then.

PR #238, opened 2020-01-07
Refactor use of deepcopy by docval

### LINKED ISSUE #237: docval uses a significant portion of runtime
From @oruebel : `docval` calls `deepcopy` several thousand times in his test run. Because `deepcopy` is recursive, this results in millions of calls to `deepcopy`. Looking at the code, it does not look like such use of `deepcopy` is necessary, but the tests break if the `deepcopy` is removed or just changed to `copy`.

rly:
Fix #237 

This PR improves runtime of hdmf tests by about 20% and runtime of pynwb tests by about 25%. 

The docval list of dictionaries does not need to be deepcopied on every call to a function with docval. The list of dictionaries consists of just strings, types, tuples of strings and types, and critically here, default values. If the default values are not properly copied when used to set the values of missing arguments, missing arguments will all share the same list/dict default value, which is bad.

Rather than deepcopy the whole list of dictionaries, just deepcopy the default value when it is used to set the value of a missing argument. 

This PR also removes a few other unnecessary/redundant calls to `deepcopy` regarding docval.

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -225,7 +225,7 @@ def __parse_args(validator, args, kwargs, enforce_type=True, enforce_shape=True,
                 ret[argname] = args[argsi]
                 argsi += 1
             else:
-                ret[argname] = arg['default']
+                ret[argname] = _copy.copy(arg['default'])
             argval = ret[argname]
             if enforce_type:
                 if not __type_okay(argval, arg['type'], arg['default'] is None):
@@ -411,7 +411,6 @@ def docval(*validator, **options):
     rtype = options.pop('rtype', None)
     is_method = options.pop('is_method', True)
     allow_extra = options.pop('allow_extra', False)
-    val_copy = __sort_args(_copy.deepcopy(validator))
 
     def dec(func):
         _docval = _copy.copy(options)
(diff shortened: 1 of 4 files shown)


src/hdmf/utils.py
```diff
@@ -411,7 +411,6 @@ def foo(self, **kwargs):
     rtype = options.pop('rtype', None)
     is_method = options.pop('is_method', True)
     allow_extra = options.pop('allow_extra', False)
-    val_copy = __sort_args(_copy.deepcopy(validator))
```
rly, line 414: The sorting performed on this line is unnecessary given sorting in lines 429-432 (old line numbers) below. The deepcopy is also unnecessary given the other changes.

src/hdmf/utils.py
```diff
@@ -225,7 +224,7 @@ def __parse_args(validator, args, kwargs, enforce_type=True, enforce_shape=True,
                 ret[argname] = args[argsi]
                 argsi += 1
             else:
-                ret[argname] = arg['default']
+                ret[argname] = _copy.deepcopy(arg['default'])
```
rly, line 227: This is one of the key changes -- only default values are deep-copied, and only when they are used.

src/hdmf/utils.py
```diff
@@ -436,7 +423,7 @@ def dec(func):
             def func_call(*args, **kwargs):
                 self = args[0]
                 parsed = __parse_args(
-                            _copy.deepcopy(loc_val),
+                            loc_val,
```
rly, line 426: This is one of the key changes. Previously, this list of dictionaries (the docval specification/validator) was deepcopied on every call to a function with docval.

oruebel (approved):

PR #239, opened 2020-01-07
Enh/add dtr to dataframe

oruebel:
## Motivation
* Add method to convert DynamicTableRegion as a whole to a dataframe to help with performance for the ICEphys extension
* Improve testing of DynamicTableRegion

## Checklist


rly (approved):

tests/unit/common/test_table.py
```diff
@@ -336,6 +319,128 @@ def setUpContainer(self):
...
+    def test_dynamic_table_region_shape(self):
+        table = self.with_columns_and_data()
+        dynamic_table_region = DynamicTableRegion('dtr', [0, 1, 2, 3, 4], 'desc', table=table)
+        self.assertTupleEqual(dynamic_table_region.shape, (5, 3))
+
+    def test_dynamic_table_region_to_dataframe(self):
+        table = self.with_columns_and_data()
+        dynamic_table_region = DynamicTableRegion('dtr', [0, 1, 2, 2], 'desc', table=table)
+        res = dynamic_table_region.to_dataframe()
+        self.assertListEqual(res.index.to_list(), [0, 1, 2, 2])
```
rly: Pandas 0.24.0 introduced the alias `to_list` for `tolist`. For compatibility with earlier versions of pandas, let's use `tolist` instead of `to_list`. 

https://stackoverflow.com/questions/57854228/pandas-dataframe-index-to-list-vs-tolist

rly (changes requested):

tests/unit/common/test_table.py
rly: We currently list pandas 0.23 as a minimum requirement.

oruebel: Sounds good. I just updated the tests accordingly to use ``tolist`` instead of ``to_list``.

rly (approved):

PR #246, opened 2020-01-09
Move BuildManager, TypeMap to manager.py, have map.py keep old imports

rly:
In HDMF 1.5.0, the `ObjectMapper` class was moved from `map.py` to `objectmapper.py`. This broke any code that used `from hdmf.build.map import ObjectMapper`. Although the preferred import statement is `from hdmf.build import ObjectMapper`, PyNWB and possibly other codes contained the former statement. 

In order to prevent a backwards compatibility breaking change in HDMF 1.5, this PR refactors `map.py` to include the `ObjectMapper` class. To do this, other classes in `map.py` had to be refactored into `manager.py` to prevent circular imports. Now, `from hdmf.build.map import ObjectMapper` still works and raises a `PendingDeprecationWarning`. The dev branch of PyNWB has already been updated to do the proper imports.

`map.py` can be removed from HDMF in version 2.0.


rly:
@ajtritt The new file with the `BuildManager` and `TypeMap` is called `manager.py`. Happy to change it to `typemap.py` or something else.

ajtritt (approved):

Issue #245, opened 2020-01-08
Test failed hdmf 1.5.0


lbazan:
+ /usr/bin/python3 test.py -v
======================================================================
2020-01-08 21:01:18,398 - INFO - running hdmf unit tests
EEEEEEEEEEEEEEEEEEEEEEEEEEE                             
[0;0;36mStatus: 
    [0;0;31mError: 27
[0;0;36mDescription:
[0;0;36mSummary: 
    +------------------------------------------+-------+------+------+-------+
    |           Test group/Test case           | Count | Pass | Fail | Error |
    +------------------------------------------+-------+------+------+-------+
    | unittest.loader._FailedTest              | 27    | 0    | 0    | 27    |
    | ---------------------------------------- | ----  | ---- | ---- | ----  |
    | Total                                    | 27    | 0    | 0    | 27    |
    +------------------------------------------+-------+------+------+-------+
    unittest.loader._FailedTest
        +-----------------------------+--------+-------------+
        |          Test name          | Output |    Status   |
        +-----------------------------+--------+-------------+
        | test_1_1_0                  |        | [0;0;31merror       |
        | test_io_build_builders      |        | [0;0;31merror       |
        | test_io_manager             |        | [0;0;31merror       |
        | test_io_map                 |        | [0;0;31merror       |
        | test_io_map_data            |        | [0;0;31merror       |
        | test_common                 |        | [0;0;31merror       |
        | test_sparse                 |        | [0;0;31merror       |
        | test_table                  |        | [0;0;31merror       |
        | test_attribute_spec         |        | [0;0;31merror       |
        | test_dataset_spec           |        | [0;0;31merror       |
        | test_dtype_spec             |        | [0;0;31merror       |
        | test_group_spec             |        | [0;0;31merror       |
        | test_load_namespace         |        | [0;0;31merror       |
        | test_ref_spec               |        | [0;0;31merror       |
        | test_spec_catalog           |        | [0;0;31merror       |
        | test_spec_write             |        | [0;0;31merror       |
        | test_container              |        | [0;0;31merror       |
        | test_io_hdf5                |        | [0;0;31merror       |
        | test_io_hdf5_h5tools        |        | [0;0;31merror       |
        | test_query                  |        | [0;0;31merror       |
        | test_core_DataChunk         |        | [0;0;31merror       |
        | test_core_DataChunkIterator |        | [0;0;31merror       |
        | test_core_DataIO            |        | [0;0;31merror       |
[...]

PR #248, opened 2020-01-09
Handle np.array(DataIO) for numpy in py38

rly:
Partially address #245


src/hdmf/data_utils.py
```diff
@@ -626,6 +626,8 @@ def __bool__(self):
 
     def __getattr__(self, attr):
         """Delegate attribute lookup to data object"""
+        if attr == '__array_struct__':  # np.array() checks __array__ or __array_struct__ attribute dep. on version
```
oruebel: Does this mean delegation of ``__array__`` is allows but delegation of ``__array_struct__`` is not? I.e., depending on the version of numpy we allow conversion of DataIO to a numpy array or not? Could you clarify what the problem with the ``__array_struct__`` method is?

oruebel (dismissed):

src/hdmf/data_utils.py
rly: My mistake. The error should only be thrown if the object is not self.valid. I have added that back.

rly:
To clarify: `np.array()` with python 3.8 was calling `self.getattr('__array_struct__')` or possibly `self.hasattr('__array_struct__')`. I could not override the method `__array_struct__` with the below: 
```python
def __array_struct__(self):
        if not self.valid:
            raise InvalidDataIOError("Cannot convert data to array. Data is not valid.")
        return self.__array_struct__()
```
I would get the error: `'ValueError: invalid __array_struct__'` on calls to `np.array()` that worked otherwise. Maybe there is an issue with name mangling. So instead of overriding the `__array_struct__` method, I check for the attribute argument `__array_struct__` in the `getattr` method.

oruebel (approved):

Issue #256, opened 2020-01-10
1.5.0 tests.unit.common.test_table.TestDynamicTableRoundTrip fails on some platforms


yarikoptic:
built neurodebian package for hdmf, and it failed that test on all but most recent debian sid/bullseye:
```
hdmf_1.5.1-1~nd80+1_amd64.build FAILED  0:15.11 real, 11.12 user, 4.49 sys, 101728 out
hdmf_1.5.1-1~nd90+1_amd64.build FAILED  0:49.21 real, 39.93 user, 13.79 sys, 808928 out
hdmf_1.5.1-1~nd100+1_amd64.build        FAILED  1:04.23 real, 59.61 user, 16.91 sys, 864600 out
hdmf_1.5.1-1~nd110+1_amd64.build        OK      1:24.23 real, 79.49 user, 21.88 sys, 1183048 out
hdmf_1.5.1-1~nd+1_amd64.build   OK      1:32.47 real, 85.06 user, 25.61 sys, 1217472 out
hdmf_1.5.1-1~nd14.04+1_amd64.build      FAILED  0:54.43 real, 40.21 user, 18.00 sys, 112232 out
hdmf_1.5.1-1~nd16.04+1_amd64.build      FAILED  1:24.55 real, 64.63 user, 25.81 sys, 734224 out
hdmf_1.5.1-1~nd18.04+1_amd64.build      FAILED  1:35.20 real, 79.69 user, 29.11 sys, 790376 out
hdmf_1.5.1-1~nd18.10+1_amd64.build      FAILED  1:43.95 real, 90.95 user, 31.98 sys, 1023984 out
hdmf_1.5.1-1~nd19.04+1_amd64.build      FAILED  1:38.25 real, 83.20 user, 30.04 sys, 832480 out

```
the failure seems to be the same:
```
======================================================================
ERROR: test_roundtrip (tests.unit.common.test_table.TestDynamicTableRoundTrip)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/hdmf-1.5.1/.pybuild/cpython3_3.7_hdmf/build/hdmf/testing/testcase.py", line 164, in test_roundtrip
    self.read_container = self.roundtripContainer()
  File "/build/hdmf-1.5.1/.pybuild/cpython3_3.7_hdmf/build/hdmf/testing/testcase.py", line 178, in roundtripContainer
    self.writer.write(self.container, cache_spec=cache_spec)
[...]

PR #258, opened 2020-01-11
Streamline CI and add testing of min requirements

rly:
h5py needs to be at minimum version 2.9. In case there are other dependency versions that are required for HDMF to work, this PR adds minimum versions for all requirements in setup.py. All tests succeed with the minimum requirements.

This PR also:
- adds unit testing and Sphinx gallery testing on Python 3.5 with the minimum requirements
- fixes Python 3.8 testing on CircleCI (was running Python 3.7)
- removes coverage testing for Python 3.5-3.7 on Windows, Mac, and Linux. Coverage testing and submission to codecov remains for Python 3.8. Coverage testing is not necessary for all versions of Python (especially now that Python 2 support is deprecated), and it slows down CI because it involves installing another environment.
- removes legacy CI crud
- removes "-r requirements.txt" from "requirements-dev.txt" and "requirements-doc.txt". Now, in order to install the dev environment, one must use `pip install -r requirements-dev.txt -r requirements.txt`. Both the installation documentation and tox specification already specify this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -365,6 +365,8 @@ workflows:
           <<: *no_filters
       - gallery38:
           <<: *no_filters
+      - gallery35-min-req:
+          <<: *no_filters
       - deploy-dev:
           requires:
             - flake8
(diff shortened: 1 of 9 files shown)


rly:
See #256 for more information. @oruebel, @ajtritt what do you think of this PR?

ajtritt (dismissed):

ajtritt (approved):

PR #269, opened 2020-01-18
Enh/read scalar dset

rly:
@t-b says that the MIES software cannot write scalar datasets, which are represented by h5py as having shape `()`, so MIES instead writes 1-element datasets, which have shape `(1, )`. HDMF reads datasets as `h5py.Dataset`s, i.e., it does not read the data into memory. Therefore, docval fails where a string/int/float is expected and an `h5py.Dataset` is provided.

This PR makes HDMF read all datasets that have shape `(1, )` into memory as if they were scalars with shape `()`.

See original PR #78.

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -465,10 +465,7 @@ class HDF5IO(HDMFIO):
                 kwargs["data"] = scalar
         elif ndims == 1:
             d = None
-            if h5obj.shape == (1, ):
-                # read scalar (1-element) dataset into memory
-                d = h5obj[0]
-            elif h5obj.dtype.kind == 'O':
+            if h5obj.dtype.kind == 'O':
                 elem1 = h5obj[0]
                 if isinstance(elem1, (str, bytes)):
                     d = h5obj
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -447,11 +447,8 @@ class HDF5IO(HDMFIO):
             name = str(os.path.basename(h5obj.name))
         kwargs['source'] = h5obj.file.filename
         ndims = len(h5obj.shape)
-        if ndims == 0 or h5obj.shape == (1, ):
-            if ndims == 0:  # read scalar dataspace
-                scalar = h5obj[()]
-            else:  # read shape (1, ) dataset as scalar
-                scalar = h5obj[0]
+        if ndims == 0:                                       # read scalar
+            scalar = h5obj[()]
             if isinstance(scalar, bytes):
                 scalar = scalar.decode('UTF-8')
 
@@ -468,7 +465,10 @@ class HDF5IO(HDMFIO):
                 kwargs["data"] = scalar
         elif ndims == 1:
             d = None
-            if h5obj.dtype.kind == 'O':
+            if h5obj.shape == (1, ):
+                # read scalar (1-element) dataset into memory
+                d = h5obj[0]
+            elif h5obj.dtype.kind == 'O':
                 elem1 = h5obj[0]
                 if isinstance(elem1, (str, bytes)):
                     d = h5obj
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -465,7 +465,10 @@ class HDF5IO(HDMFIO):
                 kwargs["data"] = scalar
         elif ndims == 1:
             d = None
-            if h5obj.dtype.kind == 'O':
+            if h5obj.shape == (1, ):
+                # read scalar (1-element) dataset into memory
+                d = h5obj[0]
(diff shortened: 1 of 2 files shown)


rly:
Actually, this will not work in cases where an array is expected. With this PR, a scalar would be given....

rly:
@oruebel, @ajtritt thoughts?

@t-b did I characterize the issue correctly? Would it be possible to push Igor Pro / MIES to support writing scalar (shape `()`) datasets?

ajtritt (changes requested):
This breaks things if a user means to write a length 1 dataset.

The ObjectMapper can check if the attribute/dataset is supposed to be a scalar, and reshape as needed.

rly:
@ajtritt Thanks, makes sense. I'll build it into the `ObjectMapper` where we can cross-check with the spec.

rly:
Tests fail for this PR because the `VectorData`, `VectorIndex`, and `DynamicTableRegion` neurodata_types have no `'shape'` (or `'dims'`) key and thus are expected to have scalar data. These should be specified in hdmf-common-schema and HDMF.

@ajtritt @oruebel @bendichter should `VectorData`, `VectorIndex`, and `DynamicTableRegion` always be 1-dimensional? Alternatively we could specify them as 1-D, 2-D, 3-D, or 4-D.

ajtritt:
@rly 

> should `VectorData`, `VectorIndex`, and `DynamicTableRegion` always be 1-dimensional? Alternatively we could specify them as 1-D, 2-D, 3-D, or 4-D.

`VectorIndex` and `DynamicTableRegion` should always be 1D. `VectorData` can be as many as you want, but should never be a scalar (hence the name _Vector_)

rly:
@ajtritt 
> VectorData can be as many as you want

Would it be reasonable to set a max number of dimensions, i.e., 1-D, 2-D, 3-D, or 4-D?

The spec language currently does not have a way to specify a dataset of any size. Lack of a 'shape' value means it is a scalar dataset -- at least that's what the [documentation](https://schema-language.readthedocs.io/en/latest/specification_language_description.html#sec-shape) says.

ajtritt:
@rly, Yeah, that's fine. If someone needs to store 5 or more dimensions, we can just add that on request.

t-b:
@rly Thanks for looking into that.

> @t-b did I characterize the issue correctly?

Yes this is correct. Here for reference the commit message from my PR:

```
 ObjectMapper.construct: Don't enforce special datastructures for scal…

…ar values

HDF5 Datasets with SIMPLE dataspace instead of SCALAR (hdmf default)
currently break the constructor call as the constructor expects e.g. a
string and not a h5py.DataSet.

We now extract these single values so that the constructor sees always
the same data.

This makes hdmf more compatible with Igor Pro as datasets are
always exported with dataspace SIMPLE in Igor Pro.

For reference the following file has been used:

HDF5 "E:\projekte\mies-igor\tools\unit-testing\HardwareTests.nwb" {
GROUP "/" {
   DATASET "session_start_time_scalar" {
      DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
      DATASPACE  SCALAR
   }
   DATASET "session_start_time_simple" {
      DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_UTF8;
            CTYPE H5T_C_S1;
         }
      DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
   }
}
}

and opening that with h5py shows the following

>>> f['/session_start_time_simple'].id.get_space().get_simple_extent_type()
1
>>> f['/session_start_time_scalar'].id.get_space().get_simple_extent_type()
0
>>> f['/session_start_time_scalar'][()]
'2019-11-19T17:47:01.070Z'
>>> f['/session_start_time_simple'][0]
'2019-11-19T17:47:01.070Z'
>>> f['/session_start_time_scalar']
<HDF5 dataset "session_start_time_scalar": shape (), type "|O">
>>> f['/session_start_time_simple']
<HDF5 dataset "session_start_time_simple": shape (1,), type "|O">
>>> f['/session_start_time_simple'][()]
array(['2019-11-19T17:47:01.070Z'], dtype=object)
>>> f['/session_start_time_scalar'][()]
'2019-11-19T17:47:01.070Z'
>>> f['/session_start_time_simple'][0]
'2019-11-19T17:47:01.070Z'
>>> print(h5py.__version__)
2.10.0

and therefore we can extract the first value from a SIMPLE dataspace to
get the same value as just indexing the SCALAR dataspace.
```

> Would it be possible to push Igor Pro / MIES to support writing scalar (shape ()) datasets?

Not easily as I would have to heavily extend the builtin HDF5 support.

rly:
PyNWB tests succeed locally under the `fix/pytest` branch, so I think this is good to go.

oruebel (approved):

ajtritt (approved):

PR #274, opened 2020-01-21
Fix broken support for pytest framework

### LINKED ISSUE #271: Pytest cannot run abstract classes in new testing module
The new `hdmf.testing` modules includes the abstract class `TestH5RoundTripMixin`. The `unittest` test runner works fine, but people who use `pytest` (including @bendichter) run into errors when running the new tests. The problem is that `pytest` automatically runs any test classes that start with 'Test*'. 

Since there is no need for the abstract test class to start with 'Test*', we can resolve this issue by renaming `TestH5RoundTripMixin` to `H5RoundTripMixin`.

rly:
Fix #271.

### DIFF
--- a/docs/source/make_roundtrip_test.rst
+++ b/docs/source/make_roundtrip_test.rst
@@ -25,13 +25,13 @@ Before writing tests, we also suggest you familiarize yourself with the
 :ref:`software architecture <software-architecture>` of HDMF.
 
 ------------------------
-``TestH5RoundTripMixin``
+``H5RoundTripMixin``
 ------------------------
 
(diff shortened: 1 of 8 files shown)


oruebel (approved):

oruebel (changes requested):
Looks like I approved to early. There is a broken test on CircelCI

rly:
@oruebel the broken the test is because the dev branch of pynwb imports some of these classes. These import statements were added in pynwb after the latest release -- they would be changed before the next release, so we would not be breaking backward compatibility.

oruebel (approved):

PR #276, opened 2020-01-21
make version required for SpecNamespace

### LINKED ISSUE #275: namespace version
the `version` arg of `NamespaceBuilder` should be required. It isn't 100% clear for the specification-language documentation whether version is required, but it should be. Some groups have been writing NWB files with uncached, unversioned, unregistered extensions and I am concerned this will cause problems. The least we can do is require that the extensions are versioned.

bendichter:
fix #275

### DIFF
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -18,7 +18,7 @@ _namespace_args = [
     {'name': 'name', 'type': str, 'doc': 'the name of this namespace'},
     {'name': 'schema', 'type': list, 'doc': 'location of schema specification files or other Namespaces'},
     {'name': 'full_name', 'type': str, 'doc': 'extended full name of this namespace', 'default': None},
-    {'name': 'version', 'type': (str, tuple, list), 'doc': 'Version number of the namespace'},
+    {'name': 'version', 'type': (str, tuple, list), 'doc': 'Version number of the namespace', 'default': None},
     {'name': 'date', 'type': (datetime, str),
      'doc': "Date last modified or released. Formatting is %Y-%m-%d %H:%M:%S, e.g, 2017-04-25 17:14:13",
      'default': None},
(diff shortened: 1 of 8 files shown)


rly:
Can you also make the change in `hdmf.spec.write.NamespaceBuilder`?

oruebel (dismissed):

src/hdmf/spec/namespace.py
```diff
@@ -18,7 +18,7 @@
     {'name': 'name', 'type': str, 'doc': 'the name of this namespace'},
     {'name': 'schema', 'type': list, 'doc': 'location of schema specification files or other Namespaces'},
     {'name': 'full_name', 'type': str, 'doc': 'extended full name of this namespace', 'default': None},
-    {'name': 'version', 'type': (str, tuple, list), 'doc': 'Version number of the namespace', 'default': None},
+    {'name': 'version', 'type': (str, tuple, list), 'doc': 'Version number of the namespace'},
```
rly: Because `'version'` is after an optional argument, this change will invisibly break code that uses only positional arguments. For example:
```python
ns = SpecNamespace('doc', 'name', [{'source': 'test.yaml'}], 'full name', 'version')
ns
```
old output:
```
{'doc': 'doc', 
'schema': [{'source': 'test.yaml'}], 
'name': 'name', 
'full_name': 'full name', 
'version': 'version'}
```
new output:
```
{'doc': 'doc', 
'schema': [{'source': 'test.yaml'}], 
'name': 'name', 
'full_name': 'version', 
'version': 'full name'}
```

So I am more in favor of keeping `'default': None` and checking for None in the constructor and raising an error there if it is none.

rly (changes requested):

src/hdmf/spec/namespace.py
rly: But since this is internal code that I believe is not used by anything outside of PyNWB/HDMF, I think this is fine. I would just reorder the docval args to make it clear that this is the new order of positional arguments.

rly (approved):

PR #292, opened 2020-02-09
Handle missing namespace version

### LINKED ISSUE #1183: error when reading file with unversioned extension
Trying to open another neuropixel file, I got that it couldn't open because the version was missing. I didn't think this through before when we decided to enforce this. I realized that writing without a version would cause an error, but did not fully appreciate that this would break read on existing Allen datasets. Can we adjust the object mapper to deal with this?

rly:
Fix https://github.com/NeurodataWithoutBorders/pynwb/issues/1183

- If a user loads a namespace (or otherwise creates a `SpecNamespace` object) without the version key, a warning is raised and the version is set to "unversioned".

- If a user creates a `NamespaceBuilder` (which is necessary to write a namespace to file) without the version key, an error is raised. Note that making `version` required cannot be done in the docval spec for `NamespaceBuilder` because there is an optional argument earlier in the arg list.

I also took the opportunity to refactor some of the unit tests.

### DIFF
--- a/src/hdmf/spec/write.py
+++ b/src/hdmf/spec/write.py
@@ -46,8 +46,13 @@ class YAMLSpecWriter(SpecWriter):
 
     def write_namespace(self, namespace, path):
         with open(os.path.join(self.__outdir, path), 'w') as stream:
-            ns = namespace
+            # version is required on write as of HDMF 1.5.4. check here instead of SpecNamespace.__init__ to maintain
+            # backwards compatibility with older namespaces that are missing version
+            if namespace['version'] is None:
+                raise TypeError("Namespace '%s' missing key 'version'. Please specify a version for the extension."
+                                % namespace['name'])
             # Convert the date to a string if necessary
+            ns = namespace
             if 'date' in namespace and isinstance(namespace['date'], datetime):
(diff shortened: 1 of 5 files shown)


bendichter (dismissed):
Yes this looks like exactly what we need

src/hdmf/spec/write.py
```diff
@@ -109,6 +114,11 @@ class NamespaceBuilder:
             {'name': 'namespace_cls', 'type': type, 'doc': 'the SpecNamespace type', 'default': SpecNamespace})
     def __init__(self, **kwargs):
         ns_cls = popargs('namespace_cls', kwargs)
+        if kwargs['version'] is None:
+            # version is required on write as of HDMF 1.5. this check should prevent the writing of namespace files
+            # without a verison
+            raise RuntimeError("Namespace '%s' missing key 'version'. Please specify a version for the extension."
```
ajtritt: I think the convention for improper arguments is to raise a ValueError

oruebel (dismissed):

src/hdmf/spec/write.py
rly: Fixed

ajtritt (dismissed):

ajtritt (approved):

PR #293, opened 2020-02-10
Allow docval to warn about use of positional arguments

rly:
This sets up infrastructure in docval for https://github.com/NeurodataWithoutBorders/pynwb/issues/1189

To raise a warning when a user provides positional arguments to a method, add `allow_positional=False` to the docval specification to the method.

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -3,6 +3,7 @@ from abc import ABCMeta
 import collections
 import h5py
 import numpy as np
+import warnings
 
 __macros = {
     'array_data': [np.ndarray, list, tuple, h5py.Dataset],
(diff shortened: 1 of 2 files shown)


src/hdmf/utils.py
```diff
@@ -116,7 +117,8 @@ def __format_type(argtype):
         raise ValueError("argtype must be a type, str, list, or tuple")
 
 
-def __parse_args(validator, args, kwargs, enforce_type=True, enforce_shape=True, allow_extra=False):   # noqa: C901
+def __parse_args(validator, args, kwargs, enforce_type=True, enforce_shape=True, allow_extra=False,  # noqa: C901
+                 allow_positional=True):
```
oruebel: If you want it configurable whether ``allow_positional`` should warn or raise and error, then you may want to make this an integer option (0=off, 1=warn, 2=error, 3="I'll call your parents")

rly: This is implemented now

ajtritt (approved):

Issue #273, opened 2020-01-21
performance of AbstractDataChunkIterator subclasses


d-sot:
## Description

Subclassing [AbstractDataChunkIterator](https://github.com/hdmf-dev/hdmf/blob/84af268a27d8343323c5eb9a6e86d994c0a2d9f6/src/hdmf/data_utils.py#L11), to iterate over chunks, e.g. in [SparseChunkIterator](https://github.com/ben-dichter-consulting/axel-lab-to-nwb/blob/master/axel_lab_to_nwb/utils/SparseChunkIterator.py), leads to considerably slower write performance than expected. It can be tested. For example, in ```SparseChunkIterator``` only chunks of interest are iterated over and written to disk, and the remaining chunks are not written.  I created a new identical data source, except that it is zeroed at chunks that are not of interest. Writing this new data to disk with ```H5DataIO``` without using any of ```ChunkIterator``` classes, gives a much faster response in some scenarios.

The timing difference is linearly related to the size of the chunk, and as a result to the number of calls [```__exhaust_dcis```](https://github.com/hdmf-dev/hdmf/blob/84af268a27d8343323c5eb9a6e86d994c0a2d9f6/src/hdmf/backends/hdf5/h5tools.py#L565) makes to [```__write_chunk__```](https://github.com/hdmf-dev/hdmf/blob/84af268a27d8343323c5eb9a6e86d994c0a2d9f6/src/hdmf/backends/hdf5/h5tools.py#L986).

In my test, for a chunk of approximately 64Kbytes ( chunk shape of (4,20,10,10) on dataset of shape (200,100,100,100) for float64 ), it'd be writing 15400 individual chunks. The average breakdown is:

2.874e-04 seconds per chunk for:
https://github.com/hdmf-dev/hdmf/blob/84af268a27d8343323c5eb9a6e86d994c0a2d9f6/src/hdmf/backends/hdf5/h5tools.py#L1012 
5.237e-05 seconds per chunk and for everything else in ```__write_chunk__``` except the call to ```__next__```:
https://github.com/hdmf-dev/hdmf/blob/84af268a27d8343323c5eb9a6e86d994c0a2d9f6/src/hdmf/backends/hdf5/h5tools.py#L1001-L1010

This is approximately 18% overhead. The overhead decreases as the size of chunk increases. That this is 4D dataset is also related, due to the N^4 number of chunks. For a higher dimensional data, this would be more pronounced.

[...]

PR #295, opened 2020-02-11
Write chunk update

d-sot:
This is regarding issue #273 . `__write_chunk__` currently checks if the new chunk is within the size of the current dataset, and uses `resize` to adjust the size if necessary. The performance of calling `dset.id.extend(max_bounds)` in each chunk iteration is much faster. `dset.id.extend` does the same thing, i.e. if `max_bounds` is larger than the current shape, `dset.id.extend` resizes the dataset, otherwise it leaves the dataset as is.

Also for performance reasons, this removes https://github.com/hdmf-dev/hdmf/blob/d432709374c7e8435f3a47fe29f906495d71ae97/src/hdmf/backends/hdf5/h5tools.py#L1001  replacing with:

```max_bounds = [x.stop or 0 for x in chunk_i.selection]```

assuming that it's reasonable that a `chunk_iterator` would only operate on chunks defined by slices.

## Motivation

hdmf performance when using chunk iterator compared to writing zeros instead.

## Checklist

### CODE REVIEW

**Comment by rly** [comment]
Thank you for digging deep into this issue @d-sot ! We are a bit in the middle of a few other things this week but will return to coding and take a close look at this early next week.

**Comment by rly** [comment]
Thanks, @d-sot. How much time is gained by replacing
```
max_bounds = cls.__selection_max_bounds__(chunk_i.selection) 
```
with
```
max_bounds = [x.stop or 0 for x in chunk_i.selection]
```

I am wary of removing existing functionality. While the only `AbstractDataChunkIterator` that we ship with HDMF uses `DataChunk` objects where the `selection` field is a tuple of numpy slices, user-defined `AbstractDataChunkIterator` subtypes may have int or array selections instead, and I would rather not remove that functionality if it does not create a significant performance hit. If we decide to restrict `DataChunk` objects to be iterables of numpy slices to work with the HDF5 backend, we should be explicit about that in the documentation of `AbstractDataChunkIterator` and `DataChunk`. 

Otherwise, this looks good.

**Comment by d-sot** [comment]
@rly The gain is not as significant, less than 2% in #273. But it is also that `h5py` does not fully support numpy's array and boolean indexing on more than one dimension (numpy fancy indexing with multtiple arrays is not their cartesian product), or in certain cases even in one dimension. for example:

```python
>>> dset.shape
(10,20,30,40)
>>> data[(2,4),0:5,2].shape
(2, 5, 40)
>>> data[2,0:5,(2,4)].shape
(2, 5, 40) # not as expected
>>> dset[2,0:5,(2,4)].shape
(5, 2, 40) # cartesian product
>>> dset[(2,4),0:5,2]=data[(2,4),0:5,2] # succeeds
>>> dset[2,0:5,(2,4)]=data[2,0:5,(2,4)]
*** TypeError: Broadcasting is not supported for complex selections
>>> dset[2,0:5,(2,4)]=np.transpose(data[2,0:5,(2,4)],[1,0,2]) # succeeds
>>> dset[np.array([True,False]*5),0:10,2] = data[[True,False]*5,0:10,2] # succeeds
>>> dset[2,0:10,np.array([True,False]*15)] = data[2,0:10,[True,False]*15]
*** TypeError: Broadcasting is not supported for complex selections
```
```python
>>> dset[2,(5,10),(2,4)].shape
*** TypeError: Only one indexing vector or array is currently allowed for advanced selection
>>> data[2,(5,10),(2,4)]=np.random.rand(2,2,40)
*** ValueError: shape mismatch: value array of shape (2,2,40) could not be broadcast to indexing result of shape (2,40)
```
Indexing with int always works and can be added to the `max_bound` list comprehension without noticeable overhead.

Would this work?

rly:
> Indexing with int always works and can be added to the max_bound list comprehension without noticeable overhead.

I think that would be good. Thanks for breaking down the use case.

d-sot:
You're welcome. int and single slice functionality is added back.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -997,21 +985,25 @@ def __write_chunk__(cls, dset, data):
...
+            # Determine the minimum array dimensions to fit the chunk selection
+            max_bounds = tuple([x.stop or 0 if isinstance(x, slice) else x+1 for x in chunk_i.selection])
+        except Exception as exc:
+            if isinstance(chunk_i.selection, int):
+                max_bounds = (chunk_i.selection+1, )
+            elif isinstance(chunk_i.selection, slice):
+                max_bounds = (chunk_i.selection.stop or 0, )
+            else:
+                msg = "Chunk selection %s must be a tuple of slices and/or integers" % str(chunk_i.selection)
+                raise TypeError(msg) from exc
```
oruebel:
```suggestion
        if isinstance(chunk_i.selection, tuple):
            # Determine the minimum array dimensions to fit the chunk selection
            max_bounds = tuple([x.stop or 0 if isinstance(x, slice) else x+1 for x in chunk_i.selection])
        elif isinstance(chunk_i.selection, int):
             max_bounds = (chunk_i.selection+1, )
        elif isinstance(chunk_i.selection, slice):
              max_bounds = (chunk_i.selection.stop or 0, )
        else:
             msg = "Chunk selection %s must be a single int, single slice, or tuple of slices and/or integers" % str(chunk_i.selection)
             raise TypeError(msg) from exc
```

I.e,. Rather than using try/except I think this should check for the type explicitly. The reason for this are: 
- If ``chunk_i.selection`` is a list the try/except block will not necessarily always trigger an exception, which I believe could lead to unexpected behavior 
- Checking for all supported cases is more explicit and safer as the reason for the Exception may not be clear

I also updated the TypeError message to reflect that single ints and slices are permitted as selections as well.

oruebel (changes requested):
Overall what the change does looks good to me. I added one suggestion for change, because the use of the Exception looks unsafe to me.

oruebel:
@d-sot thanks for your patience on this PR.

d-sot:
@oruebel Thanks for the update. This works great for excluding other iterators, and does give more clear error messages. The overhead of calling `isinstance` is also sufficiently small.

oruebel (dismissed):

oruebel:
Can you please fix the following flake8 errors. Otherwise this looks good to me:

src/hdmf/backends/hdf5/h5tools.py:994:14: E111 indentation is not a multiple of four
src/hdmf/backends/hdf5/h5tools.py:994:14: E117 over-indented
src/hdmf/backends/hdf5/h5tools.py:996:15: E111 indentation is not a multiple of four
src/hdmf/backends/hdf5/h5tools.py:996:15: E117 over-indented
src/hdmf/backends/hdf5/h5tools.py:998:14: E111 indentation is not a multiple of four
src/hdmf/backends/hdf5/h5tools.py:998:14: E117 over-indented
src/hdmf/backends/hdf5/h5tools.py:998:121: E501 line too long (135 > 120 characters)
src/hdmf/backends/hdf5/h5tools.py:999:14: E111 indentation is not a multiple of four
src/hdmf/backends/hdf5/h5tools.py:999:40: F821 undefined name 'exc'

rly (approved):

oruebel (approved):

PR #297, opened 2020-02-19
Refactor missing namespace version

rly:
Based on [commentary](https://github.com/hdmf-dev/hdmf/commit/d432709374c7e8435f3a47fe29f906495d71ae97#r37363245) from @yarikoptic and @ajtritt, refactor, and:
1. expose the magic value that represents a missing version: None
2. write "None" as the version when writing the spec to disk
3. set the `SpecNamespace.version` to None if "None" is read as the version from disk
4. set the `SpecNamespace.version` to None if "unversioned" is read as the version from disk
5. make sure the "unversioned" and "None" version namespaces are not loaded when a versioned namespace is present
6. add a bunch of tests to cover these edge-cases and backward-compatibility support

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -281,8 +281,6 @@ class HDF5IO(HDMFIO):
                 ns_builder = self.__convert_namespace(ns_catalog, ns_name)
                 namespace = ns_catalog.get_namespace(ns_name)
                 if namespace.version is None:
-                    group_name = '%s/unversioned' % ns_name
-                else:
                     group_name = '%s/%s' % (ns_name, namespace.version)
                 ns_group = spec_group.require_group(group_name)
                 writer = H5SpecWriter(ns_group)
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -10,8 +10,7 @@ from ...utils import docval, getargs, popargs, call_docval_func, get_data_shape
 from ...data_utils import AbstractDataChunkIterator
 from ...build import Builder, GroupBuilder, DatasetBuilder, LinkBuilder, BuildManager,\
                      RegionBuilder, ReferenceBuilder, TypeMap, ObjectMapper
-from ...spec import RefSpec, DtypeSpec, NamespaceCatalog, GroupSpec
-from ...spec import NamespaceBuilder
+from ...spec import RefSpec, DtypeSpec, NamespaceCatalog, GroupSpec, NamespaceBuilder
 
 from .h5_utils import BuilderH5ReferenceDataset, BuilderH5RegionDataset, BuilderH5TableDataset,\
                       H5DataIO, H5SpecReader, H5SpecWriter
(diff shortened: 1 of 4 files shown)


ajtritt (dismissed):

src/hdmf/spec/namespace.py
```diff
@@ -83,13 +85,16 @@ def contact(self):
...
 
     @property
     def version(self):
-        """String, list, or tuple with the version or None """
-        return self.get('version', None)
+        """String, list, or tuple with the version or SpecNamespace.UNVERSIONED """
+        version = self.get('version', None)
+        if version is None:
+            self['version'] = SpecNamespace.UNVERSIONED
+        return self.get('version')
```
yarikoptic: `return self.get('version', None) or SpecNamespace.UNVERSIONED` probably do the same if version can't be numeric ;-) could also benefit of it is an empty list or an empty string for some reason

rly: The code as written also sets self['version']. How about changing line 95 to:
```python
if not version:
```

yarikoptic: I love functions with side effects, they are so much fun to debug ;-)
If anything, could also just not use version variable - used only once

rly: True, side effects are not great. I'll go with your suggestion.

src/hdmf/spec/namespace.py
```diff
@@ -83,13 +85,16 @@ def contact(self):
...
 
     @property
     def version(self):
-        """String, list, or tuple with the version or None """
-        return self.get('version', None)
+        """
+        String, list, or tuple with the version or SpecNamespace.UNVERSIONED
+        if the version is missing or empty
+        """
+        return self.get('version', None) or SpecNamespace.UNVERSIONED
```
ajtritt: This would return a bool. Did you mean 

`return self.get('verzion', SpecNamespace.UNVERSIONED)`

ajtritt, line 101: This would return a bool. Did you mean this?

`return self.get('version', SpecNamespace.UNVERSIONED)`

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -280,10 +280,7 @@ def write(self, **kwargs):
             for ns_name in ns_catalog.namespaces:
                 ns_builder = self.__convert_namespace(ns_catalog, ns_name)
                 namespace = ns_catalog.get_namespace(ns_name)
-                if namespace.version is None:
-                    group_name = '%s/unversioned' % ns_name
-                else:
-                    group_name = '%s/%s' % (ns_name, namespace.version)
+                group_name = '%s/%s' % (ns_name, namespace.version)
```
ajtritt, line 294: I changed the default to be `None`. This would make the group name "None". I don't have a problem with that, since I don't think it would impact reading the namespace

ajtritt (commented):
@rly I made a couple of changes and comments. Please look them over... I suspect @yarikoptic would give a thumbs up.

src/hdmf/spec/namespace.py
rly: Interestingly, this does not return a bool. 
```python                            
>>> 'a' or 'b'
'a'
>>> False or 'b'
'b'
>>> None or 'b'
'b'
>>> None or None
>>>                      
```
As written, if `version` is somehow '', [], or (), then `SpecNamespace.UNVERSIONED` will be returned. 

If `self.get('version', SpecNamespace.UNVERSIONED)` is used, then if `version` is somehow '', [], or (), then those would get returned.

src/hdmf/backends/hdf5/h5tools.py
yarikoptic: cool.  When it is loaded from a file, would `"None"` string be casted to a proper `None` for the `.version`? Just checking, didn't spot anything obvious in the diff

rly: This is now resolved in `SpecNamespace.__init__`

rly: @yarikoptic Thanks for checking that!

ajtritt (approved):

PR #307, opened 2020-03-12
Raise more informative error when adding column w/ used name

rly:
When trying to add a column named "description", "name", or another string that is already an attribute on the `DynamicTable`, I get the error message: `AttributeError: can't set attribute 'description' -- already set`. This is not very informative. This PR raises a more informative error.

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -368,7 +368,7 @@ class DynamicTable(Container):
 
         # check to see if any of the extra columns just need to be added
         if extra_columns:
-            for col in self.__uninit_cols:
+            for col in self.__columns__:
                 if col['name'] in extra_columns:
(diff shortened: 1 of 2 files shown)


oruebel (dismissed):
Although this seems more of an API issue, i.e., from a schema perspective it seems it would be fine to have a column named ``description`` as well as the attribute ``description``.

rly:
Yeah, I agree this is more of an API issue. If someone writes a `DynamicTable` with the column name "description" using a different API, we would still want to be able to read it...

I'll create a separate ticket for that, but I think this PR will work as a stopgap to limit those cases.

rly:
For some reason, Azure builds are not triggering.

oruebel (approved):

PR #311, opened 2020-03-12
Change dtype conversion warning to include path to type

### LINKED ISSUE #306: Include object name in UserWarning "Value with data type x is being converted to y"
Follow up on #298 

Now I do get a warning like this one on write... but have no clue where it's coming from. 
I think the warning message should also contain the object to which it refers.
```python
./miniconda3/envs/nwb_latest/lib/python3.7/site-packages/hdmf/build/objectmapper.py:146:
UserWarning: Value with data type int64 is being converted to data type uint64
(min specification: uint32).
  % (g.name, ret_type.name, s.name))
```

rly:
Fix #306.

Note that this gives the path to the spec value, NOT the container hierarchy. It might not be as intuitive to users as giving the container hierarchy; however, that would require a large amount of refactoring, given how certain functions are currently encapsulated.

### DIFF
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -145,7 +145,8 @@ class Spec(ConstructableDict):
     def __hash__(self):
         return id(self)
 
-    def path_str(self):
+    @property
+    def path(self):
         stack = list()
         tmp = self
         while tmp is not None:
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -146,15 +146,17 @@ class Spec(ConstructableDict):
         return id(self)
 
     def path_str(self):
-        """
-        Return string showing full path to this spec, with spec names separated by '/', regardless of the spec type.
-        """
-        spec_path = self.name
-        spec = self.parent
-        while spec is not None:
-            spec_path = spec.name + '/' + spec_path
-            spec = spec.parent
-        return '/' + spec_path
+        stack = list()
+        tmp = self
+        while tmp is not None:
+            name = tmp.name
+            if name is None:
+                name = tmp.data_type_def
+                if name is None:
+                    name = tmp.data_type_inc
+            stack.append(name)
+            tmp = tmp.parent
+        return "/".join(reversed(stack))
 
 #    def __eq__(self, other):
 #        return id(self) == id(other)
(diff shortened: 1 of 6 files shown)


oruebel (dismissed):

oruebel:
I think having the spec path should be helpful enough. At least for developers it should make the debugging easier.

src/hdmf/spec/spec.py
```diff
@@ -145,6 +145,17 @@ def build_const_args(cls, spec_dict):
...
         return id(self)
 
+    def path_str(self):
+        """
+        Return string showing full path to this spec, with spec names separated by '/', regardless of the spec type.
+        """
+        spec_path = self.name
+        spec = self.parent
+        while spec is not None:
+            spec_path = spec.name + '/' + spec_path
```
ajtritt: I couple things...

- Not all specs have `name` set, so in the case of `data_type` specs (e.g. DynamicTable), this will be ambiguous. You might consider having the BaseStorageSpec constructor pass in `data_type_def` as `name`.
- Since parent can't be reassigned, shouldn't this value be constant i.e. it shouldn't need to be computed every time the function gets called?
- I think this should be a property. Is there a reason you didn't go that route?
- I prefer `path` to `path_str`. What do you think?

tests/unit/spec_tests/test_group_spec.py
```diff
@@ -209,3 +209,29 @@ def test_update_attribute_spec(self):
         res = spec.get_attribute('attribute1')
         self.assertEqual(res.value, 5)
         self.assertEqual(res.dtype, 'int')
+
```
ajtritt, line 212: Can you also test for the case when `Spec.name` is not set?

ajtritt (changes requested):

src/hdmf/spec/spec.py
rly: > * Not all specs have `name` set, so in the case of `data_type` specs (e.g. DynamicTable), this will be ambiguous. You might consider having the BaseStorageSpec constructor pass in `data_type_def` as `name`.

Gotcha. It turns out there is already in `Validator.get_spec_loc` that does what `Spec.path_str` was trying to do, but takes into account cases with a `data_type_def` or `data_type_inc` and no name. So I moved that code into `Spec` and refactored `Validator.get_spec_loc(cls, spec)` to call it.  

> * Since parent can't be reassigned, shouldn't this value be constant i.e. it shouldn't need to be computed every time the function gets called?

Parent can be assigned after the spec is created. It's an odd case, but you can create a `GroupSpec` `A` with a path `'A'`. Spec `A` could be added to `GroupSpec` `B` and would then have path `'B/A'`. Spec `B` could be added to `GroupSpec` `C` and would then have path `'C/B/A'`, and so on.

> * I think this should be a property. Is there a reason you didn't go that route?

No reason. It does make sense as a property (getter only). I changed it. However, if it is not a constant, should it still be a property?

> * I prefer `path` to `path_str`. What do you think?

Agreed. I like `path` better. I changed it.

tests/unit/spec_tests/test_group_spec.py
rly: Added.

src/hdmf/spec/spec.py
```diff
@@ -361,6 +375,24 @@ def resolve_spec(self, **kwargs):
         for attribute in inc_spec.attributes:
             self.__new_attributes.discard(attribute)
             if attribute.name in self.__attributes:
+                # copy over the fields 'shape' and 'dims' of the parent attribute to this spec's attribute
```
ajtritt: This modification happened in #321

src/hdmf/spec/spec.py
```diff
@@ -680,6 +712,30 @@ def __is_sub_dtype(cls, orig, new):
     @docval({'name': 'inc_spec', 'type': 'DatasetSpec', 'doc': 'the data type this specification represents'})
     def resolve_spec(self, **kwargs):
         inc_spec = getargs('inc_spec', kwargs)
+        # copy over fields of the parent dataset to this spec
```
ajtritt: This modification happened in #321

ajtritt (approved):

PR #318, opened 2020-03-20
Use the numpy dtype in get_type if the array is empty

oruebel:
## Motivation

Fix issue identified in https://github.com/oruebel/ndx-icephys-meta/issues/57 with validating empty datasets. This is triggered in the icephys extension because it may use an empty DynamicTable (with just and id column) resulting in an empty columns dataset. 

## How to test the behavior?

See https://github.com/oruebel/ndx-icephys-meta/issues/57#issuecomment-601448001

## Checklist

- [X] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured the PR description clearly describes problem and the solution?
- [X] Is your contribution compliant with our coding style ? This can be checked running `flake8` from the source directory.
- [X] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/hdmf-dev/hdmf/pulls) for the same change?
- [X] Have you included the relevant issue number using `#XXX` notation where `XXX` is the issue number ? By including "Fix #XXX" you allow GitHub to close the corresponding issue.

PR #319, opened 2020-03-24
Fix bug on gettype on empty array, add validator tests

### LINKED ISSUE #317: Validating empty array against spec should show better error
If an attribute spec specifies a dtype text with or without a shape, and an empty numpy array is validated against that spec, the error returned is: `IndexError: index 0 is out of bounds for axis 0 with size 0`. This is not very helpful.

Is an empty array allowed for `shape: None` (1D array with any length) ?

To clarify, what should be the correct validation behavior for the following cases:
1) Spec specifies `shape: None` and `dtype: float` and test value is empty array with type float -- ShapeError?
2) Spec specifies `shape: None` and `dtype: text` and test value is empty array with type float -- ShapeError and DtypeError?
3) Spec specifies `shape: None` and `dtype: text` and test value is empty list -- ShapeError?
4) Spec specifies scalar and `dtype: text` and test value is empty array/list -- ShapeError?
5) Spec specifies scalar and `dtype: text` and test value is empty string -- OK?

rly:
## Motivation

Fix #317. Supersedes fix in #318. Closes #318. Adds a bunch of missing tests for validating the data type and shape of a dataset or attribute.

Fix issue identified in https://github.com/oruebel/ndx-icephys-meta/issues/57 with validating empty datasets. This is triggered in the icephys extension because it may use an empty `DynamicTable` (with just and id column) resulting in an empty columns dataset. 

## How to test the behavior?

See https://github.com/oruebel/ndx-icephys-meta/issues/57#issuecomment-601448001

### DIFF
--- a/src/hdmf/validate/errors.py
+++ b/src/hdmf/validate/errors.py
@@ -7,11 +7,13 @@ __all__ = [
     "Error",
     "DtypeError",
     "MissingError",
+    "ExpectedScalarError",
     "ExpectedArrayError",
     "ShapeError",
     "MissingDataType",
     "IllegalLinkError",
-    "IncorrectDataType"
+    "IncorrectDataType",
+    "EmptyDataNoTypeWarning"
 ]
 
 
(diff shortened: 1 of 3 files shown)


rly:
It seems that something is going wrong with differentiating the YAML:
```yaml
shape:
- null
```
from:
```yaml
shape: null
```
No need to review yet. I will patch this up.

rly:
The error in PyNWB is due to `dtype` and `shape` not being passed from parent type to subtype, so this PR requires #321, which solves that issue, to be merged first.

rly:
Closing due to staleness. See https://github.com/hdmf-dev/hdmf/issues/317#issuecomment-4363418969. Will be replaced by a new PR.

PR #322, opened 2020-03-25
Register new child types before new parent type for dynamic cls gen

### LINKED ISSUE #1210: reading the internal dataset spec in a custom extension doesn't work as I expect it?
Apologies if I am missing something here, but I am having a hard time getting NWBHDF5IO to read the specifications in a custom extension to the schema consistently. I have written minimal example code to show what I mean:

I created a custom extension to the NWB schema using this code (generate_dummy_schema.py)
```python
from pynwb.spec import NWBAttributeSpec, NWBDatasetSpec, NWBGroupSpec, NWBNamespaceBuilder


def generate_dummy_schema():
    # set up namespace
    ns_builder = NWBNamespaceBuilder('Dummy extensions for testing NWB',
                                     'dummy_extended_schema',
                                     'Dummy data extension to NWB format',
                                     version='0.1')
    ns_builder.include_type('LabMetaData', namespace='core')
    # define attributes
    dummy_attr = NWBAttributeSpec(name='dummy_attribute',
                                  doc='dummy attribute doc',
                                  dtype='float')
    # define datasets
    dummy_ds = NWBDatasetSpec(doc='dummy data set, recording some data',
                              shape=(None, 4),
                              attributes=[],
                              neurodata_type_def='DummyDataset')
    # define groups
    dummy_ophys_specs = NWBGroupSpec(doc='A place to store dummy extended optophysiology data',
                                     attributes=[dummy_attr],
                                     datasets=[dummy_ds],
                                     neurodata_type_def='DummyOptophysiology',
                                     neurodata_type_inc='LabMetaData')
    # export as schema extension
    ext_source = 'dummy.ophys.yaml'
    ns_builder.add_spec(ext_source, dummy_ophys_specs)
    ns_builder.export('dummy.namespace.yaml')


if __name__ == '__main__':
    generate_dummy_schema()

```
Then I write a file using this extension (write_dummy_file.py)
```python
from datetime import datetime
import pkg_resources
from numpy import zeros
from pynwb import get_class, load_namespaces, NWBFile, NWBHDF5IO


def write():
    file = "dummy_extended_nwb_file.nwb"
    # assume dummy extension is in this file's directory
    load_namespaces(pkg_resources.resource_filename(__name__, "dummy.namespace.yaml"))

    # set up nwbfile
    nwb_settings = {
        'session_start_time': datetime.now(),
        'identifier': "{}; {}".format(1, datetime.now()),
        'session_description': 'nwb dummy description',
        'session_id': '1',
    }
    nwb_file = NWBFile(**nwb_settings)

    # dynamically get classes from extension
    ds_class = get_class('DummyDataset', 'dummy_extended_schema')
    zero_array = zeros((10, 4))
    dataset_instance = ds_class(name='dummy_dataset', data=zero_array)
    optophysiology_class = get_class('DummyOptophysiology', 'dummy_extended_schema')
    dummy_optophysiology_instance = optophysiology_class(name='dummy_optophysiology',
                                                         dummy_attribute=2.4,
                                                         dummy_dataset=dataset_instance)

    # write some instances of extension to file
    nwb_file.add_lab_meta_data(dummy_optophysiology_instance)

    io = NWBHDF5IO(file, 'w')
    io.write(nwb_file)
    io.close()


if __name__ == '__main__':
    write()

```
The file looks as expected in a viewer - all the information for the specifications as well as instances of the specified types containing values are there.

I then run a script trying to read the file in three different ways (read_dummy_file.py) 
```python
from pynwb import NWBHDF5IO, load_namespaces, get_class

try:
    with NWBHDF5IO('dummy_extended_nwb_file.nwb', 'r', load_namespaces=True) as io:
        file = io.read()
        print("loading namespaces worked!")
except ValueError:
    print("loading namespaces did not work")

try:
    with NWBHDF5IO('dummy_extended_nwb_file.nwb', 'r', extensions='dummy.namespace.yaml') as io:
        file = io.read()
        print("reading extensions worked!")
except ValueError:
    print("reading extensions did not work")

try:
    load_namespaces('dummy.namespace.yaml')
    get_class('DummyDataset', 'dummy_extended_schema')
    with NWBHDF5IO('dummy_extended_nwb_file.nwb', 'r') as io:
        file = io.read()
    print('reading namespace and class before reading file worked!')
except ValueError:
    print('reading namespace and class before reading file did not work')

```

When I run the read script in the same python console as the write script, all three methods of reading work, but in a new python console, only the last one works. The first two fail with a ValueError stating that one of the dataset types from the custom specs wasn't found:
```
ValueError: Cannot dynamically generate class for type 'Unknown'. Type 'DummyDataset' does not exist. Please define that type before defining 'Unknown'.
```

The last method requires an explicit reference to the yaml file describing the schema extension and two extra lines of code, so the reading of the extensions or the loading of namespaces doesn't work as I expect it to (but I may be wrong)??

I am using pynwb 1.2.1, python 3.6. The files can be found [here](https://github.com/SilverLabUCL/PySilverLabNWB/tree/dummy-io-example-test).

[edited to show python formatting]

rly:
Fixes https://github.com/NeurodataWithoutBorders/pynwb/issues/1210 .

Issue:
An extension defines a new data type that includes a child group or dataset that also defines a new data type. There are no custom classes for either data type. When reading a file with these two extension data types, if the parent data type is encountered before the child data type, then dynamic class generation for the parent data type fails because the child data type does not exist.

In this PR, I changed dynamic class generation so that generating a class for a group data type will first generate classes for all children of the group that lack classes.

### DIFF
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -549,7 +549,6 @@ class TypeMap:
                 name = spec.data_type_def
                 if name is None:
                     name = 'Unknown'
-                breakpoint()
                 raise ValueError("Cannot dynamically generate class for type '%s'. " % name
                                  + str(e)
                                  + " Please define that type before defining '%s'." % name)
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -516,6 +516,9 @@ class TypeMap:
         cls = self.__get_container_cls(namespace, data_type)
         if cls is None:
             spec = self.__ns_catalog.get_spec(namespace, data_type)
+            if isinstance(spec, GroupSpec):
+                self.__resolve_child_container_classes(spec, namespace)
+
             dt_hier = self.__ns_catalog.get_hierarchy(namespace, data_type)
             parent_cls = None
             for t in dt_hier:
@@ -534,7 +537,7 @@ class TypeMap:
                 parent_cls = bases[0]
             if type(parent_cls) is not ExtenderMeta:
                 raise ValueError("parent class %s is not of type ExtenderMeta - %s" % (parent_cls, type(parent_cls)))
-            name = data_type
+
             attr_names = self.__default_mapper_cls.get_attr_names(spec)
             fields = dict()
             for k, field_spec in attr_names.items():
@@ -543,13 +546,23 @@ class TypeMap:
             try:
                 d = self.__get_cls_dict(parent_cls, fields, spec.name, spec.default_name)
             except TypeDoesNotExistError as e:
-                name = spec.get('data_type_def', 'Unknown')
+                name = spec.data_type_def
+                if name is None:
+                    name = 'Unknown'
+                breakpoint()
                 raise ValueError("Cannot dynamically generate class for type '%s'. " % name
                                  + str(e)
                                  + " Please define that type before defining '%s'." % name)
-            cls = ExtenderMeta(str(name), bases, d)
+            cls = ExtenderMeta(str(data_type), bases, d)
             self.register_container_type(namespace, data_type, cls)
         return cls
+
+    def __resolve_child_container_classes(self, spec, namespace):
+        for child_spec in (spec.groups + spec.datasets):
+            if child_spec.data_type_inc is not None:
+                self.get_container_cls(namespace, child_spec.data_type_inc)
+            elif child_spec.data_type_def is not None:
+                self.get_container_cls(namespace, child_spec.data_type_def)
 
     def __get_container_cls(self, namespace, data_type):
         if namespace not in self.__container_types:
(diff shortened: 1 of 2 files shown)


rly:
The patch coverage is low because the error `TypeDoesNotExistError` should not happen anymore but I modified and left in the code to handle the error in the event that it somehow does. There is no test that touches that error handling code because I cannot think of a way that it would be triggered. I would also be fine with removing the error handling code.

oruebel (dismissed):

oruebel:
> The patch coverage is low because the error `TypeDoesNotExistError` should not happen anymore

For parts of code like this that are errors we don't expect in practice you could add ``# pragma: no cover`` on the lines you want to exclude from code coverage and add a comment why these lines are ignored in code coverage.

ajtritt (dismissed):

rly:
Thanks for the reviews @oruebel and @ajtritt and for the tip about `# pragma: no cover`. I added that comment to the appropriate places.

bendichter:
@rly thanks for handling this! This is a big step forward for usability!

oruebel (approved):

PR #323, opened 2020-03-25
Extract _init_class_columns for use by DynamicTable subclasses

rly:
This PR extracts code in `DynamicTable.__init__` to `DynamicTable._init_class_columns` so that it can be called by subclasses of `DynamicTable`. This is necessary for the fix in https://github.com/NeurodataWithoutBorders/pynwb/pull/1205

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -162,8 +162,8 @@ class DynamicTable(Container):
             msg = "'__columns__' must be of type tuple, found %s" % type(cls.__columns__)
             raise TypeError(msg)
 
-        if len(bases) and 'DynamicTable' in globals() and issubclass(bases[-1], Container) \
-                and bases[-1].__columns__ is not cls.__columns__:
+        if (len(bases) and 'DynamicTable' in globals() and issubclass(bases[-1], Container)
+                and bases[-1].__columns__ is not cls.__columns__):
             new_columns = list(cls.__columns__)
             new_columns[0:0] = bases[-1].__columns__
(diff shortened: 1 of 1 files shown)


oruebel (approved):

src/hdmf/common/table.py
```diff
@@ -294,6 +294,9 @@ def __init__(self, **kwargs):  # noqa: C901
 
         self.__df_cols = [self.id] + [col_dict[name] for name in self.colnames]
         self.__colids = {name: i+1 for i, name in enumerate(self.colnames)}
+        self._init_class_columns()
+
+    def _init_class_columns(self):
```
oruebel, line 299: Note, this will not fix the issue. The single ``_`` makes the function private (i.e., Python will mangle the name), which makes it awkward to call the function from child classes. I'd suggest to use a double ``__`` instead.

rly: I thought double `__` makes the function private with Python name mangling and `_` is how you signal that it should only be used internally and by subclasses (protected).

https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name/1301369

rly: Calling from a subclass works fine in my tests: https://github.com/NeurodataWithoutBorders/pynwb/pull/1205/files#diff-edc3e1a40efc7c55019dfdfbc9d9cc9cR136-R142

oruebel (approved):

PR #324, opened 2020-03-31
Raise warning not error when adding column with existing attr name

### LINKED ISSUE #312: Cannot read DynamicTable with certain column names
Reading a `DynamicTable` with the column name "description" or "name" (e.g., that was written using a non PyNWB-HDMF API) raises an error. Certain other values are not allowed either because they are attributes of the `DynamicTable` class. From a schema standpoint, these names should be allowed, but it would take possibly a substantial reworking of the `DynamicTable` code to allow them.

See https://github.com/NeurodataWithoutBorders/pynwb/issues/1180#issuecomment-583656994
and https://github.com/hdmf-dev/hdmf/pull/307 .

rly:
Fix #312 

There are two primary ways to access columns of DynamicTables by name: 1) `table['colname']` and 2) `table.colname`.

However, if the `'colname'` conflicts with an existing attribute (or method) of the table, then currently an error will be raised. This PR instead changes the behavior to be similar to [pandas](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#attribute-access), where if a column is added to a table and the column has a name that is already an attribute (or method) of the table, the table attribute will not be overridden and the column can only be accessed via other methods.

Pandas does not raise a warning in this case. Here, I added a warning, since the existing attributes for a `DynamicTable` (or one of its subclasses) are not obvious to the user.

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -2,6 +2,7 @@ from h5py import Dataset
 import numpy as np
 import pandas as pd
 from collections import OrderedDict
+from warnings import warn
 
 from ..utils import docval, getargs, ExtenderMeta, call_docval_func, popargs, pystr
 from ..container import Container, Data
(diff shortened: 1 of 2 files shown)


rly:
This PR will help unblock one of the issues with reading Allen Institute files where one of the table column names is "name" and an error was raised. Now, to access that column, users should use `table['name']`. `table.name` will return the name of the table, not the column named 'name'.

oruebel (approved):

PR #325, opened 2020-04-03
Allow specfication of custom VectorData types for add_column

oruebel:
## Motivation

Allow users to create derived ``VectorData`` classes and use them with DynamicTable. This came up as a need to implement the column with TimeSeries references as its own type in https://github.com/oruebel/ndx-icephys-meta/issues/60

## Checklist


rly:
This feature has already been added to dev, but instead of `colcls`, the argument is named `col_cls`.

PR #326, opened 2020-04-06
WIP: Add export functionality

### LINKED ISSUE #315: unable to read -> (modify) -> write due to container_source
problem
-----------

I would like to read an nwbfile in, potentially modify the nwbfile object, then write it out to some other sink (maybe a new h5 file). This fails because container source cannot be reset. 

repro
-------
```Python
from datetime import datetime

import pynwb

def main():

    with pynwb.NWBHDF5IO(path="repro.nwb", mode="w") as nwb_writer:
        nwbfile = pynwb.NWBFile(
            session_description="",
            identifier="",
            session_start_time=datetime.now()
        )
        nwb_writer.write(nwbfile)

    with pynwb.NWBHDF5IO(path="repro.nwb", mode="r") as nwb_reader:
        nwbfile_obt = nwb_reader.read()
        
        with pynwb.NWBHDF5IO(path="repro_out.nwb", mode="w") as nwb_rewriter:
            nwb_rewriter.write(nwbfile_obt)

if __name__ == "__main__":
    main()
```
produces:
```
  File "container_source_repro.py", line 25, in <module>
    main()
  File "container_source_repro.py", line 20, in main
    nwb_rewriter.write(nwbfile_obt)
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/utils.py", line 483, in func_call
    return func(args[0], **parsed['args'])
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/backends/hdf5/h5tools.py", line 280, in write
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/utils.py", line 364, in call_docval_func
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/utils.py", line 483, in func_call
    return func(args[0], **parsed['args'])
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/backends/io.py", line 43, in write
    f_builder = self.__manager.build(container, source=self.__source)
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/utils.py", line 483, in func_call
    return func(args[0], **parsed['args'])
  File "/home/nile/Documents/work/allen_institute/hdmf/src/hdmf/build/manager.py", line 157, in build
    container.__class__.__name__))
ValueError: Cannot change container_source once set: 'root' pynwb.file.NWBFile
```

with pynwb 1.3.0, hdmf 1.6.1 in python 3.7.6 on ubuntu 19.10.

more
-------
It seems like this workflow would be handy for appending data/metadata or converting to new backends. Obviously, several workarounds exist: we can copy the file and open it `"r+"`, we can traverse the nwbfile and set all of the container sources directly via private attribute. These are not great options, since they are either fragile or inflexible.

rly:
Fix #315 and https://github.com/NeurodataWithoutBorders/pynwb/issues/668

This is a work in progress.

This also addresses two other potential issues discovered during debugging:
1) iterating over a `DatasetOfReferences` returned the reference, not the referenced object, so I added `__iter__` and `__next__` to `DatasetofReferences`
2) the `link_data` keyword argument for `HDF5IO.write` was not being used when writing datasets within groups.

TODO:
- [ ] Test edge case where after read of an existing file, a new link is made to an existing container but not using LinkSpec (see objectmapper.py 947:951). It doesn't seem to work.
- [ ] Note that if a container A is removed from container B that contains it, the children list of container B needs to be updated. This can be hard to track.
- [ ] add test roundtrip with export to hdmf and pynwb testing modules and test suites
- [ ] handle external lninks

I haven't fully tested these potential issues and solutions yet, so they remain part of this WIP. The commits can be cherry-picked into their own issues/PRs if appropriate.

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -67,15 +67,15 @@ class DatasetOfReferences(H5Dataset, ReferenceResolver, metaclass=ABCMeta):
             self.__inverted = cls(**kwargs)
         return self.__inverted
 
-    def __get_ref(self, ref):
+    def _get_ref(self, ref):
         return self.get_object(self.dataset.file[ref])
 
     def __iter__(self):
         for ref in super().__iter__():
-            yield self.__get_ref(ref)
+            yield self._get_ref(ref)
 
     def __next__(self):
-        return self.__get_ref(super().__next__())
+        return self._get_ref(super().__next__())
 
 
 class BuilderResolverMixin(BuilderResolver):
@@ -118,7 +118,7 @@ class AbstractH5TableDataset(DatasetOfReferences):
             if t is RegionReference:
                 self.__refgetters[i] = self.__get_regref
             elif t is Reference:
-                self.__refgetters[i] = self.__get_ref
+                self.__refgetters[i] = self._get_ref
         self.__types = types
         tmp = list()
         for i in range(len(self.dataset.dtype)):
@@ -163,7 +163,7 @@ class AbstractH5TableDataset(DatasetOfReferences):
             row[i] = getref(row[i])
 
     def __get_regref(self, ref):
-        obj = self.__get_ref(ref)
+        obj = self._get_ref(ref)
         return obj[ref]
 
     def resolve(self, manager):
@@ -175,9 +175,9 @@ class AbstractH5ReferenceDataset(DatasetOfReferences):
     def __getitem__(self, arg):
         ref = super().__getitem__(arg)
         if isinstance(ref, np.ndarray):
-            return [self.get_object(self.dataset.file[x]) for x in ref]
+            return [self._get_ref(x) for x in ref]
         else:
-            return self.get_object(self.dataset.file[ref])
+            return self._get_ref(ref)
 
     @property
     def dtype(self):
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -67,6 +67,16 @@ class DatasetOfReferences(H5Dataset, ReferenceResolver, metaclass=ABCMeta):
             self.__inverted = cls(**kwargs)
         return self.__inverted
 
+    def __get_ref(self, ref):
+        return self.get_object(self.dataset.file[ref])
+
+    def __iter__(self):
+        for ref in super().__iter__():
+            yield self.__get_ref(ref)
+
+    def __next__(self):
+        return self.__get_ref(super().__next__())
+
 
 class BuilderResolverMixin(BuilderResolver):
     """
@@ -151,9 +161,6 @@ class AbstractH5TableDataset(DatasetOfReferences):
         for i in self.__refgetters:
             getref = self.__refgetters[i]
             row[i] = getref(row[i])
-
-    def __get_ref(self, ref):
-        return self.get_object(self.dataset.file[ref])
 
     def __get_regref(self, ref):
         obj = self.__get_ref(ref)
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -72,6 +72,7 @@ class HDF5IO(HDMFIO):
         self.__ref_queue = deque()  # a queue of the references that need to be added
         self.__dci_queue = deque()  # a queue of DataChunkIterators that need to be exhausted
         ObjectMapper.no_convert(Dataset)
+        self._written_builders = dict()  # keep track of which builders were written (or read) by this IO object
 
     @property
     def comm(self):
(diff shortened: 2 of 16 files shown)


src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -67,6 +67,16 @@ def invert(self):
             self.__inverted = cls(**kwargs)
         return self.__inverted
 
+    def __get_ref(self, ref):
```
oruebel: Doesn't ``__get_ref`` get mangled? This I think will cause trouble with the derived ``AbstractH5TableDataset`` class which uses this function in its ``__init__`` method. I think the method should be renamed to only have a single underscore so that derived classes still have access to it.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -783,6 +791,7 @@ def write_dataset(self, **kwargs):  # noqa: C901
         options['dtype'] = builder.dtype
         dset = None
         link = None
+        print('    writing hdf5 dataset', builder.name, link_data, type(data))
```
oruebel: All these debug prints should be removed. We should consider using Python's logging module for the I/O-related parts of HDMF. I think that could help with debugging.

src/hdmf/build/objectmapper.py
```diff
@@ -716,6 +716,7 @@ def __add_attributes(self, builder, attributes, container, build_manager, source
                         msg = "invalid type for reference '%s' (%s) - "\
                               "must be AbstractContainer" % (spec.name, type(attr_value))
                     raise ValueError(msg)
+                breakpoint()
```
oruebel: breakpoint should be removed

src/hdmf/backends/io.py
```diff
@@ -43,6 +43,17 @@ def write(self, **kwargs):
         f_builder = self.__manager.build(container, source=self.__source)
         self.write_builder(f_builder, **kwargs)
 
+    @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
+            {'name': 'exhaust_dci', 'type': bool,
+             'doc': 'exhaust DataChunkIterators one at a time. If False, exhaust them concurrently', 'default': True},
+            allow_extra=True)
+    def export(self, **kwargs):
```
oruebel: should add a docstring describing the purpose of the export function (and difference to calling write)

src/hdmf/build/manager.py
```diff
@@ -97,6 +98,10 @@ def namespace_catalog(self):
     def type_map(self):
         return self.__type_map
 
+    @property
+    def export(self):
```
oruebel, line 104: We should document why ``export`` is a property of the ``BuildManager`` compared to it being just an argument on the ``build`` function. I assume this is the case because it changes state of the manager overall, i.e., a BuildManager with export=true has container sources set in its __containers and with export=false it does not. And export should also not be able to change after the fact for that reason. Is this correct?

src/hdmf/build/objectmapper.py
rly: Sorry, this code doesn't actually work and needs more debugging, hence the WIP and draft PR status. I just pushed it so that you and @ajtritt could take a look at it with me when we meet.

src/hdmf/backends/hdf5/h5_utils.py
rly: Good point. I'll fix that.

rly:
Currently, the following syntax works:
```python
with NWBHDF5IO(filename, 'r') as io:
    read_nwbfile = io.read()
    read_nwbfile.acquisition.pop('raw_ephys_data')  # removing a group is OK

    NWBHDF5IO.export(read_nwbfile, type_map=io.type_map, path=copy_filename)
```

And the following does not:
```python
with NWBHDF5IO(filename, 'r') as io:
    read_nwbfile = io.read()
    copy_nwbfile = read_nwbfile.copy()

    NWBHDF5IO.export(copy_nwbfile, type_map=io.type_map, path=copy_filename)
```
due to failure to resolve a reference created as a result of `read_nwbfile.copy()`. Namely, the `Units` table of `copy_nwbfile` has a `<DynamicTableRegion> 'electrodes'` which has an attribute called `'table'` that is a broken reference to the electrodes table in the original file, rather than a reference to the electrodes table in the exported file.

oruebel:
> And the following does not:

The simplest solution may be, since the main purpose of the export function is to go from one backend store to another, one solution may be to use the HDMFIO object (i.e., ``io`` in your code) as input to the export function, rather than the container (i.e.,  ``read_nwbfile``) in your code. In this way, you could move the ``io.read`` inside the export function and the second path you describe would not be possible. In fact, this would prevent any changes from occurring between the files on export, i.e., that approach may no cover the use-case where you need to modify the file before export (e.g., drop certain containers).  However, I'm wondering whether having at least a variant of export that works like this would be useful, since it should cover a lot of the use-case (at least it would solve the problem of converting between different backends, where manipulations (e.g., copy) would break things anyways).

rly:
I agree - it makes sense to have a variant of the export function for the use case where the data does not change between backends.

I think the other use case where the `NWBFile` may change between reads is still very useful, but it will need a lot more testing to handle all the various manipulations that can be done. Also, there are issues with `NWBFile.copy` that need to be sorted out first: https://github.com/NeurodataWithoutBorders/pynwb/issues/1224 . 

I will go ahead and change this function to work as below and we can add the second use case later on.
```python
read_io = NWBHDF5IO(filename, 'r')
NWBHDF5IO.export(read_io=read_io, path=copy_filename)
```

src/hdmf/build/manager.py
rly: A `BuildManager` with export=True has no cached builders and ignores existing container source values. A `BuildManager` with export=False checks that the container is being written to is the same as the container source when it exists.

The `export` field does not change after initialization.

I added the requested documentation

PR #329, opened 2020-04-12
improve unsupported filter error message

### LINKED ISSUE #212: type checking for dynamically loaded filters, and working with hdf5plugin

bendichter:
fix https://github.com/hdmf-dev/hdmf/issues/212

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -427,9 +427,10 @@ class H5DataIO(DataIO):
         # Confirm that the compressor is supported by h5py
         if not self.filter_available(self.__iosettings.get('compression', None),
                                      self.__allow_plugin_filters):
-            raise ValueError("%s compression may not be supported by this version of h5py. Use "
-                             "`allow_plugin_filters=True` to bypass this check" %
-                             str(self.__iosettings['compression']))
+            msg = "%s compression may not be supported by this version of h5py." % str(self.__iosettings['compression'])
+            if not allow_plugin_filter:
+                msg += "Set `allow_plugin_filters=True` to enable the use of dynamically-loaded plugin filters."
+            raise ValueError(msg)
         # Check possible parameter collisions
         if isinstance(self.data, Dataset):
             for k in self.__iosettings.keys():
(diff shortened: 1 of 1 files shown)


src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -427,7 +427,8 @@ def __init__(self, **kwargs):
         # Confirm that the compressor is supported by h5py
         if not self.filter_available(self.__iosettings.get('compression', None),
                                      self.__allow_plugin_filters):
-            raise ValueError("%s compression not support by this version of h5py." %
+            raise ValueError("%s compression may not be supported by this version of h5py. Use "
+                             "`allow_plugin_filters=True` to bypass this check" %
```
oruebel: ``allow_plugin_filter=True`` does not actually bypass the check it, simply allows also non-standard filters to be used. I believe the more appropriate error message would be.

```
msg = "%s compression may not be supported by this version of h5py." % str(self.__iosettings['compression'])
if not allow_plugin_filter:
    msg += "Set `allow_plugin_filters=True` to enable the use of dynamically-loaded plugin filters."
raise ValueError(msg)
```

We require users to explicitly indicate that they want to use  custom filters because the use of such filters breaks interoperability of HDF5 files, i.e., since those filters do not ship with HDF5, the data ecomes unreadable unless whoever needs to read it so has the appropriate filter installed on their system.

bendichter: @oruebel yeah, that works too. I'd also like to add this to the tutorials, with clear notes about lack of compatibility

oruebel (approved):

PR #332, opened 2020-04-16
validate a yaml file against a json schema file.

bendichter:
* Add code for validating a yaml file against a json schema file.
* Add entry point "validate_hdmf_spec"

This will be used to validate changes to the NWB schema as well as HDMF schemas and schema for extensions of either with the command:

```sh
validate_hdmf_spec [...]/nwb-schema/core -m [...]/nwb-schema/nwb.schema.json
```

### DIFF
--- a/src/hdmf/testing/validate_spec.py
+++ b/src/hdmf/testing/validate_spec.py
@@ -1,0 +1,51 @@
+import os
+import ruamel.yaml as yaml
+import json
+from glob import glob
+from argparse import ArgumentParser
+
+import jsonschema
+
+
+def validate_spec(fpath_spec, fpath_schema):
+
+    schemaAbs = 'file://' + os.path.abspath(fpath_schema)
+
+    f_schema = open(fpath_schema, 'r')
+    schema = json.load(f_schema)
+
+    class FixResolver(jsonschema.RefResolver):
+        def __init__(self):
+            jsonschema.RefResolver.__init__(self,
+                                            base_uri=schemaAbs,
+                                            referrer=None)
+            self.store[schemaAbs] = schema
+
+    new_resolver = FixResolver()
+
+    f_nwb = open(fpath_spec, 'r')
+    instance = yaml.safe_load(f_nwb)
+
+    jsonschema.validate(instance, schema, resolver=new_resolver)
+
+
+def main():
+
+    parser = ArgumentParser(description="Validate an HDMF/NWB specification")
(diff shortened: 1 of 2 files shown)


src/hdmf/testing/validate_spec.py
```diff
@@ -0,0 +1,51 @@
...
+import ruamel.yaml as yaml
+import json
+from glob import glob
+from argparse import ArgumentParser
+
+import jsonschema
+
+
+def validate_spec(fpath_spec, fpath_schema):
+
```
oruebel, line 19: Could you add a docstring for ``validate_spec`` method

oruebel (dismissed):
The code looks good to me. It may be useful to add a unit-test that runs the validate_spec method on the hdmf-common schema that is included with the HDMF repo and then we could run that test also as a regression test on the hdmf-common (an nwb-schema repo) to make sure updates to the schema in those repos pass tests.

bendichter:
@oruebel yes, I totally agree that we should apply this as a regression test to this repo and to hdmf-common, but doing that requires that we add `hdmf.json.schema` to the `hdmf-common` repo. When I do that, I'd like to add a regression test to the `hdmf-common` repo at the same time, but that requires this code to be in place first. 

Essentially, all the changes we want to make require changes to a number of different repos and therefore must be done in steps. I'd like to get this code in place first because it allows us to actually *run* any tests we add.

bendichter:
@oruebel I take it back, I think I can implement that test. Hold on a sec

bendichter:
@oruebel I could run this test, but the necessary json file is in https://github.com/hdmf-dev/hdmf-schema-language, so I would need to make that a submodule either of this repo or of `hdmf-common`. Or I guess I could `git clone` that repo within the test. I definitely want to do all of this, but I think piecewise will make it more manageable. We've been running into issues with nwb schema alterations that are not being validated, so I want to get this running over there asap

oruebel:
> I think piecewise will make it more manageable

Agreed. this was more a suggestion for useful thing to do rather than something that would be required as part of this PR. I'd suggest to open an issue about adding the tests and  merging this PR as is.

oruebel (approved):

oruebel:
> I would need to make that a submodule either of this repo or of `hdmf-common`

I think making it a submodule of hdmf (rather than the schema) is probably the way to go

rly (approved):

PR #333, opened 2020-04-16
Update requirements-min.txt

bendichter:
We need this for the validator code to run


oruebel (dismissed):

rly:
Can you test whether `jsonschema==2.6.0` would suffice? To set minimum package requirements, I first check whether the latest supported package on debian would work: https://packages.debian.org/search?keywords=jsonschema&searchon=names&suite=stable&section=all

If 2.6.0 doesn't work, we can use 3.0.1.

oruebel (approved):

PR #335, opened 2020-04-21
Add allowed value / enum validation in docval and refactor some checks

### LINKED ISSUE #334: Add enum checking for values in docval
Suggestion from @bendichter in https://github.com/NeurodataWithoutBorders/pynwb/pull/1226 

We do not yet have enumerated types in the schema, but we can implement basic checks for these in the PyNWB / HDMF API when constructing a new neurodata_type / Container object. This eliminates the need for the same code in several places in PyNWB where the kwargs value is checked against a list of possible allowed values. 

We propose to add the key `enum` (or `allowed_values`) to the `docval` function. We prefer `enum` to match the keys in [json-schema](https://json-schema.org/).

Alternatively, we could add a separate value validator decorator, e.g., `@valueval` instead of packaging this in the already bloated `docval`. (Shape validation could also be a separate validator decorator, but a PyNWB function with 3 decorators and long argument lists would get hard to read.)

rly:
Resolve #334.

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -17,9 +17,9 @@ AllowPositional = Enum('AllowPositional', 'ALLOWED WARNING ERROR')
 __supported_bool_types = (bool, np.bool_)
 __supported_int_types = (int, np.int8, np.int16, np.int32, np.int64)
 __supported_float_types = [float, np.float16, np.float32, np.float64]
-if hasattr(np, "float128"):
+if hasattr(np, "float128"):  # pragma: no cover
     __supported_float_types.append(np.float128)
-if hasattr(np, "longdouble"):
+if hasattr(np, "longdouble"):  # pragma: no cover
     # on windows python<=3.5, h5py floats resolve float64s as either np.float64 or np.longdouble
     # non-deterministically. a future version of h5py will fix this. see #112
     __supported_float_types.append(np.longdouble)
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -14,6 +14,17 @@ __macros = {
 # code to signify how to handle positional arguments in docval
 AllowPositional = Enum('AllowPositional', 'ALLOWED WARNING ERROR')
 
+__supported_bool_types = (bool, np.bool_)
+__supported_int_types = (int, np.int8, np.int16, np.int32, np.int64)
+__supported_float_types = [float, np.float16, np.float32, np.float64]
+if hasattr(np, "float128"):
+    __supported_float_types.append(np.float128)
+if hasattr(np, "longdouble"):
+    # on windows python<=3.5, h5py floats resolve float64s as either np.float64 or np.longdouble
+    # non-deterministically. a future version of h5py will fix this. see #112
+    __supported_float_types.append(np.longdouble)
+__supported_float_types = tuple(__supported_float_types)
+
 
 def docval_macro(macro):
     """Class decorator to add the class to a list of types associated with the key macro in the __macros dict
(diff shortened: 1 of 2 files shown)


src/hdmf/utils.py
```diff
@@ -121,6 +125,51 @@ def __format_type(argtype):
...
 
 
+def __check_enum_argtype(argtype):
+    """Return True/False whether the given argtype is a supported docval enum type (bool, ints, floats, str)"""
+    if isinstance(argtype, (list, tuple)):
+        for i in argtype:
+            if not __check_enum_argtype(i):
+                return False
+        return True
+    return argtype in __supported_bool_types + __supported_int_types + __supported_float_types + (str, )
```
bendichter: I'm a bit confused by this function. Why is it recursive? We already have the line 148 that makes sure we have a list or a tuple. I also think we should perform a check that the enum values match the indicated `dtype` arg.

bendichter (changes requested):

bendichter:
@rly thanks for working on this. The tests look great! Definitely the usage pattern I was imagining and they cover a lot of ground. I was just a bit confusing about some of the internal logic.

src/hdmf/utils.py
```diff
@@ -121,6 +125,51 @@ def __format_type(argtype):
...
+    Helper function to check whether the given argument value validates against the enum specification.
+
+    :param argval: argument value passed to the function/method
+    :param arg: argument validator - the specification dictionary for this argument
+
+    :return: None if the value validates successfully, error message if the value does not.
+    """
+    allowed_vals = arg['enum']
+    if not isinstance(allowed_vals, (list, tuple)):
+        msg = 'docval for %s: enum value must be a list or tuple (received %s)' % (arg['name'], type(allowed_vals))
```
ajtritt: Is this checking that the specified enum is a valid enum?

bendichter: I think so, and I think maybe this should be in another function. Maybe the logical flow should be

1) validate individual docval args - that the specified enum is a valid enum
2) validate internal consistency of docval - that the types within the supplied enum match the allowed dtypes for that arg
3) validate the supplied data against the specified enum

Would it be better to have these steps be separated out into: 

1) docval validator (makes sure that the docval setup is valid)
2) arg validator (using the valid docval args)

Or is it better to have each of these 3 steps for each check?

src/hdmf/utils.py
rly: This checks whether the types listed under the `'type': (bool, int)` or `'type': str` part of the docval argument are allowed for enum validation. If the argument allows a `TimeSeries` to be passed in, for example, this check would fail. I agree - this does not need to be recursive. I'll simplify it. I just wanted to handle both the case where a type is passed in and a list of types is passed in.

Line 148 checks whether the values under the `'enum': (True, 1)` is a list/tuple (it would not make sense to have only one allowed value).

src/hdmf/utils.py
rly: Yeah, I think it makes sense to have a separate docval validator to check for internal consistency and an arg validator. I think the docval validator should be done within the `dec` function on line 497, so it is not called when the function is called but rather when the module is loaded. I will refactor that.

rly: The flow is now:
When the module is loaded, docval internal check:
1. check that enum values is a list or tuple
2. check that argument type is compatible with use of enum (if type is allowed to be TimeSeries, then enum is not allowed)
3. check that each enum value is of a type listed in the argument types

When the decorated function is called, check that the supplied data is one of the values listed in the enum values.

ajtritt (approved):
Looks good to me. 

`docval` is becoming its own beast. Given that `utils.py` is getting pretty big, we might want to pull out `docval` eventually

bendichter (approved):

PR #336, opened 2020-04-21
Add logging of build and hdf5 write process

rly:
This PR adds use of python's `logging` module to help debug the complex build and hdf5 write process. 

Users can use the following code to pipe the log messages to a file.
```python
logger = logging.getLogger()
logger.setLevel(logging.DEBUG)

ch = logging.FileHandler('test.log', mode='w')
ch.setLevel(logging.DEBUG)
formatter = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
ch.setFormatter(formatter)
logger.addHandler(ch)
```

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -3,6 +3,7 @@ import numpy as np
 import os.path
 from functools import partial
 from h5py import File, Group, Dataset, special_dtype, SoftLink, ExternalLink, Reference, RegionReference, check_dtype
+import logging
 import warnings
 
 from ...container import Container
@@ -41,6 +42,7 @@ class HDF5IO(HDMFIO):
 
         For `mode`, see `h5py.File <http://docs.h5py.org/en/latest/high/file.html#opening-creating-files>_`.
         '''
+        self.logger = logging.getLogger('%s.%s' % (self.__class__.__module__, self.__class__.__qualname__))
         path, manager, mode, comm, file_obj = popargs('path', 'manager', 'mode', 'comm', 'file', kwargs)
 
         if file_obj is not None and os.path.abspath(file_obj.filename) != os.path.abspath(path):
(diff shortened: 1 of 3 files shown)


bendichter:
This seems like it could be really helpful. Would you mind demonstrating an example output?

rly:
Here is an example log file from writing and reading a simple NWB file. I've added this logging to help with fixing a specific bug for reference resolution and building tables, so the log messages are a little biased toward that code flow. As other bugs come up, I will add log messages in the other parts of the codebase.

[test.log](https://github.com/hdmf-dev/hdmf/files/4512086/test.log)

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -41,6 +42,7 @@ def __init__(self, **kwargs):
 
         For `mode`, see `h5py.File <http://docs.h5py.org/en/latest/high/file.html#opening-creating-files>_`.
         '''
+        self.logger = logging.getLogger('%s.%s' % (self.__class__.__module__, self.__class__.__qualname__))
```
ajtritt, line 45: This seems like a reasonable default logger, but how does a user specify their own logger?

ajtritt (changes requested):
Please change the default logger being used.

rly:
@ajtritt sorry, I am a little confused. Each module defines its own logger through:
`self.logger = logging.getLogger('%s.%s' % (self.__class__.__module__, self.__class__.__qualname__))`
e.g., `self.logger = logging.getLogger('hdmf.build.manager.BuildManager')`

Should this be changed?

I used the default logger in the example code of this PR but that's up to the end-user to set.

ajtritt (approved):

rly:
Just to document. @ajtritt and I talked about this. We decided that it would be useful to include a way for the user to pass their own custom logger through the HDMF build process, e.g., through `HDMFIO` and `BuildManager`, which passes the logger to `ObjectMapper` during the `build` and `construct` processes. However, that would require touching a lot of code and it is not clear if that use case is desired yet, so we will hold off on implementing that until needed.

Also, note that the default logger set up for `HDF5IO`, `BuildManager`, and `ObjectMapper` are named according to the full name of the class and *not* the name of the module, as is commonly done. So configuring a set of loggers using their name and dot notation, e.g., through `logger = logging.getLogger('hdmf.backends')` will configure only loggers in classes that have a full name that starts with 'hdmf.backends'. This allows for specificity, but means that subclasses in other modules, such as 'pynwb.NWBHDF5IO', need to be configured separately.

oruebel:
Just to clarify, if I understand your comment correctly, if one wants to use a custom logger it means they need to set it on the relevant classes individually rather it being an argument to pass into a class. For now, I think that's Ok, I just wanted to confirm that is what you are suggesting.

ajtritt:
I think that's correct. Thought, I think this use case is rare. I am guessing that users might want to configure the file that the logger writes to, which they can do with [logger.basicConfig](https://docs.python.org/3/library/logging.html#logging.basicConfig)

PR #346, opened 2020-05-12
only write a specific namespace version if it does not exist

### LINKED ISSUE #350: Flake8 errors
`flake8==3.8.1` catches a few new flake8 errors that are making CI tests fail:
```
.\src\hdmf\container.py:342:26: E741 ambiguous variable name 'l'
.\src\hdmf\backends\hdf5\h5tools.py:790:5: C901 'HDF5IO.write_dataset' is too complex (31)
.\src\hdmf\backends\hdf5\h5tools.py:1182:29: E741 ambiguous variable name 'l'
.\src\hdmf\backends\hdf5\h5_utils.py:392:69: E225 missing whitespace around operator
.\src\hdmf\build\builders.py:206:13: E741 ambiguous variable name 'l'
.\src\hdmf\common\table.py:201:5: C901 'DynamicTable.__init__' is too complex (32)
.\src\hdmf\common\table.py:558:23: F507 '...' % ... has 1 placeholder(s) but 2 substitution(s)
.\src\hdmf\validate\validator.py:402:5: C901 'GroupValidator.validate' is too complex (21)
.\tests\coloredtestrunner.py:94:26: E741 ambiguous variable name 'l'
.\tests\unit\test_io_hdf5.py:24:17: E722 do not use bare 'except'
.\tests\unit\spec_tests\test_spec_write.py:54:1: E122 continuation line missing indentation or outdented
.\tests\unit\spec_tests\test_spec_write.py:71:1: E122 continuation line missing indentation or outdented
.\tests\unit\spec_tests\test_spec_write.py:200:1: E122 continuation line missing indentation or outdented
```

### LINKED ISSUE #1228: Error updating specification of older files during pynwb.NWBHDF5IO.write()
## Description

When appending data to an NWB file that was written prior to the hdmf-common-schema 1.1.3, updating the specification causes a shape error. Prior to 1.1.3, shape was determined by h5py, but after 1.1.3 shape is set to (). When the specification updates, this error occurs (specifically when updating specifications/core/2.2.4/namespace). The old spec had the shape (1,), while the new spec has the shape ().

```
File "/opt/anaconda3/envs/ipfx/lib/python3.8/site-packages/h5py/_hl/group.py", line 198, in require_dataset
    raise TypeError("Shapes do not match (existing %s vs new %s)" % (dset.shape, shape))
TypeError: Shapes do not match (existing (1,) vs new ())
```

## Steps to Reproduce
Write a file with Pynwb<=1.0.2 which will use hdmf<=1.0.1 
Then in Pynwb > 1.0.2 (I used PyNWB 1.3.0)
```python
import pynwb


fname = 'previous.nwb'

with pynwb.NWBHDF5IO(fname, 'a') as io:
    nwbfile = io.read()
    io.write(nwbfile)
```
## Environment

    Python Executable: Python
    Python Version: Python 3.6 and 3.7
    Operating System: macOS and Linux
    HDMF Version: Pynwb<=1.0.2 with hdmf<=1.0.1 and Pynwb > 1.3.0

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/NeurodataWithoutBorders/pynwb/issues)?
- [x] Have you checked our [Contributing](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/CONTRIBUTING.rst) document?

ajtritt:
## Motivation

Specs were being cached to a file even if they were already cached

help fix https://github.com/NeurodataWithoutBorders/pynwb/issues/1228
fix #350 

## How to test the behavior?
```
import hdmf

fname = 'ANY_HDMF_FILE.h5'

with hdmf.HDF5IO(fname, 'a') as io:
    f = io.read()
    io.write(f)
```

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -555,7 +555,7 @@ class DynamicTable(Container):
         region = getargs('region', kwargs)
         if isinstance(region, slice):
             if (region.start is not None and region.start < 0) or (region.stop is not None and region.stop > len(self)):
-                msg = 'region slice %s is out of range for this DynamicTable of length ' % (str(region), len(self))
+                msg = 'region slice %s is out of range for this DynamicTable of length %d' % (str(region), len(self))
                 raise IndexError(msg)
             region = list(range(*region.indices(len(self))))
         else:
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -389,7 +389,7 @@ class H5DataIO(DataIO):
             )
     def __init__(self, **kwargs):
         # Get the list of I/O options that user has passed in
-        ioarg_names = [name for name in kwargs.keys() if name not in['data', 'link_data', 'allow_plugin_filters']]
+        ioarg_names = [name for name in kwargs.keys() if name not in ['data', 'link_data', 'allow_plugin_filters']]
         # Remove the ioargs from kwargs
         ioarg_values = [popargs(argname, kwargs) for argname in ioarg_names]
         # Consume link_data parameter
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -273,7 +273,7 @@ class H5SpecWriter(SpecWriter):
     def __write(self, d, name):
         data = self.stringify(d)
         # create spec group if it does not exist. otherwise, do not overwrite existing spec
-        dset = self.__group.require_dataset(name, shape=tuple(), data=data, dtype=self.__str_type)
+        dset = self.__group.create_dataset(name, shape=tuple(), data=data, dtype=self.__str_type)
         return dset
 
     def write_spec(self, spec, path):
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -1161,9 +1161,9 @@ class HDF5IO(HDMFIO):
         # dependency
         self.__ref_queue.append(func)
 
-    def __rec_get_ref(self, l):
+    def __rec_get_ref(self, ref_list):
         ret = list()
-        for elem in l:
+        for elem in ref_list:
             if isinstance(elem, (list, tuple)):
                 ret.append(self.__rec_get_ref(elem))
             elif isinstance(elem, (Builder, Container)):
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -294,7 +294,9 @@ class HDF5IO(HDMFIO):
                 ns_builder = self.__convert_namespace(ns_catalog, ns_name)
                 namespace = ns_catalog.get_namespace(ns_name)
                 group_name = '%s/%s' % (ns_name, namespace.version)
-                ns_group = spec_group.require_group(group_name)
+                if group_name in spec_group:
+                    continue
+                ns_group = spec_group.create_group(group_name)
                 writer = H5SpecWriter(ns_group)
                 ns_builder.export('namespace', writer=writer)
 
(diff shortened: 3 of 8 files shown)


oruebel (dismissed):

rly:
If the new group has the same name as the old group, but the dataset contents are different, then I think a warning should be raised, since that could happen if someone is using a schema that is improperly versioned. i.e., '1.0.0' refers to two different schema.

ajtritt:
@rly so, check the contents of every dataset, and warn if they are different? This would happen every time a file is appended to...

src/hdmf/common/table.py
```diff
@@ -555,7 +555,7 @@ def create_region(self, **kwargs):
         region = getargs('region', kwargs)
         if isinstance(region, slice):
             if (region.start is not None and region.start < 0) or (region.stop is not None and region.stop > len(self)):
-                msg = 'region slice %s is out of range for this DynamicTable of length ' % (str(region), len(self))
```
rly, line 558: i'm surprised this wasn't caught by flake8 (or us) before!

rly:
@ajtritt sorry, maybe I wasn't clear. I mean if the group `specifications/[namespace name]/[namespace version]` exists, then read the spec datasets and the namespace datasets inside of it and compare them to the values that would have been written if the group didn't exist. If they differ, raise a warning. On second thought, though, this might be better to do on read, rather than on write.

rly:
Please add tests but this is otherwise good to go

src/hdmf/common/table.py
ajtritt: I'm guessing it was never hit by any code

ajtritt:
@rly I'm not sure we want to do that. This workflow--1) Users loads file with cached spec. 2) User then appends to file 3) Users writes file, caching spec--would be doing a lot of unnecessary checking. We would be checking cached data to see if it matches in-memory data that was read in from that very same cached data. To avoid that, we would need to track the source of namespace that exists in memory. If an in-memory namespace has the same version of a cached namespace, AND the in-memory namespace was read in, then we don't check. Otherwise, we check all the data. If we go that route, then I don't see the point of having a version namespace. If we assume versions can be flexible (i.e. the relationship between namespace version and namespace specification contents is not one-to-one), then there's not really a point in version. We have to revert to calling the contents of the namespace the version. 


> Please add tests but this is otherwise good to go

This code is already hit by existing tests.

rly:
I agree - this would be a lot of checking, and yeah we would want to add code to check whether the in-memory namespace was read in or loaded from elsewhere, but it would help catch cases where a developer does not update their version number before release. 

I have already seen this to some degree in the NDX Catalog, where I request a change in an extension and the developer makes the change, but does not update the version. I can easily see it happening for a user who creates their own extension, writes a file with cached spec, updates their extension locally without changing the version number (let's say to add a new type), appends to the file, and wonders why their new type isn't written. It's not critical to have the check, and it's the extension developer's fault,  but it could help protect users. I think the computational cost of the check would be relatively low. In any case, I think this can be solved in a subsequent PR.

ajtritt:
@rly 

> I have already seen this to some degree in the NDX Catalog, where I request a change in an extension and the developer makes the change, but does not update the version.

maybe we should check for that and prevent it somehow. My concern is that if the version-spec relationship is not 1-1, then how do we know which spec is the right one?

Also, do you know why flake8 is now complaining about using `l` as a function argument name? It's flagging code that was written years ago...

oruebel:
> I have already seen this to some degree in the NDX Catalog

For the catalog we could make this part of the CI since in the majority of cases and update there is going to go along with an update of the version number of the extension.

oruebel:
> Also, do you know why flake8 is now complaining

Did maybe the version of flake8 change?

rly:
> My concern is that if the version-spec relationship is not 1-1, then how do we know which spec is the right one?

We don't -- I suggest just raising a warning that says the cached spec does not match the loaded spec with the same name and version.

For the catalog, yes, we can check for this. I worry about the homebrew extension case that never makes it to the catalog. 

re: flake8, probably the version changed and was updated to catch this: https://www.flake8rules.com/rules/E741.html

oruebel (approved):

rly (approved):

PR #347, opened 2020-05-15
add API class for reading VocabData

ajtritt:
## Motivation

Add API class for working with VocabData data_type

## How to test the behavior?
```
```


ajtritt:
included in #355

PR #355, opened 2020-05-21
Add table functionality

ajtritt:
## Motivation

The original motivation for this PR was to add the ability to write table columns from DataChunkIterators. This work went stale and when it was brought up to date, some other issues arose. In response to those issues, a couple more features were added.

1. Add ability to write `DynamicTable` columns from a `DataChunkIterator`
2. Add `DynamicTable.get` method. This method behaves like `__getitem__`/`[]`, but can take optional arguments to modify the behavior. Currently, the only argument that exists is `df`, which will control if a selection is returned as a `pandas.DataFrame`. 
    - The work of selecting from a DynamicTable is carried out by `DynamicTable.get`. `DynamicTable.__getitem__` calls `DynamicTable.get` with whatever options are necessary to provide familiar functionality. Moving forward, this pattern with `get`/`__getitem__` will be used to provide more flexibility for selecting data from a `DynamicTable`. See `VocabData` for another example.
3. `DynamicTable.__getitem__` functionality wasn't fully tested and had some issues. One major issue was that returning a `pd.DataFrame` did not work if one of the columns was multi-dimensional or ragged. This PR will resolve those issues.
4. Add `VocabData` column type for storing controlled-vocabulary data.


src/hdmf/common/table.py
```diff
@@ -79,46 +80,91 @@ class VectorIndex(Index):
...
+            unsigned integer encoding of idx
+        """
+        if idx > self.__maxval:
+            nbits = (np.log2(self.__maxval + 1)*2)
+            self.__uint = np.dtype('uint%d' % nbits).type
+            self.__maxval = 2**nbits - 1
+            self.__adjust_precision(self.__uint)
+        return self.__uint(idx)
+
+    def __adjust_precision(self, uint):
```
ajtritt, line 114: @rly I added this because precision was not automatically getting adjusted. According to this function, it should have been.

https://github.com/hdmf-dev/hdmf/blob/712a50016fdcd015d9cd712b6dcc8434b6487c92/src/hdmf/build/objectmapper.py#L122-L129

This was the cause of failure of this test:
https://app.circleci.com/pipelines/github/hdmf-dev/hdmf/1489/workflows/d85924fb-11ac-455c-a2bd-31322ab95065/jobs/17977

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1192,6 +1195,15 @@ def __rec_get_ref(self, ref_list):
                 ret.append(elem)
         return ret
 
+    @classmethod
+    @docval(*get_docval(H5DataIO.__init__))
+    def set_dataio(cls, **kwargs):
+        """
```
oruebel, line 1201: Can you clarify where this function is being used? Also the name implies that dataio is being set but the docstring implies that it is a generator method to create an H5DataIO object?

src/hdmf/build/objectmapper.py
```diff
@@ -209,7 +209,12 @@ def convert_dtype(cls, spec, value):
...
         elif isinstance(value, AbstractDataChunkIterator):
             ret = value
-            ret_dtype, warning_msg = cls.__resolve_dtype(value.dtype, spec_dtype)
+            if spec_dtype is _unicode:
+                ret_dtype = "utf8"
+            elif spec_dtype is _ascii:
+                ret_dtype = "ascii"
+            else:
+                ret_dtype, warning_msg = cls.__resolve_dtype(value.dtype, spec_dtype)
         else:
```
oruebel, line 226: Why are _unicode and _ascii handled separately here rather than in ``__resolve_dtype``?

src/hdmf/backends/hdf5/h5tools.py
ajtritt: I use it some other code. It's mainly for convenience so you don't have to import H5DataIO. I can rename it or add more clarity to the docstring if you prefer

src/hdmf/build/objectmapper.py
ajtritt: __resolve_dtype is used for determining the precision of a _numerical dtype_, so it doesn't work for string types.

src/hdmf/container.py
```diff
@@ -407,13 +408,32 @@ def set_dataio(self, **kwargs):
         dataio.data = self.__data
         self.__data = dataio
 
+    @docval({'name': 'func', 'type': types.FunctionType, 'doc': 'a function to transform *data*'})
+    def transform(self, **kwargs):
+        """
```
ajtritt, line 413: Users should be warned about two things in the docstring:
1. This transformation will happen _in place_
2. Any transformation to the underlying data may change subsequent behavior.

src/hdmf/container.py
```diff
@@ -407,13 +408,32 @@ def set_dataio(self, **kwargs):
...
+        func = getargs('func', kwargs)
+        self.__data = func(self.__data)
+
     def __bool__(self):
-        return len(self.data) != 0
+        if self.data is not None:
+            if isinstance(self.data, (np.ndarray, tuple, list)):
+                return len(self.data) != 0
+            if self.data:
```
ajtritt, line 426: This will handle cases when `self.data` is a`DataIO`

rly:
#364 might help with some of the uint column issues. When a `VectorIndex` is built with signed integers, a warning will be raised and the signed integers will be converted to unsigned integers. It does the conversion in a less smart way than what you have here, but it's simpler and used more broadly, which is a plus.

src/hdmf/common/table.py
rly: I think we would still want to store uints in the `VectorIndex` to avoid the conversion warning (or we could just remove the warning), so all of this uint code would still be used...

PR #361, opened 2020-05-23
Improve `get_class` and `docval` support for uint

### LINKED ISSUE #360: `get_class` fails for uncommon dtypes
## Description

Classes can be dynamically generated **only** for the types listed here:
https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/build/manager.py#L380-L390

which leaves out a lot of the possible dtypes listed in https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype

If a type outside of those is used, a KeyError results when `self._type_map.get(spec.dtype)` is called.

## Environment

    Python Executable: Python
    Python Version: 3.7
    Operating System: Windows
    HDMF Version: dev

## Checklist

rly:
Fix #360.

Also add 'uint' docval macro and allow uint in `docval` enum types.

### DIFF
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -377,16 +377,35 @@ class TypeMap:
                     self.register_container_type(new_ns, dt, container_cls)
         return deps
 
-    _type_map = {
-        'text': str,
-        'float': float,
-        'float32': float,
-        'float64': float,
-        'int': int,
-        'int32': int,
-        'bool': bool,
+    # mapping from spec types to allowable python types for docval for fields during dynamic class generation
+    # see https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype
+    _spec_dtype_map = {
+        'float32': (float, np.float32, np.float64),
+        'float': (float, np.float32, np.float64),
+        'float64': (float, np.float64),
+        'double': (float, np.float64),
+        'int8': (np.int8, np.int16, np.int32, np.int64, int),
+        'int16': (np.int16, np.int32, np.int64, int),
+        'short': (np.int16, np.int32, np.int64, int),
+        'int32': (int, np.int32, np.int64),
+        'int': (int, np.int32, np.int64),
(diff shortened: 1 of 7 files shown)


src/hdmf/build/manager.py
```diff
@@ -377,16 +377,35 @@ def load_namespaces(self, **kwargs):
...
-    _type_map = {
+    # mapping from spec types to allowable python types for docval for fields during dynamic class generation
+    # see https://schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype
+    _spec_dtype_map = {
+        'float32': (float, np.float32, np.float64),
+        'float': (float, np.float32, np.float64),
+        'float64': (float, np.float64),
+        'double': (float, np.float64),
+        'int8': (np.int8, np.int16, np.int32, np.int64, int),
+        'int16': (np.int16, np.int32, np.int64, int),
```
oruebel, line 393: I'm not exactly sure exactly how this is used, but it seems imply that both upconversion and downcoversion of precision would be allowed? E.g., doesn't allowing an ``np.int64`` for an ``int8`` lead to a warning on write?

rly: > doesn't allowing an np.int64 for an int8 lead to a warning on write?
An `int64` is allowed for an `int8` spec on write without a warning because `int8` is the minimum precision.

This change says that if a dataset/attribute spec has dtype `int32`, then `get_class` should generate a `docval` for the class `__init__` that allows int32 and int64 for the corresponding argument. There is no conversion here. Passing an int16 to `__init__` would raise a docval type error. Passing an int64 to `__init__` results in the value being stored as an int64 (and subsequently written as an int64). Is that any clearer?

oruebel: Got it. It may be useful to add this as a comment in the code as well.

oruebel (dismissed):

oruebel (approved):

PR #367, opened 2020-05-28
allow for optional path and mode in HDF5IO

### LINKED ISSUE #366: HDF5IO should be able to take only File object
Right now, when you call HDF5IO, you can supply a File object, but you also need to specify a path, the code files if the filename of the object does not match the path. It does not make much sense to me why you would need to specify both the path and the File object. One or the other should be enough

I am trying to read an HDF5 file on S3 using [s3fs-fuse](https://github.com/s3fs-fuse/s3fs-fuse). I have made the adjustments to hdmf and pynwb to confirm that this will work, we just need to make this modification, because in this case I have a File object but no local 'path'.

I would like to change this so that you can enter a file arg and no path arg. In order for this to work, both the `path` and the `mode` arguments will need to become optional.

bendichter:
fix #366

* [x] implement feature
* [ ] add tests


oruebel (dismissed):

bendichter:
We decided it would be better to refactor this so that the function that takes the path uses it to open and file and then there is a separate function that initializes the class using that file

rly:
This PR is superseded by #746

Issue #327, opened 2020-04-09
Add the ability to store and resolve foreign fields


ajtritt:
Add the ability to reference data that is stored across the web. For example, a user may want to distribute data in a file, but include a reference to metadata stored in database that is accessible thru a REST API.

rly:
@oruebel @ajtritt Do we want to continue development of this feature? It is partially resolved by HERD but not entirely because I think the initial spec included the ability to resolve/read linked arbitrary JSON files and NWB containers within NWB files on the internet.

oruebel:
A couple of thoughts for potential more specific follow-up tickets. To support the notion of foreign field, I think there a couple of simple things we could do in HERD to make it more general:

1. Support links for data objects (i.e., associate a link with a dataset or group directly without requiring a term). This could be done in HERD simply by having a NaN value for the term, e.g., using empty string for the term. 
2. Support defining the "type" of the relationship. HERD currently supports linking terms to their definitions. By adding a column to the `EntityKeyTable` we could indicate what type of relationship exists between the given entity and the key

@rly @mavaylon1 what do you think?

> I think the initial spec included the ability to resolve/read linked arbitrary JSON files and NWB containers within NWB files on the internet

I think that particular aspect is better dealt with on the storage side as part of a potentially more modular storage backend.

> Do we want to continue development of this feature?

I think we can close this for now as it is superseded by HERD and the data storage discussions

mavaylon1:
Agreed

oruebel:
> Agreed

In this case, can you please create issues for item 1 and 2 so we don't forget

rly:
I opened #1484 and #1485 to track the two items above.

PR #371, opened 2020-06-02
Fix adding of optional predefined columns to `DynamicTable`

### LINKED ISSUE #357: Clarify and fix 'colnames' argument in `DynamicTable.__init__`
## Description

From the Allen Institute:

1. In `DynamicTable.__init__`, if the keyword argument 'colnames' is provided and 'columns' is not provided, a cryptic `KeyError` is raised. This line:
https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/common/table.py#L265
should be reached but is unreachable because of the earlier logic setting `columns` to list() if `columns` is None.

2. The documentation for the 'colnames' argument for `DynamicTable.__init__` is not clear. It should say that it is used to specify the order of the columns in the 'columns' argument.

## Steps to Reproduce

```python
DynamicTable(name="test", description="description", colnames=['a', 'b'])
```
## Environment

    Python Executable: Python
    Python Version: Python 3.7
    Operating System: Windows
    HDMF Version: dev

## Checklist

### LINKED ISSUE #359: Cannot write references when root container is not called "root"
## Description

The below code creates a table, adds an indexed column, adds a row, and writes the table to file. 

```python
from hdmf.common import DynamicTable, get_manager
from hdmf.backends.hdf5 import HDF5IO

table = DynamicTable('table0', 'an example table')
table.add_column('qux', 'a boolean column', index=True)
table.add_row(qux=[True, True])

filename = 'test.h5'
manager = get_manager()
with HDF5IO(filename, manager=manager, mode='w') as io:
    io.write(table)

with HDF5IO(filename, manager=manager, mode='r') as io:
    io.read()
```

Reading the file results in:

```
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 592, in __add_refs
    call()
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 718, in _filler
    obj.attrs[key] = self.__get_ref(value)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 560, in func_call
    return func(args[0], **pargs)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 1157, in __get_ref
    return self.__file[path].ref
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "C:\Users\Ryan\Miniconda3\envs\dev\lib\site-packages\h5py\_hl\group.py", line 264, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\h5o.pyx", line 190, in h5py.h5o.open
KeyError: 'Unable to open object (component not found)'

  File "test_table.py", line 11, in <module>
    io.write(table)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 560, in func_call
    return func(args[0], **pargs)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 302, in write
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 404, in call_docval_func
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 560, in func_call
    return func(args[0], **pargs)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\io.py", line 44, in write
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\utils.py", line 560, in func_call
    return func(args[0], **pargs)
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 574, in write_builder
    self.__add_refs()
  File "c:\users\ryan\documents\nwb\hdmf\src\hdmf\backends\hdf5\h5tools.py", line 595, in __add_refs
    raise RuntimeError('Unable to resolve reference')
RuntimeError: Unable to resolve reference
```

## Environment

    Python Executable: Python
    Python Version: 3.7
    Operating System: Windows
    HDMF Version: dev

## Checklist

### LINKED ISSUE #370: Adding optional predefined column should raise warning not error
## Description

With PR #305, adding a column that is an optional column defined in `__columns__` raises an error. However, this new error breaks existing code such as in the AllenSDK: https://travis-ci.org/github/AllenInstitute/AllenSDK/jobs/692971698
where they get the error: `ValueError: column 'spike_times' already exists in Units 'units'` after calling `Units.add_column`. This had previously worked for optional columns without complaint.

Raising this error is a breaking change on code that is not clearly wrong and so should be changed to a warning until the next major version. Users should also be able to add a predefined optional column with a different description than the default supplied by PyNWB, if desired (the column description is typically not specified in the NWB spec).

## Steps to Reproduce

```python


class SubTable(DynamicTable):

    __columns__ = (
        {'name': 'col2', 'description': 'optional column', 'required': False},
    )


table = SubTable(name='subtable', description='subtable description')
table.add_column('col2', 'column #2')
```
## Environment

    Python Executable: Python
    Python Version: Python 3.7
    Operating System: Windows
    HDMF Version: dev

## Checklist

rly:
## Motivation

Fix #370, fix #357.

Also do minor refactoring, add comments, and improve table test coverage. Total coverage is up 0.5%!

## How to test the behavior?

See `test_table.py`

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -330,7 +330,7 @@ class DynamicTable(Container):
         self._init_class_columns()
 
     def __set_table_attr(self, col):
-        if hasattr(self, col.name):
+        if hasattr(self, col.name) and col.name not in self.__uninit_cols:
             msg = ("An attribute '%s' already exists on %s '%s' so this column cannot be accessed as an attribute, "
                    "e.g., table.%s; it can only be accessed using other methods, e.g., table['%s']."
                    % (col.name, self.__class__.__name__, self.name, col.name, col.name))
@@ -354,8 +354,13 @@ class DynamicTable(Container):
                                     **{k: col[k] for k in col.keys()
                                         if k not in ['name', 'description', 'index', 'table', 'required']})
                 else:
-                    # track the not yet initialized optional columns
+                    # track the not yet initialized optional predefined columns
                     self.__uninit_cols[col['name']] = col
+
+                    # set the table attributes for not yet init optional predefined columns
+                    setattr(self, col['name'], None)
+                    if col.get('index', False):
+                        setattr(self, col['name'] + '_index', None)
 
     @staticmethod
     def __build_columns(columns, df=None):
@@ -542,7 +547,7 @@ class DynamicTable(Container):
             # else, the ObjectMapper will create a link from self (parent) to col_index (child with existing parent)
             col = col_index
             self.__indices[col_index.name] = col_index
-            setattr(self, col_index.name, col_index)
+            self.__set_table_attr(col_index)
 
         if len(col) != len(self.id):
             raise ValueError("column must have the same number of rows as 'id'")
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -365,6 +365,7 @@ class DynamicTable(Container):
                     # set the table attributes for not yet init optional predefined columns
                     setattr(self, col['name'], None)
                     if col.get('index', False):
+                        self.__uninit_cols[col['name'] + '_index'] = col
                         setattr(self, col['name'] + '_index', None)
 
     @staticmethod
@@ -535,6 +536,8 @@ class DynamicTable(Container):
         col.parent = self
         columns = [col]
         self.__set_table_attr(col)
+        if col in self.__uninit_cols:
+            self.__uninit_cols.pop(col)
 
         # Add index if it's been specified
         if index is not False:
@@ -555,6 +558,8 @@ class DynamicTable(Container):
             col = col_index
             self.__indices[col_index.name] = col_index
             self.__set_table_attr(col_index)
+            if col_index in self.__uninit_cols:
+                self.__uninit_cols.pop(col_index)
 
         if len(col) != len(self.id):
             raise ValueError("column must have the same number of rows as 'id'")
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -504,16 +504,18 @@ class DynamicTable(Container):
         if name in self.__uninit_cols:  # column is a predefined optional column from the spec
             # check the given values against the predefined optional column spec. if they do not match, raise a warning
             # and ignore the given arguments. users should not be able to override these values
+            table_bool = table or not isinstance(table, bool)
             spec_table = self.__uninit_cols[name].get('table', False)
-            if table != spec_table:
+            if table_bool != spec_table:
                 msg = ("Column '%s' is predefined in %s with table=%s which does not match the entered "
                        "table argument. The entered table argument will be ignored."
                        % (name, self.__class__.__name__, spec_table))
                 warn(msg)
                 table = spec_table
 
+            index_bool = index or not isinstance(index, bool)
             spec_index = self.__uninit_cols[name].get('index', False)
-            if index != spec_index:
+            if index_bool != spec_index:
                 msg = ("Column '%s' is predefined in %s with index=%s which does not match the entered "
                        "index argument. The entered index argument will be ignored."
                        % (name, self.__class__.__name__, spec_index))
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -195,7 +195,7 @@ class DynamicTable(Container):
         if (len(bases) and 'DynamicTable' in globals() and issubclass(bases[-1], Container)
                 and bases[-1].__columns__ is not cls.__columns__):
             new_columns = list(cls.__columns__)
-            new_columns[0:0] = bases[-1].__columns__
+            new_columns[0:0] = bases[-1].__columns__  # prepend superclass columns to new_columns
             cls.__columns__ = tuple(new_columns)
 
     @docval({'name': 'name', 'type': str, 'doc': 'the name of this table'},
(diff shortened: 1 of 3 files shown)


src/hdmf/common/table.py
```diff
@@ -218,30 +222,31 @@ def __init__(self, **kwargs):  # noqa: C901
         else:
             id = ElementIdentifiers('id')
 
-        if columns is not None:
-            if len(columns) > 0:
```
rly, line 222: I just combined these if statements

src/hdmf/common/table.py
```diff
@@ -218,30 +222,31 @@ def __init__(self, **kwargs):  # noqa: C901
...
-                        colset.pop(c.target.name)
-                lens = [len(c) for c in colset.values()]
-                if not all(i == lens[0] for i in lens):
-                    raise ValueError("columns must be the same length")
-                if lens[0] != len(id):
-                    if len(id) > 0:
-                        raise ValueError("must provide same number of ids as length of columns")
-                    else:
-                        id.data.extend(range(lens[0]))
-        else:
```
rly, line 241: I removed this else statement because `columns` is checked against None later on and handled there. With this line, `columns` can never be None

src/hdmf/common/table.py
```diff
@@ -470,37 +487,39 @@ def __eq__(self, other):
...
+                table = spec_table
+
+            spec_index = self.__uninit_cols[name].get('index', False)
+            if index != spec_index:
+                msg = ("Column '%s' is predefined in %s with index=%s which does not match the entered "
+                       "index argument. The entered index argument will be ignored."
+                       % (name, self.__class__.__name__, spec_index))
+                warn(msg)
+                index = spec_index
+
```
rly, line 522: Note that this behavior differs from when columns are specified via `DynamicTable.__init__`. In that case, the given columns override the predefined columns. I don't think that overriding should be allowed, or at least a warning should be raised. See new line #327 for the TODO.

src/hdmf/common/table.py
```diff
@@ -470,37 +487,39 @@ def __eq__(self, other):
...
             msg = "column '%s' already exists in %s '%s'" % (name, self.__class__.__name__, self.name)
             raise ValueError(msg)
 
+        if name in self.__uninit_cols:  # column is a predefined optional column from the spec
+            # check the given values against the predefined optional column spec. if they do not match, raise a warning
+            # and ignore the given arguments. users should not be able to override these values
+            spec_table = self.__uninit_cols[name].get('table', False)
+            if table != spec_table:
+                msg = ("Column '%s' is predefined in %s with table=%s which does not match the entered "
+                       "table argument. The entered table argument will be ignored."
```
njmei: @rly I'm a bit wary of this different behavior. It seems like with the 1.6.1 version of `hdmf` one could use `add_column()` to overwrite an existing column. This may not have been the *intended* behavior, but since it was allowed, it is effectively part of the `feature set` of version < 2.0.0 of `hdmf`.

I see a lot of instances in the AllenSDK where a `Table` is created and filled with some placeholder column and then real data for those columns are used to overwrite the placeholder column in the `Table` using `add_column()`.

See: https://github.com/AllenInstitute/AllenSDK/blob/c0d5d40be5c48e089b7121b2dcb4557504aa2dd4/allensdk/brain_observatory/ecephys/write_nwb/__main__.py#L718

https://github.com/AllenInstitute/AllenSDK/blob/c0d5d40be5c48e089b7121b2dcb4557504aa2dd4/allensdk/brain_observatory/ecephys/write_nwb/__main__.py#L458

Is there a workaround such as being able to delete specified existing columns if we *do* want the ability to overwrite existing columns with `add_column()`?

njmei: Maybe adding an additional default parameter to `add_column()` like `overwrite_existing=False`?

rly: @njmei You raise a valid concern. However, let's take the `Units` table and the optional `spike_times` dataset as an example. The `spike_times` column is specified in the NWB spec as an optional, ragged array column. The `Units` class predefines the `spike_times` column but does not add it to the `Units` table until the user either adds the column manually or adds a row with a value for `spike_times`. The check in the code here is to ensure that if the user adds the `spike_times` column, that certain column properties match the predefined column in the `Units` class, which conforms with the NWB spec. If the user is allowed to add the `spike_times` column but make it not a ragged array or make it a table region reference, then the resulting file will not be compliant with the NWB spec. 

This change does break existing functionality, but this was uncaught buggy functionality -- HDMF should not allow users to write files that are non-compliant with the spec. As long as the code that uses `add_column` does not use it to create non-compliant files, it should not be affected. Let me know if you find that not to be the case. @ajtritt @oruebel feel free to chime in.

rly: I decided to revert the behavior and allow overriding of predefined columns with a warning that the user does this at their own risk. It is more dangerous to ignore the passed in `table` or `index` arguments because the `data` rely on those arguments and will not make sense anymore. 

We can make this condition a hard error in a future major version of HDMF.

src/hdmf/common/table.py
```diff
@@ -470,37 +488,47 @@ def __eq__(self, other):
...
         If data is provided, it must contain the same number of rows as the current state of the table.
 
-        :raises ValueError
+        :raises ValueError: if the column has already been added to the table
         """
-        name, data = getargs('name', 'data', kwargs)
+        name, data, description = getargs('name', 'data', 'description', kwargs)
         index, table = popargs('index', 'table', kwargs)
 
+        if isinstance(index, VectorIndex):
```
rly, line 501: @ajtritt In what situation would a user pass a `VectorIndex` to `add_column`? When this happens, `add_column` creates a new `VectorData` column for the given name, description, and data, and also adds the `VectorIndex` which must have a different target `VectorData`, because its target is required in the constructor. I added this warning because I could not figure out how one would use this. It looks intentional too since there is code below that checks `if isinstance(index, VectorIndex):` and `if not isinstance(col_index.parent, Container):`

ajtritt: I am not sure, but I think the warning is appropriate.

ajtritt (approved):

PR #381, opened 2020-06-16
Move written flag from builder to HDF5IO, add tests

rly:
## Motivation

Required for the export PR: #326. 

Currently, `Builder` objects have a field `written` that is used only by `HDF5IO` to mark whether the `Builder` object has been written to disk so that it does not get written multiple times in case of links and references. `Builder` objects should contain minimal state information such as whether it has been written, and this should really be information stored in the IO object. Refactoring to make `HDF5IO` track whether a builder has been written by that object will allow us to use a different `HDF5IO` instance to write the *same builder* to a different file.

Note:
- `HDF5IO.__set_written` is private because no other class should be able to change whether a builder has been written by this IO object. `HDF5IO.get_written` is public because other backends and code may want to be able to know whether a builder has been written to disk, but this could be made private. 
- I replaced the test `test_double_cache_spec` with a new set of tests for writing/appending to check that the written flag is used correctly.

Close #380. 

## How to test the behavior?

See tests.

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -72,6 +72,7 @@ class HDF5IO(HDMFIO):
         self.__ref_queue = deque()  # a queue of the references that need to be added
         self.__dci_queue = deque()  # a queue of DataChunkIterators that need to be exhausted
         ObjectMapper.no_convert(Dataset)
+        self._written_builders = dict()  # keep track of which builders were written (or read) by this IO object
 
     @property
     def comm(self):
@@ -344,6 +345,21 @@ class HDF5IO(HDMFIO):
             self.__read[self.__file] = f_builder
         return f_builder
 
+    def set_written(self, builder):
+        """Mark this builder as written."""
+        # currently all values in self._written_builders are True, so this could be a set but is a dict for
+        # future flexibility
+        builder_id = self.__bldrhash__(builder)
+        self._written_builders[builder_id] = True
+
+    def get_written(self, builder):
+        """Return True if this builder has been written to (or read from) disk by this IO object."""
+        builder_id = self.__bldrhash__(builder)
+        return self._written_builders.get(builder_id, False)
+
+    def __bldrhash__(self, obj):
+        return id(obj)
+
     def __set_built(self, fpath, id, builder):
         """
         Update self.__built to cache the given builder for the given file and id.
@@ -440,7 +456,7 @@ class HDF5IO(HDMFIO):
                         self.__set_built(sub_h5obj.file.filename,  sub_h5obj.file[target_path].id, builder)
                     builder.location = parent_loc
                     link_builder = LinkBuilder(builder, k, source=h5obj.file.filename)
-                    link_builder.written = True
+                    self.set_written(link_builder)
                     kwargs['links'][builder_name] = link_builder
                 else:
                     builder = self.__get_built(sub_h5obj.file.filename, sub_h5obj.id)
@@ -462,7 +478,7 @@ class HDF5IO(HDMFIO):
                 continue
         kwargs['source'] = h5obj.file.filename
         ret = GroupBuilder(name, **kwargs)
-        ret.written = True
+        self.set_written(ret)
         return ret
 
     def __read_dataset(self, h5obj, name=None):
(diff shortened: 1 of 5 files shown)


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -884,10 +913,10 @@ def _filler():
         # NOTE: we can ignore options['io_settings'] for scalar data
         elif self.__is_ref(options['dtype']):
             _dtype = self.__dtypes.get(options['dtype'])
+            self.__set_written(builder)
```
rly: all branches of the if statement below had `builder.written = True` so I moved it to before the if statement

rly:
Also consider the use case where a user writes an in-memory container to a file, then adds to the in-memory container, then writes the updated in-memory container to the same file using the following code:

```python
    def test_write_append_unwritten(self):
        """Test writing a container, adding to the in-memory container, then writing it again in append mode."""
        with HDF5IO(self.path, manager=self.manager, mode='w') as io:
            io.write(self.foofile)

        # append new container
        foo3 = Foo('foo3', [10, 20], "I am foo3", 2, 0.1)
        new_bucket1 = FooBucket('new_bucket1', [foo3])
        self.foofile.buckets.append(new_bucket1)
        new_bucket1.parent = self.foofile

        # write to same file with same manager in APPEND mode
        with HDF5IO(self.path, manager=self.manager, mode='a') as io:
            io.write(self.foofile)
```

Previously this would succeed because the builder for `self.foofile` was already marked as written. Now it will fail with the h5py error `ValueError: Unable to create group (name already exists)` because the second IO object does not know that a group for `self.foofile` already exists in the file. 

Although possibly confusing, I think this is OK behavior. Users should NOT use the above code to write an in-memory container and then append to it. Instead, they should do one of the following:
1) write the in-memory container only once after everything has been added,
2) on the second write, open the file in 'w' mode to overwrite the existing contents, or
3) open the file in 'a' mode after the first write, append to the read container, then call `io.write`

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -344,6 +345,33 @@ def read_builder(self):
...
+
+        :param builder: Builder object to get the written flag for
+        :type builder: Builder
+
+        :return: True if the builder is found in self._written_builders using the builder ID, False otherwise
+        """
+        builder_id = self.__bldrhash__(builder)
+        return self._written_builders.get(builder_id, False)
+
+    def __bldrhash__(self, obj):
```
oruebel: I'd suggest to rename to ``__builderhash``

src/hdmf/backends/hdf5/h5tools.py
oruebel: The written flag should be set after the if/elif statements since that is when the state is modified. E.g., if an error occurs in ``require_dataset`` then the builder is falsely recorded as written.

oruebel (approved):

src/hdmf/backends/hdf5/h5tools.py
rly: Done. (objectmapper.py has a function `__bldrhash__` which is where I copied this from.)

src/hdmf/backends/hdf5/h5tools.py
rly: Ah, that makes sense.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -344,6 +345,33 @@ def read_builder(self):
...
+
+        :param builder: Builder object to get the written flag for
+        :type builder: Builder
+
+        :return: True if the builder is found in self._written_builders using the builder ID, False otherwise
+        """
+        builder_id = self.__builderhash(builder)
+        return self._written_builders.get(builder_id, False)
+
+    def __builderhash(self, obj):
```
ajtritt, line 371: The same function is implemented here: https://github.com/hdmf-dev/hdmf/blob/3fa0587ae2943aab15a27cce99a17a33883804d8/src/hdmf/build/manager.py#L193-L194

Not sure if it should be generalized, by just adding `__hash__` to Builder. It seems simple enough that its not a big deal to leave as is, but thought I would mention it.

ajtritt (approved):

src/hdmf/backends/hdf5/h5tools.py
rly: Yeah, that's where I got it from. I figured there might have been a reason that it was on the `BuildManager` rather than on `Builder` in the first place, e.g., maybe builders should by default have a hash that is based on its contents, so I left it and re-implemented it here.

rly:
Note that since this technically includes a breaking change, this PR is being merged to a branch for HDMF 2.0, not dev.

PR #383, opened 2020-06-16
Add ability to close open links

rly:
Current behavior:
If file A links to dataset D in file B and file A is closed, then file B is still open. H5Dataset D can still be accessed. On Mac/Linux, file B can be deleted or overwritten, but on Windows, trying to delete or overwrite file B results in an error. 

This PR adds a new function `HDF5IO.close_linked_files` which will close all linked-to files that were opened by this `HDF5IO` instance.

Practically speaking, this is useful only for Windows users to get around the inconsistent behavior of h5py and in CI on Windows.

This also fixes an issue where links are not added to unnamed groups (e.g., `NWBFile.acquisition`).

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -564,7 +564,8 @@ class HDF5IO(HDMFIO):
         """Close all opened, linked-to files.
 
         MacOS and Linux automatically releases the linked-to file after the linking file is closed, but Windows does
-        not, which prevents the linked-to file from being deleted or truncated.
+        not, which prevents the linked-to file from being deleted or truncated. Use this method to close all opened,
+        linked-to files.
         """
         for obj in self.__open_links:
             if obj:
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -556,12 +556,18 @@ class HDF5IO(HDMFIO):
                 kwargs = {}
             self.__file = File(self.__path, open_flag, **kwargs)
 
-    def close(self, close_links=False):
+    def close(self, close_links=True):
+        """Close the file.
+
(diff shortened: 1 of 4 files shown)


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -553,9 +556,13 @@ def open(self):
                 kwargs = {}
             self.__file = File(self.__path, open_flag, **kwargs)
 
-    def close(self):
+    def close(self, close_links=False):
```
oruebel: I would add a docstring for ``close_links``. I'm not sure whether the default should be ``close_links=True`` or ``close_links=False``. Setting the default as ``false`` may be safer, because it matched the current behavior, but in practice, the more common case may be to use ``true`` as it ensures that no files remain silently open.

rly: Agreed. Default should be `close_links=True` (this was just a test).

rly:
It looks like normally, h5py keeps the *linked* file open after the *linking* file is closed on all OS's. But on Mac/Linux, the *linked* file can be deleted or truncated, but on Windows, trying to delete or truncate the *linked* file results in an error.
```
  File "C:\Users\Ryan\Documents\NWB\hdmf\tests\unit\test_io_hdf5_h5tools.py", line 996, in tearDown
    os.remove(self.path1)  # linked file may not be closed
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\Ryan\\AppData\\Local\\Temp\\tmpabnbjx67'
```

See also https://github.com/NeurodataWithoutBorders/pynwb/issues/975

This seems like a bug in h5py or just a Windows filesystem quirk, so I filed an issue ticket there: https://github.com/h5py/h5py/issues/1569

rly:
@oruebel I changed the PR to create a new method rather than add to `HDF5IO.close`.

oruebel (approved):

ajtritt (approved):

PR #388, opened 2020-06-18
Add export functionality

### LINKED ISSUE #315: unable to read -> (modify) -> write due to container_source

rly:
## Motivation

Supersedes #326. 

Fix #315 and NeurodataWithoutBorders/pynwb#668

**This is a work in progress.**
- [x] Fix datasets linking when they should not. Currently datasets in the exported file are links to the original file
- [x] Fix failure to append or pop data between read and export. Builder needs to be marked as modified


export_faq.md
```diff
@@ -0,0 +1,55 @@
+## Export FAQ
+
+### What happens if I export an HDF5 file to a new HDF5 file with soft or external links?
```
oruebel: For development this is fine, but for user-docs I would suggest a slightly different approach for this  FAQ. Specificaly I would suggest to: 

1. Split the question into "internal links (i.e., soft links)" and "links to other files (i.e., external links)". I believe the answer for SoftLinks is imply that they will be preserved in all cases. It's only for external links where things get a tricky.
2. For external links, rather than breaking things up right away into "Situations", I would suggest to divide this into two main cases: 1) preserve external links and 2) resolve external links (i.e., copy the data) and then list the different cases there.
3. Instead of just numbering the different cases it would be nice to give them names
3.

src/hdmf/backends/io.py
```diff
@@ -36,23 +36,70 @@ def read(self, **kwargs):
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
-            {'name': 'exhaust_dci', 'type': bool,
-             'doc': 'exhaust DataChunkIterators one at a time. If False, exhaust them concurrently', 'default': True})
+            {'name': 'write_config', 'type': 'WriteConfig', 'doc': 'configuration settings for writing builders',
+             'default': None})
```
rly: maintain API. create `WriteConfig`

src/hdmf/backends/io.py
```diff
@@ -36,23 +36,70 @@ def read(self, **kwargs):
...
         self.write_builder(f_builder, **kwargs)
 
+    @docval({'name': 'src_io', 'type': 'HDMFIO', 'doc': 'the HDMFIO object for reading the data to export'},
+            {'name': 'container', 'type': Container,
+             'doc': ('the Container object to export. If None, then the entire contents of the HDMFIO object will be '
+                     'exported'),
+             'default': None},
+            {'name': 'read_args', 'type': dict, 'doc': 'dict of arguments to use when calling read_io.read_builder',
+             'default': dict()},
+            {'name': 'write_config', 'type': 'WriteConfig', 'doc': 'configuration settings for writing builders',
```
rly: 'write_config' -> 'write_args'. Make it a dict. `WriteConfig` is used only internally

src/hdmf/backends/io.py
```diff
@@ -36,23 +36,70 @@ def read(self, **kwargs):
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
-            {'name': 'exhaust_dci', 'type': bool,
-             'doc': 'exhaust DataChunkIterators one at a time. If False, exhaust them concurrently', 'default': True})
```
rly, line 40: remove exhaust_dci from HDMFIO

src/hdmf/build/manager.py
```diff
@@ -193,6 +196,27 @@ def __conthash__(self, obj):
     def __bldrhash__(self, obj):
         return id(obj)
 
+    def purge_outdated(self):
+        containers_copy = self.__containers.copy()
+        for container in containers_copy.values():
+            if container.modified:
+                container_id = self.__conthash__(container)
```
rly, line 204: use the convenience function created below

rly:
This is ready for review. Tests pass in PyNWB after adding roundtrip tests with additional export to new file.

I will work on improving the documentation and checking the edge cases (might need to test scalar dataset that is a reference since that is handled differently). Region references are probably not supported properly. 

Before I go too deep, it would be great to get your input @oruebel @ajtritt since this touches a lot of code.

Tests fail on Python 3.5 because of dictionary ordering. What do you think about dropping Python 3.5 support in HDMF 2 (which is where this PR will go)? End of life for Python 3.5 is Sep. 13, 2020 and a number of popular packages already do not release py3.5 wheels.

oruebel:
> Tests fail on Python 3.5 because of dictionary ordering.

Could you clarify the problem. Does this mean HDMF relies on ordering of keys in dicts?

rly:
> > Tests fail on Python 3.5 because of dictionary ordering.
> 
> Could you clarify the problem. Does this mean HDMF relies on ordering of keys in dicts?

`Builder` objects use the native Python `dict`. If a `GroupBuilder` has multiple groups, then iterating over the `GroupBuilder` groups has a non-deterministic order in Python 3.5. For example, if `FooFile` contains multiple `Foo` objects, the list of `Foo` objects in `FooFile` may be ordered differently on subsequent reads from a file. 

We could replace the use of dict with OrderedDict, we could ignore the issue and work around it in tests, or we could drop Python 3.5 support.

oruebel:
> We could replace the use of dict with OrderedDict, we could ignore the issue and work around it in tests, or we could drop Python 3.5 support.

Ordering of objects in dict is not something we should rely on, as it is not guaranteed and the behavior may be different between implementations of pythons and OS (it seems like Python 3.5 is particular is just particularly vulnerable to this). If we rely on objects being sorted in the order they are added to the dict, then using OrderedDict seems like the correct solution. I don't think we rely on this, but if HDMF does, then we should switch to OrderedDict. If the ordering is not important and only an (in)convenience for tests, then working around it in the tests seems like the right solutions. I'm not opposed to dropping Python 3.5 support, but it doesn't seem like the right solution for this problem.

rly:
> Ordering of objects in dict is not something we should rely on, as it is not guaranteed and the behavior may be different between implementations of pythons and OS (it seems like Python 3.5 is particular is just particularly vulnerable to this). If we rely on objects being sorted in the order they are added to the dict, then using OrderedDict seems like the correct solution. I don't think we rely on this, but if HDMF does, then we should switch to OrderedDict. If the ordering is not important and only an (in)convenience for tests, then working around it in the tests seems like the right solutions. I'm not opposed to dropping Python 3.5 support, but it doesn't seem like the right solution for this problem.

Yeah, I agree. HDMF does not inherently rely on objects being sorted in the order they are added to the dict, though you can easily and naively create a Container where the order matters, so from a user perspective, I think the order should at least be deterministic, whether that be order of insertion, alphanumeric, or something else.

Note that PyNWB allows access to a list of groups only through a dict with the key as the container's name (this is what PyNWB does with `MultiContainerInterface`, so e.g., `nwbfile.acquisition[0]` is not allowed).

export_faq.md
rly: I updated this in `docs/source/export.rst`

rly:
I addressed the issue of dictionary ordering of objects in the tests and I updated the export documentation. Everything looks good on my end. Please review. @ajtritt @oruebel

docs/source/export.rst
```diff
@@ -0,0 +1,65 @@
...
+- If the modifications are removals or additions of containers, then no special action must be taken, as long as the
+  container hierarchy is updated correspondingly.
+- If the modifications are changes to attributes, then
+  :py:meth:`Container.set_modified() <hdmf.container.AbstractContainer.set_modified>` must be called
+  on the container before exporting.
+
+.. note::
+
+  Modifications to :py:class:`h5py.Dataset <h5py.Dataset>` objects act *directly* on the read file on disk.
+  Changes are applied immediately and do not require exporting or writing the file.
```
oruebel:
```suggestion
  Changes are applied immediately and do not require exporting or writing the file. If you want to modify a dataset only in the new file, than you should replace the whole object with a new array holding the modified data. To prevent unintentional changes to the source file, the source file should be opened with ``mode='r'``.
```

docs/source/export.rst
```diff
@@ -0,0 +1,65 @@
...
+``link_data=False`` to :py:meth:`HDF5IO.write <hdmf.backends.hdf5.h5tools.HDF5IO.write>` when writing a file with a
+copy of externally linked datasets.
+
+What happens to references when I export?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+References will be preserved in the exported file.
+NOTE: Exporting a file involves loading into memory all datasets that contain references and attributes that are
+references. The HDF5 reference IDs within an exported file may differ from the reference IDs in the original file.
+
+Can I write a newly instantiated container to two different files?
```
oruebel: This section should be moved up so that the ``Can I do X`` sections appear together and the ``What happens to Y`` section appear together. It may also be useful to divide these into two sections with these as subsections.

docs/source/export.rst
```diff
@@ -0,0 +1,65 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+HDMF does not allow you to write a container that was not read from a file to two different files. For example, if you
+instantiate container A and write it file 1 and then try to write it to file 2, an error will be raised. However, you
+can read container A from file 1 and then export it to file 2, with or without modifications to container A in
+memory.
```
oruebel: I would add a section on ``What happens to object_ids?`` I assume we create new object_ids on export, but it will be useful to document the behavior.

oruebel (approved):
Congratulations, that was quite a bit of work. Looks good to me. I added a couple of nit-picky comments on the documentation, but nothing critical.

docs/source/export.rst
rly: Object IDs are actually kept the same. A Container read from a file has a particular object ID. If the read Container is exported to a new file, the exported Container maintains its original object ID. As implemented currently, the object ID is only unique within the file. We can change that behavior, but that is how it stands currently.

It would be useful to have an ID that is unique on every write. In NWB, I suggest we use the `identifier` field for this.

oruebel: Thanks for the clarification. It would be good to at least document this behavior. Ideally, I think this should be a parameter on the export function to ``keep_object_ids=True``. I think it is fine to address this as a separate issue (rather than delaying the merge of this PR) if adding the option to generate new object_ids is tricky.

docs/source/export.rst
```diff
@@ -0,0 +1,71 @@
...
+What happens to references when I export?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+References will be preserved in the exported file.
+NOTE: Exporting a file involves loading into memory all datasets that contain references and attributes that are
+references. The HDF5 reference IDs within an exported file may differ from the reference IDs in the original file.
+
+What happens to object IDs when I export?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+After exporting a container, the object IDs of the container and its child containers will be identical to the object
+IDs of the read container and its child containers. The object ID of a container uniquely identifies the container
```
ajtritt, line 70: Are we sure we want this to be the case?

ajtritt (commented):
Everything looks good. But, I hadn't thought about the Object ID part of this. Has this been discussed or have the implications of this been full hashed out? Does DANDI rely on OIDs?

docs/source/export.rst
bendichter: I think probably not. Would it be hard to generate new IDs? Ideally I think there would be a flag for this, but I'd expect most use-cases to want to generate new IDs. If data from a dataset changes, then you'd have two different datasets with the same id, which seems like it might break integrations that rely on them

rly: I am not opposed to generating new IDs, but what is the use case for that? 

We have been recommending that users and developers NOT use object IDs as globally unique identifiers (because of the dataset issue that you mentioned and also, if you add a new child container to a parent and write the modified parent to a file in append mode, the modified parent object ID does not change). Globally unique identifiers should be managed by a data archive where changes to a data file can be controlled.

As I see it, the intended use for object IDs is to uniquely identify containers within a file -- it is effectively a backend-agnostic path. So it does not matter that the object IDs are the same as in another file.

Maybe I am missing a use case?

If not, then I am wary of generating new IDs on export because we would be adding support for a use case that doesn't exist and we might suggest to users/developers that the IDs can be used to uniquely identify a container like a hash.

ajtritt (approved):

docs/source/export.rst
ajtritt: > I am not opposed to generating new IDs, but what is the use case for that?

I agree with @rly. Regenerating OIDs doesn't add any value, and if we can't articulate a concrete use case, it's not worth the extra code.

rly: @oruebel, @bendichter, and I discussed this and decided that it would probably be good to at least have a flag for generating new IDs. Since users are changing some aspect of the data on export, they will expect that the object ID is updated. The question is whether this should be default True or False. For now, I will merge this PR, and we can modify the behavior before the public release or at a later date.

PR #390, opened 2020-06-22
Update

ajtritt:
## Motivation

The original motivation for this PR was to add the ability to write table columns from DataChunkIterators. This work went stale and when it was brought up to date, some other issues arose. In response to those issues, a couple more features were added.

1. Add ability to write `DynamicTable` columns from a `DataChunkIterator`
2. Add `DynamicTable.get` method. This method behaves like `__getitem__`/`[]`, but can take optional arguments to modify the behavior. Currently, the only argument that exists is `df`, which will control if a selection is returned as a `pandas.DataFrame`. 
    - The work of selecting from a DynamicTable is carried out by `DynamicTable.get`. `DynamicTable.__getitem__` calls `DynamicTable.get` with whatever options are necessary to provide familiar functionality. Moving forward, this pattern with `get`/`__getitem__` will be used to provide more flexibility for selecting data from a `DynamicTable`. See `VocabData` for another example.
3. `DynamicTable.__getitem__` functionality wasn't fully tested and had some issues. One major issue was that returning a `pd.DataFrame` did not work if one of the columns was multi-dimensional or ragged. This PR will resolve those issues.
4. Add `VocabData` column type for storing controlled-vocabulary data.

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -7,7 +7,7 @@ import logging
 import warnings
 
 from ...container import Container
-from ...utils import docval, getargs, popargs, call_docval_func, get_docval, get_data_shape, fmt_docval_args
+from ...utils import docval, getargs, popargs, call_docval_func, get_data_shape, fmt_docval_args, get_docval
 from ...data_utils import AbstractDataChunkIterator
 from ...build import Builder, GroupBuilder, DatasetBuilder, LinkBuilder, BuildManager,\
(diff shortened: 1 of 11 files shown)


oruebel:
Thanks for reopenning the PR. As far as I can  tell there seems to be only one test failing:

```
ERROR in test test_get_not_found_default (common.test_table.TestDynamicTable):
 
    File "/root/project/tests/unit/common/test_table.py", line 129, in test_get_not_found_default
 
      self.assertEqual(table.get('qux', 1), 1)
 
  AssertionError: None != 1
 ```

oruebel (approved):

PR #396, opened 2020-07-09
Work around codecov.io connection failures

rly:
## Motivation

codecov.io has known random connection issues such that around 10% of the time, coverage reports from CI services fail to be uploaded to codecov.io. See:
https://github.com/codecov/codecov-python/issues/158
https://community.codecov.io/t/unreliable-coverage-report-uploads/322
https://community.codecov.io/t/github-pr-not-being-updated-with-final-coverage-data/1425/15
https://github.com/h5py/h5py/issues/1398

The most likely issue is that the codecov server is randomly blocking requests from the CI machine. This could be due to rate limiting or a long-term ban on the IP address. When testing this PR, re-running the codecov script on failure did not seem to resolve the issue: (`Error: HTTPSConnectionPool(host='codecov.io', port=443): Max retries exceeded with url: /upload/v4<...>(Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x000001FA35482DC0>: Failed to establish a new connection: [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond')`)
`) 

The error may also result from proxy issues or network issues, though this seems less likely given the repeated failures and other people's experiences. **So, this might not be solvable on our end.** The only resolution might be to re-run the CI if uploading to codecov fails. 

This PR tries re-running the codecov script on failure, after a delay to get around rate limits and intermittent network failures, but really, I think the problem needs to be addressed by codecov.

## Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -117,7 +117,7 @@ jobs:
       fi
     displayName: 'Run coverage tests if coverageToxEnv != ""'
     env: |
-      CODECOV_TOKEN: $(CODECOV_TOKEN)
+      CODECOV_TOKEN: $(MY_CODECOV_TOKEN)
 
   - bash: |
       tox -e $(buildToxEnv)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -116,6 +116,8 @@ jobs:
         tox -e $(coverageToxEnv)
       fi
     displayName: 'Run coverage tests if coverageToxEnv != ""'
+    env: |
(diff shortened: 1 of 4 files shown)


oruebel (dismissed):

rly:
Analysis of the failed CI runs shows that the attempted fix here does not resolve the problem. It is likely that the codecov server has blocked some CI servers from pinging it, so even repeated attempts after a delay do not resolve the issue and simply slow down CI. So I am closing this PR unless the issue returns and can be resolved by these changes.

PR #400, opened 2020-07-14
Create reserved attributes in builders

### LINKED ISSUE #222: Make backend handle reserved attributes
A request from @oruebel:

Currently, there are three reserved attributes on the backend for HDMF: 'namespace', 'data_type', and 'object_id'. These are stored as attributes on builders and then the HDF5 backend writes them out as HDF5 attributes. Instead, the builders should store these reserved attributes under specific fields and let the backend decide how the attributes should be written. This is relevant for the implementation of dimension scales where 'dimensions' and 'coordinates' would be added as new reserved attributes.

Practically speaking, this change _should_ just be a refactoring and addition of fields, getters, and setters on the builder. 

What do you think @ajtritt?

rly:
## Motivation

I made `BaseBuilder` aware of the three reserved values: 'namespace', 'data_type', and 'object_id', and added functionality in `HDF5IO` that writes these values separately from `set_attribute`. 

See #222. Fix #222 

Internal changes:
- `BaseBuilder` previously had keys 'namespace', 'data_type', 'object_id' in the `attributes` dictionary. These would be written as attributes in `HDF5IO`, and `HDF5IO` would have no knowledge that these attributes were special. These keys are now stored in a new dict `BaseBuilder.reserved`. 

Potentially breaking changes:
- `HDF5IO` will raise an error when trying to write a builder attribute with a reserved name ('namespace', 'data_type', or 'object_id'). Note: if the loaded namespace has a different data type key or object ID key (e.g., 'neurodata_type'), then that is used as the reserved name.
- Both 'namespace' and 'data_type' are now required for all typed groups and datasets. You cannot set one without the other. Previously you could have one without the other, but I am not sure why you would do that. 'object_id' is optional because older files may not have object ids.
- Previously it was possible for the `GroupSpec.type_key` to be different from `DatasetSpec.type_key` (e.g., "data_type" or "neurodata_type"). This should not happen and now raises an error. Same for `GroupSpec.id_key` and `DatasetSpec.id_key`.

## Checklist

- [x] Is your contribution compliant with our coding style? This can be checked running `flake8` from the source dirctory.


rly:
Now that the issue with changing object IDs on export has been resolved and after further discussion with @ajtritt, I will close and put this PR on hold until there is a need, e.g., if a backend wants to store reserved values differently than as attributes.

@ajtritt brings up an issue with the current PR: the IO layer should not need to know about the specifics of the data model. I am not totally convinced by this; if a backend wants to store the reserved values not as attributes but as, say, datasets or in a separate database table, it should be free to, but we do not yet have that use case with HDF5 or Zarr. If the need arises, we can reopen this PR.

PR #405, opened 2020-07-21
Enhance LabelledDict with add callable, remove callable, and restricted methods

### LINKED ISSUE #404: Removing items from a LabelledDict
PyNWB uses `LabelledDict` in `MultiContainerInterface`. A `MultiContainerInterface` type can contain any number of `LabelledDict`. For example, `NWBFile` contains one for the acquisition group, the analysis group, the processing group, and so on.

With the recent addition of export to HDMF, users wish to remove items from a group, e.g., remove a `TimeSeries` from the `acquisition` group. What should the interface be for this?

Option 1:
I think the most intuitive and pythonic ways to do this are to call `del nwbfile.acquisition['my_timeseries']` or `nwbfile.acquisition.pop('my_timeseries')`. Currently, this will not remove the `TimeSeries` properly because the parent, `NWBFile`, needs to know that that child has been removed, but we can modify the `__delitem__` and `pop` methods of `LabeledDict` to call `parent._remove_child(child)`.

Option 2:
`MultiContainerInterface` creates methods for adding an item to a group (`add_acquisition`), getting an item (`get_acquisition`), and accessing the attribute (`.acquisition`). We could add a method for removing an item (`remove_acquisition`) which removes the item from the dict and calls `parent.remove_child(child)`. This is more consistent with the current API. Consistency is important, but I am personally not a fan of this method because it would add at least 13 methods to the already bloated [NWBFile API](https://pynwb.readthedocs.io/en/stable/pynwb.file.html) which has 50+ methods. 

Note: I understand that the `create_*` methods are based off of h5py. h5py does not have `remove_*` methods but favors the `del group['name']` syntax.

Which method would be better?

rly:
## Motivation

This is technically a breaking change on `LabelledDict`, but `LabelledDict` is not used within HDMF or by PyNWB. 

Fix #404, implementing option 1

## How to test the behavior?
```
# removing. these will now call a function that is passed in to __init__ that calls parent._remove_child(child)
del nwbfile.acquisition['my_time_series']
nwbfile.acquisition.pop('my_time_series')
nwbfile.acquisition.popitem()
nwbfile.acquisition.clear()

# adding. these will now call a function that is passed in to __init__ that calls child.parent = parent
nwbfile.acquisition['new_key'] = xx
nwbfile.acquisition.add(xx)
```

## Checklist

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -4,6 +4,7 @@ import collections
 import h5py
 import numpy as np
 import warnings
+import types
 from enum import Enum
 
 __macros = {
@@ -790,7 +791,7 @@ def pystr(s):
 
 
 class LabelledDict(dict):
-    """A dict wrapper class with a label and which allows retrieval of values based on an attribute of the values
+    """A dict wrapper that allows querying by an attribute of the values and running a callable on removed items.
 
     For example, if the key attribute is set as 'name' in __init__, then all objects added to the LabelledDict must have
     a 'name' attribute and a particular object in the LabelledDict can be accessed using the syntax ['object_name'] if
@@ -803,8 +804,15 @@ class LabelledDict(dict):
     condition, a KeyError is raised. Note that if 'attr' equals the key attribute, then the single matching value is
     returned, not a set.
 
+    LabelledDict does not support changing items that have already been set. A TypeError will be raised when using
+    __setitem__ on keys that already exist in the dict. The __delitem__, setdefault, and update methods are not
+    supported. A TypeError will be raised when these are called.
+
+    A callable function may be passed to the constructor to be run on the result of pop and popitem, and each removed
+    item from clear.
+
     Usage:
-      LabelledDict(label='my_objects', def_key_name = 'name')
+      LabelledDict(label='my_objects', key_attr='name')
       my_dict[obj.name] = obj
       my_dict.add(obj)  # simpler syntax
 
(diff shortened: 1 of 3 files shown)


src/hdmf/utils.py
```diff
@@ -887,5 +914,47 @@ def add(self, value):
...
+        if self.__remove_callable:
+            self.__remove_callable(ret)
+        return ret
+
+    def popitem(self):
+        """Remove the last added item. If remove_callable was initialized, call that on the returned value.
+
+        Note: popitem returns a tuple (key, value) but the remove_callable will be called only on the value.
+
+        Note: in Python 3.5, dictionaries and not ordered, so popitem removes an arbitrary item.
```
oruebel:
```suggestion
        Note: in Python 3.5 and earlier, dictionaries are not ordered, so popitem removes an arbitrary item.
```

oruebel (dismissed):

bendichter:
@rly What was the motivation for the addition of the add and remove callables? I don't have an issue with them being there, and they appear to be well tested, but I must have missed what they are there for.

oruebel:
@bendichter this is related the Export use case where users want to read a file, modify it, and export to a new file. In this case modifying a file means, e.g., removing a TimeSeries from NWBFile.acquisition.

bendichter:
@oruebel yes I understand why we need a remove method, but not why we also need a customizable remove callback

oruebel:
The callback is needed because in addition the LabeledDict itself, the owner of the LabeledDict may need to perform additional actions when the container is being removed. In the case of PyNWB e.g,. the TimeSeries also needs to be removed the list of children in the parent Container (e.g., NWBFile) when it is being removed from the LabeledDict.

rly:
Yes, exactly what @oruebel said. The parent container needs to 1) remove the removed container from its list of children and 2) signal that it has been modified so it gets rebuilt on write/export without the removed child. Only modified containers are rebuilt.

There are other ways to resolve the issue that don't involve a callback, but I thought it would be nice to make it flexible, in case we want to use `LabelledDict` for other purposes. It seems generally useful. Note: Very few users, if any, should ever need to instantiate `LabelledDict`.

Re: the add callback. This allows users to better use the LabelledDict like a dict, which they may do given that LabelledDict is a dict. For example, in PyNWB, you can access a behavior `ProcessingModule` via `nwbfile.processing['behavior']`. This exposes the `nwbfile.processing` LabelledDict. Users may accidentally try to add a behavior `ProcessingModule` using `nwbfile.processing['behavior'] = my_module`. This does not raise an error, but the parent of the new module would not be set and so the new module would not be written correctly to file. We should either prevent users from setting `LabelledDict` items this way or support it properly by setting the new item's parent (this is done by `nwbfile.create_processing_module` and `nwbfile.add_processing_module` but not when operating directly on `nwbfile.processing`)

ajtritt (approved):

PR #422, opened 2020-08-26
Improve load namespaces warning

rly:
## Motivation

When `load_namespaces=True` is passed to `NWBHDF5IO`, these warnings are always printed:
```
UserWarning: ignoring namespace 'hdmf-common' because it already exists
UserWarning: ignoring namespace 'core' because it already exists
```
This is 1) confusing to the user (should I be concerned about this warning??) and 2) not that informative.

This PR changes the warning so that it is printed ONLY if the cached namespace version differs from the current namespace version (this assumes that different namespaces of the same name have different versions). And it clarifies the warning to be like:
```
UserWarning: Ignoring cached namespace 'core' version 2.1.0 because version 2.2.5 is already loaded.
```

This PR also adds the utility function `remove_test_file` in `hdmf.testing` for use in tests.

## Checklist


ajtritt (dismissed):

rly:
Tests fail in PyNWB because the warning no longer occurs / has changed in some cases. This is acceptable to me. I will update the PyNWB tests accordingly.

oruebel (approved):

ajtritt (approved):

PR #426, opened 2020-09-03
add capability to add a row to a column after IO

bendichter:
## Motivation

Tables that have columns that use DataIO should now be able to be built using the standard `.add_row` syntax.

tested:
* 1-D `list`
* `DataIO` is used for an n-d `np.ndarray`
* `DataIO` is used for a column index
* chunked `h5py.Dataset` (after read)

I moved the addend and extend logic to data_utils because I needed to use the same logic there.

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -3,8 +3,43 @@ from collections.abc import Iterable
 import numpy as np
 from warnings import warn
 import copy
+import h5py
 
 from .utils import docval, getargs, popargs, docval_macro, get_data_shape
+
+
+def append_data(data, arg):
+    if isinstance(data, (list, DataIO)):
+        data.append(arg)
+        return data
+    elif isinstance(data, np.ndarray):
+        return np.append(data, [arg])
+    elif isinstance(data, h5py.Dataset):
+        shape = list(data.shape)
+        shape[0] += 1
+        data.resize(shape)
+        data[-1] = arg
+        return data
+    else:
+        msg = "Data cannot append to object of type '%s'" % type(data)
+        raise ValueError(msg)
+
+
+def extend_data(data, arg):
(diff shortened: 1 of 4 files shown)


bendichter:
@rly I am having trouble with using `add_row` on an indexed ndarray. See most recent commit. Could you help out with this?

edit: I figured it out

src/hdmf/data_utils.py
```diff
@@ -3,10 +3,45 @@
...
 from .utils import docval, getargs, popargs, docval_macro, get_data_shape
 
 
+def append_data(data, arg):
+    if isinstance(data, (list, DataIO)):
+        data.append(arg)
+        return data
+    elif isinstance(data, np.ndarray):
+        return np.append(data, [arg])
+    elif isinstance(data, h5py.Dataset):
```
ajtritt, line 17: When would this happen? DataIO is used on write. The only time you use an `h5py.Dataset` on write is if you are making a copy to another file, in which case, you shouldn't be altering it.

bendichter: The motivation for this is to be able to add rows to dynamic tables post-write. As the test is showing on line 1314, this would involve reading a dataset in in append mode and altering the datasets by adding rows to the data, the table ids, and any index columns

ajtritt: But at that point isn’t add_row just operating on the VectorData object?

bendichter: Yes. VectorData inherits from Data and Data calls this function. This is invoked when appending to a column of a file that was read in append mode. See https://github.com/hdmf-dev/hdmf/blob/2544eb9528a745a89210d3644aa77b171a8779ec/tests/unit/common/test_table.py#L1314-L1325

ajtritt (approved):

PR #430, opened 2020-10-01
add multi-index code and test

### LINKED ISSUE #1296: specify in the API class the a column is multi-indexed
Now that we have double indexes working

https://github.com/hdmf-dev/hdmf/blob/535985ca4eafed6b5bdcd34ab3bc32000ee54a88/tests/unit/common/test_table.py#L1067-L1078

we can use this for specifying waveforms. But how do we specify this in the python API as an optional column? in `__columns__`, we can add `{'name': 'waveforms', 'description': 'snippet for each spike', 'index': True}`. That specifies a single index, but how to specify a double index? Maybe we can have `index` take an `int`, specifying the number of dimensions that are indexed?

https://github.com/NeurodataWithoutBorders/pynwb/blob/c4631e9e8e09a46e427442fd173818fb40414283/src/pynwb/misc.py#L143-L152

bendichter:
## Motivation

fix https://github.com/NeurodataWithoutBorders/pynwb/issues/1296

Allows for the index arg of add_column to take an `int` type, in which case the column will be indexed `int` times

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -85,7 +85,11 @@ class VectorIndex(VectorData):
         Add the given data value to the target VectorData and append the corresponding index to this VectorIndex
         :param arg: The data value to be added to self.target
         """
-        self.target.extend(arg)
+        if isinstance(self.target, VectorIndex):
+            for a in arg:
+                self.target.add_vector(a)
+        else:
+            self.target.extend(arg)
         self.append(self.__check_precision(len(self.target)))
 
     def __check_precision(self, idx):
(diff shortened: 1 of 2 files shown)


bendichter:
to do: 

* add description to docval about int
* add test: use __columns__

src/hdmf/common/table.py
```diff
@@ -557,7 +561,7 @@ def __eq__(self, other):
              'doc': 'a dataset where the first dimension is a concatenation of multiple vectors', 'default': list()},
             {'name': 'table', 'type': (bool, 'DynamicTable'),
              'doc': 'whether or not this is a table region or the table the region applies to', 'default': False},
-            {'name': 'index', 'type': (bool, VectorIndex, 'array_data'),
+            {'name': 'index', 'type': (bool, VectorIndex, 'array_data', int),
```
rly, line 564: Document 'int' dtype (and others)

tests/unit/common/test_table.py
```diff
@@ -205,6 +205,20 @@ def test_add_column_vectorindex(self):
         with self.assertWarnsWith(FutureWarning, msg):
             table.add_column(name='bad', description='bad column', index=ind)
 
+    def test_add_column_multi_index(self):
```
rly, line 209: Add test with `DynamicTable` subclass with `__columns__` where index=int

ajtritt (changes requested):
Add more documentation to add_column. Eventually lets refactor this function to make it easier to maintain.

rly (approved):
Looks good to me

ajtritt (approved):

PR #441, opened 2020-10-17
Add method AbstractContainer.get_fields_conf, fix fields handling

### LINKED ISSUE #440: keep child info in __fields__
see https://github.com/NeurodataWithoutBorders/pynwb/issues/1304

We'd like to be able to determine if an argument is destined to be a link or not, but the 'child' fields in `cls.__field__` are currently overwritten, so there is no way to determine this from the API

rly:
Fix #440.

- Add new method `AbstractContainer.get_fields_conf` 
- Fix handling of `__fields__` for `Data` subclasses. `__fields__` should contain the field names of the class' parents, but this was only happening for `Container` subclasses. (Note: this used to happen for both `NWBContainer` and `NWBData` subclasses but got lost in the move to HDMF)
- Add many missing tests for specifying `__fields__`
- Refactor check on valid keys in `__fields__`
- Add requirement that fields conf cannot contain a field conf of a parent class
- Fix `DynamicTableRegion` having duplicate fields conf 'table'

## Checklist


oruebel (dismissed):

rly:
Thanks for the review @oruebel. I made a minor change in `manager.py` to address the failing pynwb test -- `'child': True` was incorrectly being added to `__nwbfields__` for a dynamically generated subclass of `Data`. Please re-review.

oruebel (approved):

PR #442, opened 2020-10-21
Add resource reference tables

### LINKED ISSUE #156: Move NWBTable functionality to generic Table class in hdmf.common
The `NWBTable` and `NWBTableRegion` classes in PyNWB are general-purpose classes to handle 1D datasets with a compound type, which has table-like behavior. They do not reference anything NWB specific, are not in the core NWB schema, and are not used by anything in NWB, as far as I can see. The code is standalone, general-purpose code that may eventually be useful. As such, I think it would be more appropriate for these types to exist in hdmf.common rather than in PyNWB.

ajtritt:
## Motivation

Add the ability to track the use of external resource references across a file.

Fix #156 

## How to test the behavior?
```python
import pandas as pd
from hdmf.common import ExternalResources

er = ExternalResources('terms')
er.add_ref('01234567-89ab-cdef-0123-456789abcdef', 'species', 'mouse', 'NCBI Taxonomy', '10090',
           'https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10090')
er.add_ref('fedcba98-7654-3210-fedc-ba9876543210', 'gene', 'Cacna1s', 'Mouse Genome Informatics', 'MGI:88294',
           'http://www.informatics.jax.org/marker/MGI:88294')


# Alternatively, keys can be added in bulk

df = pd.DataFrame(
    data = [['mouse', 'NCBI Taxonomy', '10090', 'https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10090'],
            ['Cacna1s', 'Mouse Genome Informatics', 'MGI:88294', 'http://www.informatics.jax.org/marker/MGI:88294']],
    columns = ['key_name', 'resource_name', 'resource_id', 'resource_uri']
)

keys = er.add_keys(df)
er.add_ref('01234567-89ab-cdef-0123-456789abcdef', 'species', keys['mouse'])
er.add_ref('fedcba98-7654-3210-fedc-ba9876543210', 'gene', keys['Cacna1s'])
```

## Checklist

### DIFF
--- a/src/hdmf/common/io/resources.py
+++ b/src/hdmf/common/io/resources.py
@@ -7,6 +7,10 @@ from .. import register_map
 class ExternalResourcesMap(ObjectMapper):
 
     def construct_helper(self, name, parent_builder, table_cls, manager):
+    	"""Create a new instance of table_cls with data from parent_builder[name].
+
+    	   The DatasetBuilder for name is associated with data_type Data and container class Data,
+    	   but users should use the more specific table_cls for these datasets.```
         parent = manager._get_proxy_builder(parent_builder)
         builder = parent_builder[name]
         src = builder.source
--- a/src/hdmf/common/io/resources.py
+++ b/src/hdmf/common/io/resources.py
@@ -1,0 +1,28 @@
+from ...build import ObjectMapper
+from ..resources import ExternalResources, KeyTable, ResourceTable, ObjectTable, ObjectKeyTable
+from .. import register_map
+
+
+@register_map(ExternalResources)
+class ExternalResourcesMap(ObjectMapper):
+
+    @classmethod
+    def construct_helper(cls, name, builder, table_cls):
+        builder = builder[name]
+        return table_cls(name=name, data=builder.data)
+
+    @ObjectMapper.constructor_arg('keys')
+    def keys(self, builder, manager):
+        return self.construct_helper('keys', builder, KeyTable)
+
+    @ObjectMapper.constructor_arg('resources')
+    def resources(self, builder, manager):
+        return self.construct_helper('resources', builder, ResourceTable)
+
+    @ObjectMapper.constructor_arg('objects')
+    def objects(self, builder, manager):
+        return self.construct_helper('objects', builder, ObjectTable)
+
+    @ObjectMapper.constructor_arg('object_keys')
+    def object_keys(self, builder, manager):
+        return self.construct_helper('object_keys', builder, ObjectKeyTable)
--- a/src/hdmf/common/io/resources.py
+++ b/src/hdmf/common/io/resources.py
@@ -6,23 +6,26 @@ from .. import register_map
 @register_map(ExternalResources)
 class ExternalResourcesMap(ObjectMapper):
 
-    @classmethod
-    def construct_helper(cls, name, builder, table_cls):
-        builder = builder[name]
-        return table_cls(name=name, data=builder.data)
+    def construct_helper(self, name, parent_builder, table_cls, manager):
+        parent = manager._get_proxy_builder(parent_builder)
+        builder = parent_builder[name]
+        src = builder.source
+        oid = builder.attributes.get(self.spec.id_key())
+        kwargs = dict(name=builder.name, data=builder.data)
+        return self.__new_container__(table_cls, src, parent, oid, **kwargs)
 
     @ObjectMapper.constructor_arg('keys')
     def keys(self, builder, manager):
-        return self.construct_helper('keys', builder, KeyTable)
+        return self.construct_helper('keys', builder, KeyTable, manager)
 
     @ObjectMapper.constructor_arg('resources')
     def resources(self, builder, manager):
-        return self.construct_helper('resources', builder, ResourceTable)
+        return self.construct_helper('resources', builder, ResourceTable, manager)
 
     @ObjectMapper.constructor_arg('objects')
     def objects(self, builder, manager):
-        return self.construct_helper('objects', builder, ObjectTable)
+        return self.construct_helper('objects', builder, ObjectTable, manager)
 
     @ObjectMapper.constructor_arg('object_keys')
     def object_keys(self, builder, manager):
-        return self.construct_helper('object_keys', builder, ObjectKeyTable)
+        return self.construct_helper('object_keys', builder, ObjectKeyTable, manager)
--- a/src/hdmf/common/io/resources.py
+++ b/src/hdmf/common/io/resources.py
@@ -7,10 +7,11 @@ from .. import register_map
 class ExternalResourcesMap(ObjectMapper):
 
     def construct_helper(self, name, parent_builder, table_cls, manager):
-    	"""Create a new instance of table_cls with data from parent_builder[name].
+        """Create a new instance of table_cls with data from parent_builder[name].
 
-    	   The DatasetBuilder for name is associated with data_type Data and container class Data,
-    	   but users should use the more specific table_cls for these datasets.```
+           The DatasetBuilder for name is associated with data_type Data and container class Data,
+           but users should use the more specific table_cls for these datasets.
+        """
         parent = manager._get_proxy_builder(parent_builder)
         builder = parent_builder[name]
         src = builder.source
--- a/src/hdmf/common/resources.py
+++ b/src/hdmf/common/resources.py
@@ -1,18 +1,10 @@
-import uuid
 import numpy as np
 import pandas as pd
 
 from . import register_class
-from ..container import Table, Row, Container
+from ..container import Table, Row, Container, AbstractContainer
 
 from ..utils import docval, popargs
-
-
-def _check_id(table, id):
-    if id >= 0 and len(table.which(id=id)) == 0:
-        return np.uint64(id)
-    else:
-        raise ValueError('id must be a non-negative integer that is not already in the table: %d' % id)
 
 
 class KeyTable(Table):
@@ -44,14 +36,15 @@ class ResourceTable(Table):
     __defaultname__ = 'resources'
 
     __columns__ = (
-        {'name': 'keytable_idx', 'type': (int, str, Key),
-         'doc': 'The user key that maps to the resource term / registry symbol.'},
+        {'name': 'keytable_idx', 'type': (int, Key),
+         'doc': ('The index into the keys table for the user key that '
+                 'maps to the resource term / registry symbol.')},
         {'name': 'resource_name', 'type': str,
          'doc': 'The resource/registry that the term/symbol comes from.'},
         {'name': 'resource_entity_id', 'type': str,
-         'doc': 'The unique resource identifier for the resource term / registry symbol.'},
+         'doc': 'The unique ID for the resource term / registry symbol.'},
         {'name': 'resource_entity_uri', 'type': str,
-         'doc': 'The unique resource identifier for the resource term / registry symbol.'},
+         'doc': 'The URI for the resource term / registry symbol.'},
     )
 
 
@@ -65,15 +58,16 @@ class ResourceEntity(Row):
 
 class ObjectTable(Table):
     """
-    A table for storing objects that contain keys that refer to external resources
+    A table for storing objects (i.e. Containers) that contain keys that refer to external resources
     """
 
     __defaultname__ = 'objects'
 
     __columns__ = (
-        {'name': 'object_id', 'type': (str, uuid.UUID),
-         'doc': 'The user key that maps to the resource term / registry symbol.'},
-        {'name': 'field', 'type': str, 'doc': 'The resource/registry that the term/symbol comes from.'},
+        {'name': 'object_id', 'type': str,
+         'doc': 'The object ID for the Container/Data'},
+        {'name': 'field', 'type': str,
+         'doc': 'The field on the Container/Data that uses an external resource reference key'},
     )
 
 
@@ -93,10 +87,10 @@ class ObjectKeyTable(Table):
     __defaultname__ = 'object_keys'
 
     __columns__ = (
-        {'name': 'objecttable_idx', 'type': (int, str, Object),
-         'doc': 'The user key that maps to the resource term / registry symbol.'},
-        {'name': 'keytable_idx', 'type': (int, str, Key),
-         'doc': 'The user key that maps to the resource term / registry symbol.'},
+        {'name': 'objecttable_idx', 'type': (int, Object),
+         'doc': 'the index into the objects table for the object that uses the key'},
+        {'name': 'keytable_idx', 'type': (int, Key),
+         'doc': 'the index into the key table that is used to make an external resource reference'}
     )
 
 
@@ -165,14 +159,15 @@ class ExternalResources(Container):
                                          table=self.resources)
         return resource_entity
 
-    @docval({'name': 'container', 'type': (Container, str), 'doc': 'the Container to add'},
+    @docval({'name': 'container', 'type': (str, AbstractContainer),
+             'doc': 'the Container/Data object to add or the object_id for the Container/Data object to add'},
             {'name': 'field', 'type': str, 'doc': 'the field on the Container to add'})
     def add_object(self, **kwargs):
         """
         Add an object that references an external resource
         """
         container, field = popargs('container', 'field', kwargs)
-        if isinstance(container, Container):
+        if isinstance(container, AbstractContainer):
             container = container.object_id
         obj = Object(container, field, table=self.objects)
         return obj
@@ -205,7 +200,9 @@ class ExternalResources(Container):
             return self.add_object(container, field)
 
     @docval({'name': 'key_name', 'type': str, 'doc': 'the name of the key to get'},
-            {'name': 'container', 'type': (str, Container), 'doc': 'the Container that uses the key', 'default': None},
+            {'name': 'container', 'type': (str, AbstractContainer), 'default': None,
+             'doc': ('the Container/Data object that uses the key or '
+                     'the object_id for the Container/Data object that uses the key')},
             {'name': 'field', 'type': str, 'doc': 'the field of the Container that uses the key', 'default': None})
     def get_key(self, **kwargs):
         key_name, container, field = popargs('key_name', 'container', 'field', kwargs)
@@ -214,7 +211,7 @@ class ExternalResources(Container):
             # if same key is used multiple times, determine
             # which instance based on the Container
             object_field = self._check_object_field(container, field)
-            key_tmp = self.object_keys['keytable_idx', object_field.id]
+            key_tmp = self.object_keys['keytable_idx', object_field.idx]
             if key_tmp in key_id:
                 return self.keys.row[key_tmp]
             else:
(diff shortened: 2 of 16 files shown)


src/hdmf/common/resources.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    @docval(*__columns__)
+    def add_row(self, **kwargs):
+        id, item = popargs('id', 'item', kwargs)
+        kwargs['id'] = _check_id(self, id)
+        if item >= 0:
+            kwargs['item'] = np.uint64(item)
+        else:
+            raise ValueError('item must be a non-negative integer: %d' % id)
```
```suggestion
            raise ValueError('item must be a non-negative integer: %d' % item)
```

oruebel: It looks like the code to check the item is the same as ``_check_id``. Maybe rename ``_check_id`` to ``_check_index`` and pass in key that needs to be used for the error message.

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,116 @@
...
+def _check_id(table, id):
+    if id >= 0 and len(table.which(id=id)) == 0:
+        return np.uint64(id)
+    else:
+        raise ValueError('id must be a non-negative integer that is not already in the table: %d' % id)
+
+
+@register_class('ResourceReferenceMap')
+class ResourceReferenceMap(Table):
+
```
oruebel, line 29: It would be nice to have docstrings on all classes in resources.py to describe what each of these classes is for.

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,116 @@
...
+        call_docval_func(super().__init__, kwargs)
+        if resource_map is None:
+            self.resource_map = ResourceReferenceMap()
+            if references is not None:
+                raise ValueError('Cannot specify references without specifying the accompanying resource map')
+        else:
+            self.resource_map = resource_map
+            self.references = references or ResourceReferences()
+
+    def get_crid(self, object_id, field, key):
```
```suggestion
    def get_resource_identifier(self, object_id, field, key):
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,116 @@
...
+        """Return the CRIDs (tuple of (resource, URI) tuples) associated with the given object_id, field, and key.
```
```suggestion
        """Return the resource identifiers (a.k.a. CRID)  associated with the given object_id, field, and key.
            :returns: Tuple of (resource, URI) tuples
```

tests/unit/common/test_resources.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_constructor(self):
+        self.build_tables()
+
+
+class TestExternalResources(H5RoundTripMixin, TestCase):
+
+    def test_add_reference(self):
+        ExternalResources()
+
+    def test_get_crid(self):
```
oruebel:
```suggestion
    def test_get_resource_identifier(self):
```

tests/unit/common/test_resources.py
```diff
@@ -0,0 +1,46 @@
...
+        rrmap, resrefs = TestResourceReferences.build_tables()
+        er = ExternalResources(rrmap, resrefs)
+        result = er.get_crid('ca885753-e8a3-418a-86f4-7748fc2252a8', 'foo', 'ATP Binding')
```
oruebel:
```suggestion
        result = er.get_resource_identifier('ca885753-e8a3-418a-86f4-7748fc2252a8', 'foo', 'ATP Binding')
```

tests/unit/common/test_resources.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertEqual(result[0][0], 'Gene Ontology')
+        self.assertEqual(result[0][1], 'http://amigo.geneontology.org/amigo/term/GO:0005524')
+
+        result = er.get_crid('da85e056-caff-4ddd-838c-5f5463e313e6', 'baz', 'Cacna1s')
```
oruebel:
```suggestion
        result = er.get_resource_identifier('da85e056-caff-4ddd-838c-5f5463e313e6', 'baz', 'Cacna1s')
```

src/hdmf/container.py
```diff
@@ -830,3 +833,218 @@ def __build_conf_methods(cls, conf_dict, conf_index, multi):
...
+
+
+class TableRegion(DataRegion):
+    '''
+    A class for representing regions i.e. slices or indices into an Table
+    '''
+
+    @docval({'name': 'name', 'type': str, 'doc': 'the name of this container'},
+            {'name': 'table', 'type': Table, 'doc': 'the Table this region applies to'},
+            {'name': 'region', 'type': (slice, list, tuple, RegionReference), 'doc': 'the indices of the table'})
```
oruebel: Why do we need to HDF5 region references here? The reason I am asking it that this seems to make the class depended on the I/O backend.

src/hdmf/container.py
```diff
@@ -830,3 +833,218 @@ def __build_conf_methods(cls, conf_dict, conf_index, multi):
         get = conf_dict.get('get')
         if get is not None:
             setattr(cls, get, cls.__make_get(get, attr, container_type))
+
+
+class Table(Data):
+    r'''
```
oruebel, line 951: I assume this for representing compound tables, correct? If so, I think it would be be useful to mention in the docstring what the difference between Table and DynamicTable is (i.e., one is a column-store allows you to add columns, whereas, Table is a row-store that allows us to store a table in a single data array)

src/hdmf/common/resources.py
oruebel, line 29:
```suggestion
    """A table for mapping user terms (i.e. keys) to resource entities."""
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,116 @@
...
+    @docval(*__columns__)
+    def add_row(self, **kwargs):
+        id = popargs('id', kwargs)
+        kwargs['id'] = _check_id(self, id)
+        return super().add_row(kwargs)
+
+
+@register_class('ResourceReferences')
+class ResourceReferences(Table):
+
```
oruebel, line 63:
```suggestion
    """
    A table for identifying which objects in a file contain
    values that correspond to external resource references. An
    external resource reference takes the form of an *id*, which
    is resolvable within a reference. Some of examples are
    GO:0005524, an entry in the Genome Ontology, ABE-0008876, an entry
    in the ASAP microbial genome database, or MGI:88294, an entry in the
    Mouse Genome Informatics database.
    """
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,116 @@
...
+        if item >= 0:
+            kwargs['item'] = np.uint64(item)
+        else:
+            raise ValueError('item must be a non-negative integer: %d' % id)
+        return super().add_row(kwargs)
+
+
+@register_class('ExternalResources')
+class ExternalResources(Container):
+
```
```suggestion
    """
    Container containing a ResourceReferences and ResourceReferenceMap table for
    tracking external resource references in a file.
    """
```

src/hdmf/container.py
ajtritt: Yeah, this is for working with datasets that have compound dtypes. It was formerly `NWBTable` in `pynwb`

src/hdmf/container.py
ajtritt: I think the idea was so you could create a `TableRegion` from an HDF5 region reference. This was pulled over from pynwb and was written years ago. I'm not sure we need it. The `get_region_slicer` function creates an import problem, too. So I'd be happy to axe the whole thing :)

oruebel: It looks like the class is not being used right now. As such, I think it is reasonable to remove it for now until we get an actual need.

rly:
Note this will also resolve #156.

satra:
@ajtritt - what's the benefit of the uuid? also are the different fields required? for example, i can imagine that name, id can come from the uri in many cases.

ajtritt:
> @ajtritt - what's the benefit of the uuid?

All HDMF Containers (e.g. ElectricalSeries, Device, etc) have a UUID assigned to them. the UUID is used to identify these objects within the file. Users almost never need to know about them and in this case, will not have to supply them. Where a UUID is passed in, a Container can be passed in, and we pull the UUID off of it. 

> also are the different fields required? for example, i can imagine that name, id can come from the uri in many cases.

theoretically, they could all be optional. Are you thinking of a case where the user just wants to provide the URL?

satra:
for example, in some cases, i may simply want to associate with species with the uri

`'species', 'http://purl.obolibrary.org/obo/NCBITaxon_10092'`

in other cases, we may want to populate name and curie using the uri, so 

`'species', 'http://purl.obolibrary.org/obo/NCBITaxon_10092'` would translate internally to:

`'species', 'http://purl.obolibrary.org/obo/NCBITaxon_10092', 'NCBITaxon:210727', 'western European house mouse'`

based on the content of the uri: http://www.ontobee.org/ontology/NCBITaxon?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCBITaxon_10092

which could be done through ontology browsing helper functions like ontospy.

ajtritt:
held up by https://github.com/hdmf-dev/hdmf/pull/459

ajtritt:
test **ci/circleci: pynwb-dev-python38** will not pass until https://github.com/NeurodataWithoutBorders/pynwb/pull/1317 is merged.

src/hdmf/common/io/resources.py
```diff
@@ -0,0 +1,31 @@
+from ...build import ObjectMapper
+from ..resources import ExternalResources, KeyTable, ResourceTable, ObjectTable, ObjectKeyTable
+from .. import register_map
+
+
+@register_map(ExternalResources)
+class ExternalResourcesMap(ObjectMapper):
+
+    def construct_helper(self, name, parent_builder, table_cls, manager):
```
rly, line 9:
```suggestion
    def construct_helper(self, name, parent_builder, table_cls, manager):
    	"""Create a new instance of table_cls with data from parent_builder[name].

    	   The DatasetBuilder for name is associated with data_type Data and container class Data,
    	   but users should use the more specific table_cls for these datasets.```

File: src/hdmf/container.py [hunk 13]
```diff
@@ -839,3 +842,269 @@ def __build_conf_methods(cls, conf_dict, conf_index, multi):
         get = conf_dict.get('get')
         if get is not None:
             setattr(cls, get, cls.__make_get(get, attr, container_type))
+
+
+class Row(object, metaclass=ExtenderMeta):
+
```
rly, line 850: Add documentation for the `__table__` class attribute

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+    __columns__ = (
+        {'name': 'keytable_idx', 'type': (int, str, Key),
+         'doc': 'The user key that maps to the resource term / registry symbol.'},
+        {'name': 'resource_name', 'type': str,
+         'doc': 'The resource/registry that the term/symbol comes from.'},
+        {'name': 'resource_entity_id', 'type': str,
+         'doc': 'The unique resource identifier for the resource term / registry symbol.'},
+        {'name': 'resource_entity_uri', 'type': str,
+         'doc': 'The unique resource identifier for the resource term / registry symbol.'},
```
rly: Please check all the `__columns__` docstrings. Many seem to be the same.

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+
+class ObjectTable(Table):
+    """
+    A table for storing objects that contain keys that refer to external resources
+    """
+
+    __defaultname__ = 'objects'
+
+    __columns__ = (
+        {'name': 'object_id', 'type': (str, uuid.UUID),
```
rly: Is there code to convert the uuid.UUID into a string somewhere, and is this tested? It might be easier to allow only strings.

rly: AbstractContainer.object_id returns a string anyway, so I am not sure when a user would have a `uuid.UUID` object

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+        resource_name = kwargs['resource_name']
+        resource_entity_id = kwargs['resource_entity_id']
+        resource_entity_uri = kwargs['resource_entity_uri']
+        if not isinstance(key, Key):
+            key = self.add_key(key)
+        resource_entity = ResourceEntity(key, resource_name, resource_entity_id, resource_entity_uri,
+                                         table=self.resources)
+        return resource_entity
+
+    @docval({'name': 'container', 'type': (Container, str), 'doc': 'the Container to add'},
```
```suggestion
    @docval({'name': 'container', 'type': (AbstractContainer, str), 'doc': 'the Container to add'},
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+            {'name': 'field', 'type': str, 'doc': 'the field on the Container to add'})
+    def add_object(self, **kwargs):
+        """
+        Add an object that references an external resource
+        """
+        container, field = popargs('container', 'field', kwargs)
+        if isinstance(container, Container):
```
```suggestion
        if isinstance(container, AbstractContainer):
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+        resource_entity_id = kwargs['resource_entity_id']
+        resource_entity_uri = kwargs['resource_entity_uri']
+        if not isinstance(key, Key):
+            key = self.add_key(key)
+        resource_entity = ResourceEntity(key, resource_name, resource_entity_id, resource_entity_uri,
+                                         table=self.resources)
+        return resource_entity
+
+    @docval({'name': 'container', 'type': (Container, str), 'doc': 'the Container to add'},
+            {'name': 'field', 'type': str, 'doc': 'the field on the Container to add'})
```
rly, line 164:
```suggestion
            {'name': 'field', 'type': str, 'doc': 'the field on the container to add', 'default': ''})
```

rly: The `field` will be tricky because fields can be remapped between the schema and the API.... The `field` here should specify the schema field, which the user does not yet have access to, I believe.

Relatedly, `Data` objects have a data field that does not map to a field in the schema, in which case the `field` should be None or empty string.

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+    def add_object(self, **kwargs):
+        """
+        Add an object that references an external resource
+        """
+        container, field = popargs('container', 'field', kwargs)
+        if isinstance(container, Container):
+            container = container.object_id
+        obj = Object(container, field, table=self.objects)
```
rly, line 172:
```suggestion
        if not field:
            field = ''
        obj = Object(container, field, table=self.objects)
```

src/hdmf/common/resources.py
rly: Thinking this through more, I think the `field` here should be the API class attribute name, and this function should map the `field` name to the schema name and store that.

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+            field_id = self.objects.which(field=field)
+            object_id = list(set(object_id) & set(field_id))
+
+        if len(object_id) == 1:
+            return self.objects.row[object_id[0]]
+        else:
+            return self.add_object(container, field)
+
+    @docval({'name': 'key_name', 'type': str, 'doc': 'the name of the key to get'},
+            {'name': 'container', 'type': (str, Container), 'doc': 'the Container that uses the key', 'default': None},
```
```suggestion
            {'name': 'container', 'type': (str, AbstractContainer), 'doc': 'the Container that uses the key', 'default': None},
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+        return ObjectKey(obj, key, table=self.object_keys)
+
+    def _check_object_field(self, container, field):
+        """
+        A helper function for checking if a container and field have been added.
+
+        If the container and field have not been added, add the pair and return
+        the corresponding Object. Otherwise, just return the Object.
+        """
+        object_id = self.objects.which(object_id=container)
```
rly: To reduce confusion, please change `object_id` to `object_idx`

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,355 @@
...
+        if len(object_id) > 0:
+            field_id = self.objects.which(field=field)
+            object_id = list(set(object_id) & set(field_id))
+
+        if len(object_id) == 1:
```
rly: I suggest adding a check here if len(object_id) > 1: raise warning and return the first one. We cannot guarantee that other APIs are ensuring only one row for a given (object, field)

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,357 @@
...
+            return self.add_object(container, field)
+        else:
+            raise ValueError("Found multiple instances of the same object_id and field in object table")
+
+    @docval({'name': 'key_name', 'type': str, 'doc': 'the name of the key to get'},
+            {'name': 'container', 'type': (str, AbstractContainer), 'default': None,
+             'doc': ('the Container/Data object that uses the key or '
+                     'the object_id for the Container/Data object that uses the key')},
+            {'name': 'field', 'type': str, 'doc': 'the field of the Container that uses the key', 'default': None})
+    def get_key(self, **kwargs):
```
rly, line 209:
```suggestion
    def get_key(self, **kwargs):
        """
        Return a Key or a list of Key objects that correspond to the given key_name.
        
        If container and field are provided, the Key that corresponds to the given key_name
        for the given container and field is returned.
        """
```

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,357 @@
...
+                key_tmp = self.object_keys['keytable_idx', object_field.idx]
+                if key_tmp in key_id:
+                    key = self.keys.row[key_tmp]
+                else:
+                    key = self.add_key(key)
+            else:
+                key = self.keys.row[key_id[0]]
+
+        if object_field is None:
+            object_field = self._check_object_field(container, field)
```
rly: Logic comment -- since object_field is set to this no matter what, can line 266 be changed to 
`object_field = self._check_object_field(container, field)` ?

src/hdmf/common/resources.py
```diff
@@ -0,0 +1,357 @@
...
+        resource_id = kwargs['entity_id']
+        resource_uri = kwargs['entity_uri']
+        add_rsc = False
+
+        if resource_name is not None and resource_id is not None and resource_uri is not None:
+            add_rsc = True
+        elif not (resource_name is None and resource_id is None and resource_uri is None):
+            msg = ("Specify all or none of resource_name, entity_id, and entity_uri arguments. "
+                   "All three are required to create a reference")
+            raise ValueError(msg)
```
rly, line 267: I think we had considered the use case of no (resource_name, entity_id, entity_uri) for a given (container, field, key), but I think that use case doesn't make any sense. Why would you add a key that is not associated with anything?

src/hdmf/container.py
```diff
@@ -839,3 +842,297 @@ def __build_conf_methods(cls, conf_dict, conf_index, multi):
...
+                func_args = list()
+                for col in columns:
+                    func_args.append(col)
+                func_args.append({'name': 'table', 'type': Table, 'default': None,
+                                  'help': 'the table this row is from'})
+                func_args.append({'name': 'idx', 'type': int, 'default': None,
+                                  'help': 'the index for this row'})
+
+                @docval(*func_args)
+                def __init__(self, **kwargs):
```
rly, line 902: I think there should be a base `Row.__init__` that initializes `self.__keys`, `self.__idx`, `self.__table`, and `self.__todict()` and the Row subclass should call that init. Otherwise if line 891 is false, several methods will break.

tests/unit/common/test_resources.py
```diff
@@ -0,0 +1,168 @@
+import pandas as pd
+
+from hdmf.common.resources import ExternalResources
+from hdmf.testing import TestCase, H5RoundTripMixin
+
+
+class TestExternalResources(H5RoundTripMixin, TestCase):
```
rly, line 7: These are all good tests for ExternalResources, but it would be ideal to have unit tests for the new Row, RowGetter, and Table classes as well.

src/hdmf/common/resources.py
ajtritt:
```python
self.add_ref('uuid1', 'field1', 'key1', 'rsc1', 'entity1', 'uri1')
key = self.get_key('key1')
self.add_ref('uuid2', 'field2', key)                 # reusing 'key1' 
self.add_ref('uuid3', 'field3', key, 'rsc2')         # reusing 'key1' or adding another resource without sufficient information? raise an error
```

src/hdmf/container.py
ajtritt: I'd rather just raise an error if `__table__` is not set.

src/hdmf/common/resources.py
rly: Hm. That's an interesting use case I think we had not thought about -- the same key being used for multiple objects. Probably rare, but we might as well support it.

rly:
Looking at https://codecov.io/gh/hdmf-dev/hdmf/pull/442/diff

Can you also write tests for `Table.to_dataframe()`, `Table.from_dataframe()`, and `ExternalResources.get_key()`? It would be nice to get 100% diff coverage for all of this new code (we'll have to write these tests eventually and it's better done now than later), but if not, these functions should be tested at least partially.

rly:
Otherwise, it looks good to me!

rly (approved):

PR #444, opened 2020-10-21
Fix case where non-AbstractContainer is base class

### LINKED ISSUE #443: Fix broken edge case of AbstractContainer multiple inheritance
## Description

`AbstractContainer.__gather_fields()` assumes that, if the class has multiple inheritance, that the highest base class (assuming parents are higher than their children) is an `AbstractContainer`. That might not always be the case and should be checked.  

## Steps to Reproduce

See https://github.com/oruebel/ndx-icephys-meta/issues/76

## Environment

    Python Executable: Conda 
    Python Version: Python 3.7
    Operating System: Windows
    HDMF Version: dev

## Checklist

rly:
## Motivation

Fix #443.

## Checklist

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -120,18 +120,19 @@ class AbstractContainer(metaclass=ExtenderMeta):
         # check whether this class overrides __fields__
         if len(bases):
             # find highest base class that is an AbstractContainer (parent is higher than children)
-            base_container_ind = -1
-            while base_container_ind*-1 <= len(bases) and not issubclass(bases[base_container_ind], AbstractContainer):
-                base_container_ind = base_container_ind - 1
-            base_fields = bases[base_container_ind]._get_fields()  # tuple of field names from base class
+            base_cls = None
+            for base_cls in reversed(bases):
+                if issubclass(base_cls, AbstractContainer):
+                    break
+            base_fields = base_cls._get_fields()  # tuple of field names from base class
             if base_fields is not fields:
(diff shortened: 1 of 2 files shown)


oruebel:
I'm wondering whether this is really a bug in HDMF or an error on the user side (i.e., in this case the ndx-icephys extension that I wrote). In Python in general the order in which you inherit from other classes matters, i.e., I'm wondering whether it may be better to throw an error to force developers (in this case me ;-)   ) to fix their classes early on, rather than trying to support "bad code"? 

The fix you added looks good, I'm just wondering whether trying to allow this could cause more trouble down the road, or whether this is a case that we have to support because it is a legitimate use-case?

rly:
In general, I agree, it's hard to think of a use case here. How about if a user wants to create a subclass of both `Container` and maybe `dict` and it is not obvious which class should go earlier/later in the hierarchy?

ajtritt:
I just ran into this problem. Reordering the parent classes did not solve my problem.

ajtritt (approved):

rly:
@ajtritt Does this branch solve the problem? Reordering the parent classes (assuming the last one is an `AbstractContainer` should have done it; otherwise there might be something else going on)>

ajtritt:
@rly Reordering solved the problem, but then exposed some other problem. I checked out this branch and this solved my problem, so I haven't dug into why reordering didn't work. Based on that, I'd say this fix is the best solution long-term

rly:
Sounds good

PR #446, opened 2020-10-29
Add get_docval_macro method

rly:
## Motivation

Add a method to get a tuple of the current values for a docval_macro, e.g., 'array_data' and 'scalar_data'. This is generally useful and specifically useful for working with the allowed types where a docval macro is used to restrict the argument types.

## How to test the behavior?
```python
from hdmf.utils import get_docval_macro
get_docval_macro('array_data')
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Add ability to specify a custom class for new columns to a `DynamicTable` that are not `VectorData`,
   `DynamicTableRegion`, or `VocabData` using `DynamicTable.__columns__` or `DynamicTable.add_column(...)`. @rly (#436)  
 - Add capability to add a row to a column after IO. @bendichter (#426)
+- Add method `hdmf.utils.get_docval_macro` to get a tuple of the current values for a docval_macro, e.g., 'array_data'  
+  and 'scalar_data'. @rly (#456)
 
 ### Bug fixes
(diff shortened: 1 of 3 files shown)


oruebel:
FYI, I added a variant of that also in https://github.com/hdmf-dev/hdmf/pull/98 The main difference is that I had implemented this as a hidden function and I also need to be able to get the full dict, not just a single item. We can update the Zarr PR to use this PR to get this in first, but it would be great to have the feature for ``key=None`` to get a copy of the full array as well, as I need it at least for implementing tests. 

```
def __get_docval_macros(key=None):
    """
    Return a deepcopy of __macros. Used in the test-suite to avoid hardcoding in the tests
    """
    if key is None:
        return _copy.deepcopy(__macros)
    else:
        return __macros[key]
```

oruebel:
Can we just add the ``macro=None`` case directly in this PR then I won't have to change in the other PR and just need to update the other places in the Zarr PR to use this function instead.

rly:
Sorry about the early request for review. I clicked it before reading your first comment.

I updated the `get_docval_macro` function to be basically the same as your function. 

I also added the macro `'data': []` to help show that this is used. When the `DataIO` class is loaded, it is added to the 'data' macro.

oruebel (approved):

oruebel:
I added a note to the TODO list in the Zarr PR to make sure I replace the function there with this one, to avoid accidentally creating two. This PR looks good to me.

PR #449, opened 2020-11-03
Fix DynamicTable slicing and add SimpleMultiContainer

ajtritt:
## Motivation

Fix DynamicTableRegion and add SimpleMultiContainer 

## How to test the behavior?
```python
from hdmf.common import SimpleMultiContainer
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
   and 'scalar_data'. @rly (#446)
 - Add SimpleMultiContainer, a data_type for storing a Container and Data objects together. @ajtritt (#449)
 - Support `pathlib.Path` paths in `HDMFIO.__init__`, `HDF5IO.__init__`, and `HDF5IO.load_namespaces`. @dsleiter (#439)
-
+- Use hdmf-common-schema 1.2.1. See https://hdmf-common-schema.readthedocs.io/en/latest/format_release_notes.html for details.
 
 ### Internal improvements
 - Refactor `HDF5IO.write_dataset` to be more readable. @rly (#428)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Add ability to specify a custom class for new columns to a `DynamicTable` that are not `VectorData`,
   `DynamicTableRegion`, or `VocabData` using `DynamicTable.__columns__` or `DynamicTable.add_column(...)`. @rly (#436)  
 - Add capability to add a row to a column after IO. @bendichter (#426)
+- Add SimpleMultiContainer, a data_type for storing a Container and Data objects together. @ajtritt (#449)
+- Fix bug in slicing tables with DynamicTableRegions. @ajtritt (#449)
 
 ### Bug fixes
 - Fix handling of empty lists against a spec with text/bytes dtype. @rly (#434)
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -306,10 +306,11 @@ class BuildManager:
 
     @docval({'name': 'builder', 'type': (GroupBuilder, DatasetBuilder), 'doc': 'the builder to check'},
             {'name': 'parent_data_type', 'type': (str, type), 'doc': 'the potential parent data_type'},
-            returns="a tuple with the type hierarchy", rtype=tuple)
+            returns="True if data_type of *builder* is a sub-data_type of *parent_data_type*, False otherwise",
+            rtype=bool)
     def is_sub_data_type(self, **kwargs):
         '''
-        Return True if data_type of *builder* is a sub-data_type of *parent_data_type*, False otherwise
+        Return whether or not data_type of *builder* is a sub-data_type of *parent_data_type*
         '''
         builder, parent_dt = getargs('builder', 'parent_data_type', kwargs)
         dt = self.get_builder_dt(builder)
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -304,6 +304,18 @@ class BuildManager:
         builder = getargs('builder', kwargs)
         return self.__type_map.get_builder_dt(builder)
 
+    @docval({'name': 'builder', 'type': (GroupBuilder, DatasetBuilder), 'doc': 'the builder to check'},
+            {'name': 'parent_data_type', 'type': (str, type), 'doc': 'the potential parent data_type'},
+            returns="a tuple with the type hierarchy", rtype=tuple)
+    def is_sub_data_type(self, **kwargs):
+        '''
+        Return True if data_type of *builder* is a sub-data_type of *parent_data_type*, False otherwise
+        '''
+        builder, parent_dt = getargs('builder', 'parent_data_type', kwargs)
+        dt = self.get_builder_dt(builder)
+        ns = self.get_builder_ns(builder)
+        return self.namespace_catalog.is_sub_data_type(ns, dt, parent_dt)
+
 
 class TypeSource:
     '''A class to indicate the source of a data_type in a namespace.
--- a/src/hdmf/common/multi.py
+++ b/src/hdmf/common/multi.py
@@ -1,0 +1,23 @@
+from ..container import Container, Data, MultiContainerInterface
+from ..utils import docval, call_docval_func, popargs
+
+from . import register_class
+
+
+@register_class('SimpleMultiContainer')
+class SimpleMultiContainer(MultiContainerInterface):
+
+    __clsconf__ = {
+        'attr': 'containers',
+        'type': (Container, Data),
+        'add': 'add_container',
+        'get': 'get_container',
+    }
+
+    @docval({'name': 'name', 'type': str, 'doc': 'the name of this container'},
+            {'name': 'containers', 'type': (list, tuple), 'default': None,
+             'doc': 'the Container or Data objects in this file'})
+    def __init__(self, **kwargs):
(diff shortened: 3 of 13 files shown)


src/hdmf/spec/namespace.py
```diff
@@ -327,6 +327,21 @@ def get_hierarchy(self, **kwargs):
             raise KeyError("'%s' not a namespace" % namespace)
         return spec_ns.get_hierarchy(data_type)
 
+    @docval({'name': 'namespace', 'type': str, 'doc': 'the name of the namespace containing the data_type'},
+            {'name': 'data_type', 'type': (str, type), 'doc': 'the data_type to check'},
+            {'name': 'parent_data_type', 'type': (str, type), 'doc': 'the potential parent data_type'},
+            returns="a tuple with the type hierarchy", rtype=tuple)
```
rly: Looks like this function returns a bool, so this should be updated.

src/hdmf/build/manager.py
```diff
@@ -304,6 +304,18 @@ def get_builder_dt(self, **kwargs):
         builder = getargs('builder', kwargs)
         return self.__type_map.get_builder_dt(builder)
 
+    @docval({'name': 'builder', 'type': (GroupBuilder, DatasetBuilder), 'doc': 'the builder to check'},
+            {'name': 'parent_data_type', 'type': (str, type), 'doc': 'the potential parent data_type'},
+            returns="a tuple with the type hierarchy", rtype=tuple)
```
rly: Looks like this function returns a bool, so this should be updated.

tests/unit/common/test_table.py
```diff
@@ -666,6 +666,57 @@ def test_repr(self):
...
+
+    def setUp(self):
+        self.table, self.target_table = self.make_tables()
+        super().setUp()
+
+    def setUpContainer(self):
+        multi_container = SimpleMultiContainer('multi', [self.table, self.target_table])
+        return multi_container
+
+    def setUpExtras(self):
```
rly: Where is this called? I don't recall this in `H5RoundTripMixin`

tests/unit/common/test_multi.py
```diff
@@ -0,0 +1,55 @@
+from hdmf.common import DynamicTable, VectorData, \
+    DynamicTableRegion, SimpleMultiContainer
+from hdmf.testing import TestCase, H5RoundTripMixin
+
+
+class DynamicTableRegionRoundTrip(H5RoundTripMixin, TestCase):
```
rly: I think the test for SimpleMultiContainer should not be dependent on DynamicTable and DynamicTableRegion working. It could be done with generic Container and Data objects.

src/hdmf/common/multi.py
```diff
@@ -0,0 +1,23 @@
...
+        'get': 'get_container',
+    }
+
+    @docval({'name': 'name', 'type': str, 'doc': 'the name of this container'},
+            {'name': 'containers', 'type': (list, tuple), 'default': None,
+             'doc': 'the Container or Data objects in this file'})
+    def __init__(self, **kwargs):
+        containers = popargs('containers', kwargs)
+        call_docval_func(super().__init__, kwargs)
+        self.containers = containers
```
rly, line 23: It looks like this `__init__` does not do anything different from the default `__init__` generated by `MultiContainerInterface`. Could it be removed?

src/hdmf/common/table.py
```diff
@@ -1046,12 +1044,36 @@ def get(self, arg, index=False, **kwargs):
...
             if not index:
-                ret = self.table.get(ret, **kwargs)
+                ret = self.table.get(ret, df=df, **kwargs)
+            return ret
+        elif isinstance(arg, slice) or isinstance(arg, (list, np.ndarray)):
+            idx = arg
+
+            # ensure a list/ndarray if indices
+            if isinstance(idx, slice):
+                idx = list(range(*idx.indices(len(self.data))))
```
rly: Why must idx be a list? Shouldn't indexing with a slice work just as well?

ajtritt: You can't directly index a list with a list, so if we are given a list, it has to be done manually (see line 1063). Putting this here simplified the conditional below. It might be possible to put this check under the if at line 1062.

rly: Sorry, I think I wasn't clear. If a slice is given, the code here converts the slice into a list of indices. Most (if not all) data types that a VectorData allows can take a slice as input, so it seems that converting the slice to a list is unnecessary. Also, some data types (like h5py.Dataset) may have specialized functions for handling slices that may be better than converting them into a list of indices. So I question whether these two lines are necessary. I think line 1065 should be able to handle slices.

src/hdmf/common/multi.py
ajtritt: The autogenerated one flips the order--`containers` is the first argument, `name` is the second. I prefer it the other way around

rly: OK

src/hdmf/common/table.py
```diff
@@ -1046,16 +1044,64 @@ def get(self, arg, index=False, **kwargs):
...
             ret = self.data[arg]
             if not index:
-                ret = self.table.get(ret, **kwargs)
+                ret = self.table.get(ret, df=df, **kwargs)
+            return ret
+        elif isinstance(arg, slice) or isinstance(arg, (list, np.ndarray)):
+            idx = arg
+
+            # get the data at the specified indices
+            if isinstance(self.data, list) and isinstance(idx, list):
```
```suggestion
            if isinstance(self.data, (list, tuple)) and isinstance(idx, list):
```

src/hdmf/common/table.py
```diff
@@ -1046,16 +1044,64 @@ def get(self, arg, index=False, **kwargs):
...
-            if np.issubdtype(type(arg), np.integer) and arg >= len(self.data):
+        elif np.issubdtype(type(arg), np.integer):
+            if arg >= len(self.data):
                 raise IndexError('index {} out of bounds for data of length {}'.format(arg, len(self.data)))
             ret = self.data[arg]
             if not index:
-                ret = self.table.get(ret, **kwargs)
+                ret = self.table.get(ret, df=df, **kwargs)
+            return ret
+        elif isinstance(arg, slice) or isinstance(arg, (list, np.ndarray)):
```
```suggestion
        elif isinstance(arg, (slice, list, np.ndarray)):
```

tests/unit/common/test_table.py
```diff
@@ -666,6 +666,147 @@ def test_repr(self):
...
+        table = mc.containers['table_with_dtr']
+        return table.get(arg, df=False)
+
+    def _getitem(self, arg):
+        mc = self.roundtripContainer()
+        table = mc.containers['table_with_dtr']
+        return table[arg]
+
+    def test_getitem_oor(self):
+        with self.assertRaisesRegex(IndexError, 'Row index 12 out of range'):
```
rly: I recommend using `self.assertRaisesWith` which tests for an exact match instead of potentially a partial match.

ajtritt: I can't find any documentation on that method. I'd rather not use an undocumented method.

rly: It's actually part of `hdmf.testing.TestCase` https://hdmf.readthedocs.io/en/stable/hdmf.testing.testcase.html

rly:
Before this is merged, hdmf-common-schema should be updated with the changes here

tests/unit/common/test_multi.py
```diff
@@ -0,0 +1,16 @@
+from hdmf.container import Container, Data
+from hdmf.common import SimpleMultiContainer
+from hdmf.testing import TestCase, H5RoundTripMixin
+
+
+class DynamicTableRegionRoundTrip(H5RoundTripMixin, TestCase):
```
rly:
```suggestion
class SimpleMultiContainerRoundTrip(H5RoundTripMixin, TestCase):
```

ajtritt:
@rly what is the process for that? this repo is pinned to the working commit from hdmf-common-schema

rly:
> @rly what is the process for that? this repo is pinned to the working commit from hdmf-common-schema

The dev branch of HDMF should always point to a public release of hdmf-common-schema to prevent any issues where a new version of HDMF is released that includes an unreleased schema. So we should:
1) release hdmf-common-schema 1.2.1 
2) update the git submodule in HDMF in this PR to point to the 1.2.1 tagged commit
3) merge this PR

https://hdmf-common-schema.readthedocs.io/en/latest/software_process.html#merging-prs-and-making-releases

rly:
Please update the hdmf-common-schema submodule:
```bash
cd src/hdmf/common/hdmf-common-schema
git pull origin master
git checkout 1.2.1
```

CHANGELOG.md
```diff
@@ -10,11 +10,14 @@
   @bendichter, @rly (#430)
 - Add capability to add a row to a column after IO. @bendichter (#426)
 - Add method `hdmf.utils.get_docval_macro` to get a tuple of the current values for a docval_macro, e.g., 'array_data'  
-  and 'scalar_data'. @rly (#456)
+  and 'scalar_data'. @rly (#446)
+- Add SimpleMultiContainer, a data_type for storing a Container and Data objects together. @ajtritt (#449)
 - Support `pathlib.Path` paths in `HDMFIO.__init__`, `HDF5IO.__init__`, and `HDF5IO.load_namespaces`. @dsleiter (#439)
```
```suggestion
- Use hdmf-common-schema 1.2.1. See https://hdmf-common-schema.readthedocs.io/en/latest/format_release_notes.html for details.
```

rly (approved):

PR #480, opened 2020-12-02
update HDMF to work with h5py 3.x

### LINKED ISSUE #458: Add support for h5py 3
h5py has released 3.0 and is already on 3.1. I am opening this issue to keep track of things we will need to be aware of as we move to support h5py 3

- [ ] You can no longer set a `list` of `str` as an attribute directly. It needs to be converted to a numpy array with a `h5py.special_dtype`. HDMF will need to be modified here to support this:
https://github.com/hdmf-dev/hdmf/blob/729746025f8cc27e96d02384c8cc4d9830284651/src/hdmf/backends/hdf5/h5tools.py#L859

- [ ] Datasets of variable length strings, whether UTF-8 or ASCII, get returned as a `bytes` object. 
  - check if h5py >= 3.0, use `h5py.Dataset.asstr()` wrapper and add that to docval macro `array_data`. `AsStrWrapper` does not allow write and does not inherit from `h5py.Dataset` so we either tell users to call `.asstr()` on read or we add a wrapper class that does this behind the scenes (best option). Both options are a major version bump. 
  - If we go this route, we may as well remove support for lists and tuples and require numpy arrays, which removes many edge cases and complexity. We can deprecate gradually by raising a warning whenever a list is used and saying it is being converted to a numpy array of dtype x.

- [ ] In previous versions, creating a dataset from a list of bytes objects would choose a fixed length string datatype to fit the biggest item. It will now use a variable length string datatype. To store fixed length strings, use a suitable dtype from h5py.string_dtype().
  - this probably does not affect us. we always store vlen strings.

ajtritt:
## Motivation

Fix #458 

This more or less passes along the changes h5py 3.x introduced. Instead of using their `AsStr` wrapper (see `StrDataset`), I added a similar version that has more functionality to work where an `h5py.Dataset` works. 

## How to test the behavior?
Run test suite with h5py 3.x installed


requirements-min.txt
```diff
@@ -1,6 +1,6 @@
 # package dependencies and their minimum versions for installing HDMF
 # the requirements here specify '==' for testing; setup.py replaces '==' with '>='
-h5py==2.9,<3  # support for setting attrs to lists of utf-8 added in 2.9
+h5py==3.1.0  # support for setting attrs to lists of utf-8 added in 2.9
```
rly: Do we need to cut h5py 2.9-2.10 compatibility already? I think we should try to maintain compatibility for some time, at least until ubuntu supports h5py 3.x. 

Ubuntu: current support 2.10: https://packages.ubuntu.com/search?suite=groovy&searchon=names&keywords=h5py
Fedora currently supports 3.1: https://src.fedoraproject.org/rpms/h5py
Debian currently supports 2.8: https://packages.debian.org/search?keywords=h5py&searchon=names&suite=stable&section=all

ajtritt: I'm not opposed to it, but we should discuss how we decide when to stop supporting 2.x.

yarikoptic: Ideally (unless too much effort) it would be great to retain compatibility with h5py 2.x since even if bleeding edge releases of distributions would support it, stable releases would not

bendichter (changes requested):
edit: nvm I found a problem but it's true in dev as well

satra:
can dev be merged into this so that pynwb can install with this branch?

satra:
thank you @rly

rly (approved):

oruebel (approved):

rly:
The miniconda36 test fails because the latest h5py does not have wheels released for python 3.6, so I suggest merging #620 into this branch before merging this into dev.

satra:
Python 3.6 will reach end of life in Dec 2021. It may be worthwhile dropping support for that at this stage. Just as h5py 2.x will not release wheels for 3.9.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -699,7 +702,7 @@ def __read_dataset(self, h5obj, name=None):
             if h5obj.dtype.kind == 'O' and len(h5obj) > 0:
                 elem1 = h5obj[0]
                 if isinstance(elem1, (str, bytes)):
-                    d = h5obj
+                    d = self._check_str_dtype(h5obj)
```
rly, line 705: @ajtritt This only covers 1D string datasets. Multi-dim string datasets stay as H5Datasets that return as bytes objects. Can we cover multi-dim string datasets too?

MWE:

```python
from hdmf.backends.hdf5.h5tools import HDF5IO
from hdmf.build import DatasetBuilder
import os
import numpy as np

a = np.array([['a', 'bb'], ['ccc', 'dddd']])
attr = 'foobar'
path = 'test_h5py3.h5'
try:
    with HDF5IO(path, mode='a') as io:
        io.write_dataset(io._file, DatasetBuilder('test_dataset', a, attributes={'test_attr': attr}, dtype='text'))
        io.close()

    with HDF5IO(path, 'r') as io:
        bldr = io.read_builder()
        print(str(bldr['test_dataset'].data))
        print(bldr['test_dataset'].data[:])
        print(type(bldr['test_dataset'].data[0]))
finally:
    os.remove(path)
```

CHANGELOG.md
```diff
@@ -3,6 +3,21 @@
...
+  - h5py 3 introduced [breaking changes regarding how strings are handled]
+  (https://docs.h5py.org/en/latest/whatsnew/3.0.html#breaking-changes-deprecations), specifically that
+  variable-length UTF-8 strings in datasets are now read as `bytes` objects instead of `str` by default.
+  To reduce the impact of this change on HDMF users, when HDMF reads a variable-length UTF-8 string
+  dataset, instead of returning an `h5py.Dataset` that is read as `bytes` objects, HDMF will return a
+  `hdmf.utils.StrDataset` object that extends `h5py.Dataset` and is read as `str` objects, which preserves
+  previous behavior. For example, under HDMF 2.x, an HDF5 dataset `d` with data ['a', 'b'] is read as a
+  `h5py.Dataset` object, and `d[:]` returns `str` objects. Under HDMF 3.x, the same dataset `d` is read
+  as a `hdmf.utils.StrDataset` object and `d[:]` still returns `str` objects.
+-
```
rly, line 17: @ajtritt I wrote this for the changelog. Please edit as you see fit.

src/hdmf/backends/hdf5/h5tools.py
rly: If not, we should note this clearly in the release notes.

rly:
I think this is good to merge into `rc/3.0.0`. @ajtritt please merge when ready. Thanks!

bendichter (approved):

src/hdmf/backends/hdf5/h5tools.py
ajtritt: How about now?

rly: Looks good. Thanks

rly (approved):

PR #481, opened 2020-12-02
Move Codecov CI to GitHub Actions

rly:
## Motivation

Codecov requires passing a private token when called from tox (our current method), but not when called directly from supported CI services such as CircleCI, Azure Pipelines, and GitHub Actions. This prevents coverage from being run when contributors create PRs. I moved coverage testing and uploading of coverage data to codecov to GitHub Actions. I had tried doing this in CircleCI but it was too complicated to integrate with the way we were currently matrixing the jobs. I could have created a new workflow just for coverage in both Circle and Azure, but I figured it would be easier to have coverage runs for all OS's in one place, and this would be a good opportunity to test out GitHub Actions. 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,13 +34,6 @@ references:
           # While pip installing tox does not output by default. Circle thinks task is dead after 10 min.
           no_output_timeout: 30m
       - run:
-          name: Run coverage and submit to codecov-io if COVERAGE_TOX_ENV != ""
-          command: |
-            . ../venv/bin/activate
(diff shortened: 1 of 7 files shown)


oruebel (changes requested):
Please update also the CI section https://github.com/hdmf-dev/hdmf/blob/dev/docs/source/software_process.rst accordingly, i.e., mention what GitHub Actions are being used for and update the parts on Circle/Azure as necessary.

oruebel (dismissed):

oruebel (dismissed):

oruebel (approved):

Issue #473, opened 2020-11-24
validating an hdmf file can no longer be done by executing the module


dsleiter:
## Description

The HDMF documentation shows the following process for validating a file: https://hdmf.readthedocs.io/en/stable/validation.html

In talking with @rly, this used to work prior to extracting HDFM out of PyNWB, and now only works through the `pynwb.validate`  module: https://pynwb.readthedocs.io/en/stable/validation.html

Two options for resolving are to:
1. remove this from the documentation
2. re-add this functionality matching the documentation

@rly suggested that we implement the second.

## Update

Option 1. has been implemented in #482 

This issue has been kept open until option 2 (i.e., adding validation functionality back) has been implemented. 

## Steps to Reproduce

Attempt to validate a hdmf file using the command provided in the documentation:
```
$ python -m hdmf.validate -p namespace.yaml test.h5
```

## Environment

    Python Executable: Conda
    Python Version: Python 3.7
    Operating System: Linux
    HDMF Version: 2.2.0

## Checklist

- [x] Have you ensured the feature or change was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?
- [x] Have you included a brief and descriptive title?
- [x] Have you included a clear description of the problem you are trying to solve?
- [x] Have you included a minimal code snippet that reproduces the issue you are encountering?
- [x] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?

dsleiter:
I tagged this as a bug because it sounds like it was intended functionality that no longer works, but I don't know what the bug criteria is for this project, so this could very easily be an enhancement.

I'd be happy to submit a PR for this. @rly suggested it could be very similar to this: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/validate.py

oruebel:
>     1. remove this from the documentation
> 
>     2. re-add this functionality matching the documentation

Depending on the timeline for the fix, I would suggest to do 1 as a quick "hot fix" (i.e, simply comment this out in the documentation) and then implement 2 (and add back the appropriate documentation) once ready. If the timeline for implementing this is only a few weeks, then simply doing 2 is fine, but if plan to have additional releases in-between, then I would suggest to do 1 first. 

> I tagged this as a bug

[...]

PR #482, opened 2020-12-02
Update documentation on validation to warn that the example does not work

dsleiter:
Ref #473 - until an implementation is added, here is an update to the documentation as @oruebel suggested.

## Motivation

As described in #473, the example command in validation documentation does not work.

## How to test the behavior?
Check the documentation change to see if you agree.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
 - Fix inheritance when non-`AbstractContainer` is base class. @rly (#444)
 - Fix use of `hdmf.testing.assertContainerEqual(...)` for `Data` objects. @rly (#445)
 - Add missing support for data conversion against spec dtypes "bytes" and "short". @rly (#456)
+- Update documentation on validation to indicate that the example command is not implemented @dsleiter (#482)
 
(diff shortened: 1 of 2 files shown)


dsleiter:
@oruebel You mentioned commenting-out the offending part of the documentation, is there a particular way to do that you have in mind? I couldn't seem to find a good solution for commenting in reStructuredText so it would remain in the .rst file but not appear in the compiled docs. But I'm open to suggestions on how better accomplish this.

t-b:
@dsleiter From https://docutils.sourceforge.io/docs/ref/rst/directives.html#id67

> Text following " .. " is a comment and is ignored.

rly (approved):

rly:
Thanks for updating the documentation on this, @dsleiter !

dsleiter:
Thanks @t-b. At first I had been trying to use `..` to comment out the reference to `validation.rst` within `index.rst` and had some difficulty (maybe because it was an element within a list), but trying again here it worked :-) I will have to brush up on my reStructuredText.

PR #484, opened 2020-12-02
Update circleci to fail job when publish token not available

dsleiter:
Ref #478

## Motivation

The CircleCI `deploy-dev` job fails silently when the publish token is not available.

## How to test the behavior?
```
* remove the publish token from circleCI and see if it fails
* alternatively, test a build on a fork without the token and see if it fails (this is what I did)
```

## Checklist


rly (approved):

rly:
Looks like CircleCI ran correctly and codecov also ran, probably because I pass the CircleCI env variable through to tox which runs codecov. I might also be able to do the same with Azure, but I would like to give GH Actions a try.

PR #485, opened 2020-12-02
Update CSRMatrix and push coverage to 100%, add to_uint_array util

rly:
## Motivation

With the change in hdmf-common-schema of dtypes in `CSRMatrix` from int to uint (https://github.com/hdmf-dev/hdmf-common-schema/pull/46), the API needs to be updated in order to avoid data type conversion warnings.

Also add testing of the different error and edge cases of `CSRMatrix`.

## Checklist

### DIFF
--- a/src/hdmf/common/sparse.py
+++ b/src/hdmf/common/sparse.py
@@ -23,7 +23,7 @@ class CSRMatrix(Container):
         data = getargs('data', kwargs)
         if isinstance(data, (np.ndarray, h5py.Dataset)):
             if data.ndim == 2:
-                data = sps.csr_matrix(self.data)
+                data = sps.csr_matrix(data)
             elif data.ndim < 2:
                 indptr, indices, shape = getargs('indptr', 'indices', 'shape', kwargs)
                 if any(_ is None for _ in (indptr, indices, shape)):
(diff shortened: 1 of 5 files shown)


ajtritt (dismissed):

src/hdmf/common/sparse.py
```diff
@@ -22,32 +22,37 @@ def __init__(self, **kwargs):
...
+            raise ValueError("Cannot convert '%s' to an array of unsigned integers." % arg) from ve
+        if ar.ndim != 1:
+            raise ValueError("'%s' must be a 1D array of unsigned integers." % arg)
+        return ar
 
     def __getattr__(self, val):
-        return getattr(self.__data, val)
-
-    @property
-    def shape(self):
```
oruebel, line 49: Is there a particular reason for removing the ``shape`` property?

oruebel (dismissed):

src/hdmf/common/sparse.py
rly: It is redundant with `self.__getattr__('shape')` which defers to `self.__data.shape` (equal to before), but then the values are converted to uints.

ajtritt (approved):

src/hdmf/common/sparse.py
oruebel: Thanks for the clarification. That makes sense.

oruebel (approved):

Issue #460, opened 2020-11-12
Different errors raised for accessing ragged DynamicTableRegions in a DynamicTable


rly:
## Description

In PyNWB, the Units table (a subclass of `DynamicTable`) has an indexed `DynamicTableRegion` column called "electrodes". Trying to access this column raises several errors:

Before write, `nwbfile.units[0]` should return a `pandas.DataFrame` for the first row but raises a Pandas error instead: `ValueError: could not broadcast input array from shape (2) into shape (1)`

After write and read, 
- `read_nwbfile.units['electrodes'][0]`
- `read_nwbfile.units[0, 'electrodes']`
- `read_nwbfile.units[0]`
- `read_nwbfile.units.to_dataframe()`

all return an h5py error: `TypeError: Argument sequence too long`

## Steps to Reproduce

```python
from pynwb import NWBFile, NWBHDF5IO, validate
import datetime


nwbfile = NWBFile(
    session_description='session_description',
    identifier='identifier',
    session_start_time=datetime.datetime.now(datetime.timezone.utc),
)

filename = 'nwbfile.nwb'

dev = nwbfile.create_device(
    name='dev'
)

group = nwbfile.create_electrode_group(
    name='group',
    description='description',
    location='location',
    device=dev
)

nwbfile.add_electrode(id=10, x=1.0, y=2.0, z=3.0, imp=0.0, location='x', filtering='x', group=group)
nwbfile.add_electrode(id=11, x=1.0, y=2.0, z=3.0, imp=0.0, location='x', filtering='x', group=group)
nwbfile.add_electrode(id=12, x=1.0, y=2.0, z=3.0, imp=0.0, location='x', filtering='x', group=group)

nwbfile.add_unit(
    id=1,
    spike_times=[0.1, 0.2, 0.3],
    electrodes=[1, 2],
)
nwbfile.add_unit(
    id=2,
    spike_times=[0.1, 0.2, 0.3],
    electrodes=[0, 2],
)
print(nwbfile.units['spike_times'][0])  # prints list of the 3 spike times of the first unit
print(nwbfile.units[0, 'spike_times'])  # prints list of the 3 spike times of the first unit
print(nwbfile.units.spike_times[0])  # prints first spike time (0.1) in table regardless of unit

print(nwbfile.units.electrodes)  # prints hdmf.common.table.DynamicTableRegion
print(nwbfile.units.electrodes.table)  # prints the electrode table
[...]

PR #491, opened 2020-12-04
Fix access of DynamicTableRegion of table with col of refs

rly:
## Motivation

Fix part of #460. Relevance in PyNWB to error when accessing rows of the Units table where electrodes are set. Odd that this has not come up yet, but perhaps users are not setting the electrodes column, or not reading the rows of the Units table...

## Checklist


rly:
In `query.HDMFDataset`, if `dataset` is an `h5py.Dataset` of references, then `dataset[(0, 1, 2)]` results in `TypeError: Argument sequence too long`. This is fixed by calling `__getitem__` with a **list** of indices rather than a tuple. But we also need to account for `dataset[()]` being a valid command that is very different from `dataset[[]]`.

oruebel (approved):

ajtritt (approved):

PR #492, opened 2020-12-08
Fix check quantity for non-data-type datasets

rly:
## Motivation

When datasets without a `data_type_def` or `data_type_inc` (and therefore must have a name) are encountered, the `ObjectMapper` raises an error when checking quantity. This PR should fix that error.

This is blocking the release of HDMF 3.0.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Block usage of h5py 3+. h5py>=2.9, <3 is supported. @rly (#461)
 - Block usage of numpy>=1.19.4 due to a known issue with numpy on some Windows 10 systems. numpy>1.16, <1.19.4 is supported.
   @rly (#461)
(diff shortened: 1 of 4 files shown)


rly:
TODO: Add a test for this.

rly:
I added tests. This is ready for review.

oruebel (approved):
It would be good to also raise a warning in the spec object when a users tries to  create an untyped object with a quantity other than `?`, `zero_or_one`, or `1`. Ignoring the quantity in the object mapper is probably fine here, because you should not be able to create an untyped object more than once, but in the spec we'd need to check for all three possible settings. Updating the spec object to warn on bad values does not need to be part of this PR, but we should probably make an issue to improve error checking in the Spec objects in general, and this is just one case that needs to be checked for.

rly:
Thanks. I agree, and I think that should be in a separate PR.

ajtritt (approved):

PR #496, opened 2020-12-08
Cache builders in HDF5IO.written map so it doesn't get gc'd

### LINKED ISSUE #494: Two build time test errors in connection with hdf5
## Bug when building under Debian

There was a [bug](https://bugs.debian.org/973148) which essentially boils down to

    ======================================================================
    FAIL: test_link_h5py_dataset_h5dataio_input (tests.unit.test_io_hdf5_h5tools.H5IOTest)
    ----------------------------------------------------------------------
      File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_hdmf/build/tests/unit/test_io_hdf5_h5tools.py", line 688, in test_link_h5py_dataset_h5dataio_input
        self.assertTrue(isinstance(self.f.get('test_softlink', getlink=True), SoftLink))
    AssertionError: False is not true

    ======================================================================
    FAIL: test_link_h5py_dataset_input (tests.unit.test_io_hdf5_h5tools.H5IOTest)
    ----------------------------------------------------------------------
      File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_hdmf/build/tests/unit/test_io_hdf5_h5tools.py", line 671, in test_link_h5py_dataset_input
        self.assertTrue(isinstance(self.f.get('test_softlink', getlink=True), SoftLink))
    AssertionError: False is not true

    ----------------------------------------------------------------------
    Ran 748 tests in 3.146s

    FAILED (failures=2, skipped=3, expected failures=1)

## Steps to Reproduce

I was able to reproduce this under Debian sid.

## Environment

    Python Executable: Python
    Python Version: Python 39
    Operating System: Debian GNU Linux
    HDMF Version: 2.2.0

Kind regards, Andreas.

rly:
## Motivation

Fix #494 

The bug reported in #494 is due to what I think is a change in garbage collector behavior in Python 3.9.
Consider the test code:
```python
self.io.write_dataset(self.f, DatasetBuilder('ds1', np.arange(10)))
self.io.write_dataset(self.f, DatasetBuilder('ds2', np.arange(10)))
```

`HDF5IO.write_dataset` checks whether the builder being passed to it has been written. It does this by checking the `HDF5IO._written_builders` dictionary to see if the ID of the builder is present and has value True. Only if the builder has not been written, will `write_dataset` continue to write the dataset.

I *think* Python 3.9 looks at the above code and garbage collects the first DatasetBuilder after the first line completes because nothing later in the code or in memory maintains a reference to the DatasetBuilder. So when the second DatasetBuilder is instantiated, it is *sometimes* given the same ID as the first. In those cases, `HDF5IO.write_dataset` will think that this new builder has already been written.

I see a few main ways to get around this. 1) update the test to create local variables for the first and second DatasetBuilder so that the ID is not reused by the second. Realistically, the BuildManager should be maintaining references to all builders anyway, so the error #494 is a quirk of the unit test. But maybe there are cases where the BuildManager does not have a reference to the builder. A better way might be: 2) update how `HDF5IO._written_builders` works so that it maintains a reference to the builder.

Interestingly and frustratingly, adding a `breakpoint` between the first and second lines above changes the flow so that the bug cannot be reproduced...

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@
 - Update documentation on validation to indicate that the example command is not implemented @dsleiter (#482)
 - Fix generated docval for classes with a LinkSpec. @rly (#478)
 - Fix access of `DynamicTableRegion` of a `DynamicTable` with column of references. @rly (#491)
+- Fix garbage collection issue in Python 3.9. @rly (#496)
 
 ## HDMF 2.2.0 (August 14, 2020)
 
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -457,7 +457,7 @@ class HDF5IO(HDMFIO):
         :return: True if the builder is found in self._written_builders using the builder ID, False otherwise
         """
         builder_id = self.__builderhash(builder)
-        return self._written_builders.get(builder_id, False)
+        return builder_id in self._written_builders
 
     def __builderhash(self, obj):
         """Return the ID of a builder for use as a unique hash."""
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -445,8 +445,6 @@ class HDF5IO(HDMFIO):
         :param builder: Builder object to be marked as written
         :type builder: Builder
         """
-        # currently all values in self._written_builders are True, so this could be a set but is a dict for
-        # future flexibility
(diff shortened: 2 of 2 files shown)


ajtritt (approved):
FTR, I think you can merge this as is i.e. without writing additional tests

rly:
Thanks. Yeah, I think this would be annoying to test.

PR #505, opened 2021-01-04
Allow np.bool_ as a valid type when bool is specified in spec

### LINKED ISSUE #504: Validator should allow np.bool_ when validating against a bool spec
## Description

`np.bool_` should be allowed when validating against a `bool` spec.

Ref: https://github.com/NeurodataWithoutBorders/pynwb/issues/1298

Support for `np.bool_` was added here: https://github.com/hdmf-dev/hdmf/pull/203 and it should be supported in the validator as well.

## Steps to Reproduce

See https://github.com/NeurodataWithoutBorders/pynwb/issues/1298

## Environment

    Python Executable: Conda
    Python Version: 3.8
    Operating System: Linux
    HDMF Version: 1.6.1
    PyNWB Version: 1.3.0

## Checklist

- [x] Have you ensured the bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues)?

dsleiter:
* Allow `np.bool_` as a valid type when `bool` is specified in a spec
* Fix #504
* Ref https://github.com/NeurodataWithoutBorders/pynwb/issues/1298
* Include bool as a scalar type for docval

## Motivation

To fix this pynwb issue: https://github.com/NeurodataWithoutBorders/pynwb/issues/1298

As it is currently possible to create an attribute with a `np.bool_` when the spec lists the dtype as `bool`, the validator should similarly allow `np.bool_` as a valid bool type.

While testing, I discovered that a `DatasetBulder` will not accept an `np.bool_` as valid data when the dtype is `bool` because `bool` is not included as a scalar type. This seemed like an inconsistency to me, so I added it as a scalar type for docval, but please let me know if you disagree and I'll revert the change.

## How to test the behavior?
```python
import numpy as np
from hdmf.spec import GroupSpec, DatasetSpec, AttributeSpec, SpecCatalog, SpecNamespace
from hdmf.build import GroupBuilder, DatasetBuilder


def set_up_vmap(dtype):
    spec_catalog = SpecCatalog()
    spec = GroupSpec('A test group specification with a data type',
                     data_type_def='Bar',
                     datasets=[DatasetSpec('an example dataset', dtype, name='data')],
                     attributes=[AttributeSpec('attr1', 'an example attribute', dtype)])
    namespace = SpecNamespace(


vmap = set_up_vmap('bool')
value = np.bool_(True)
bar_builder = GroupBuilder('my_bar',
                           attributes={'data_type': 'Bar', 'attr1': value},
                           datasets=[DatasetBuilder('data', value)])
vmap.validate(bar_builder)
```

## Checklist

### DIFF
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -32,7 +32,7 @@ class DtypeHelper:
         'long': ["int64", "long"],
         'utf': ["text", "utf", "utf8", "utf-8"],
         'ascii': ["ascii", "bytes"],
-        'bool': ["bool", "bool_"],
+        'bool': ["bool"],
         'int8': ["int8"],
         'uint8': ["uint8"],
         'uint16': ["uint16"],
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -32,7 +32,7 @@ class DtypeHelper:
         'long': ["int64", "long"],
         'utf': ["text", "utf", "utf8", "utf-8"],
         'ascii': ["ascii", "bytes"],
-        'bool': ["bool"],
+        'bool': ["bool", "bool_"],
         'int8': ["int8"],
         'uint8': ["uint8"],
         'uint16': ["uint16"],
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -23,7 +23,7 @@ class DtypeHelper:
     # make sure keys are consistent between hdmf.spec.spec.DtypeHelper.primary_dtype_synonyms,
     # hdmf.build.objectmapper.ObjectMapper.__dtypes, hdmf.build.manager.TypeMap._spec_dtype_map,
     # hdmf.validate.validator.__allowable, and backend dtype maps
-    # see https://hdmf-schema-language.readthedocs.io/en/latest/specification_language_description.html#dtype
+    # see https://hdmf-schema-language.readthedocs.io/en/latest/description.html#dtype
     primary_dtype_synonyms = {
         'float': ["float", "float32"],
         'double': ["double", "float64"],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Internal improvements
 - Update CI. @rly (#432)
 
+### Bug fixes
+- Allow `np.bool_` as a valid `bool` dtype when validating. @dsleiter (#505)
+
 ## HDMF 2.3.0 (December 8, 2020)
 
 ### New features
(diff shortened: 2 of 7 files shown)


src/hdmf/spec/spec.py
```diff
@@ -23,7 +23,7 @@ class DtypeHelper:
     # make sure keys are consistent between hdmf.spec.spec.DtypeHelper.primary_dtype_synonyms,
```
dsleiter, line 23: Per the comment here, I tried to maintain consistency between the type maps listed. However, it is very possible that there is some side-effect from adding support for `np.bool_` that I am not aware of, so I would appreciate if someone could do a double-check for me.

src/hdmf/spec/spec.py
```diff
@@ -32,7 +32,7 @@ class DtypeHelper:
         'long': ["int64", "long"],
         'utf': ["text", "utf", "utf8", "utf-8"],
         'ascii': ["ascii", "bytes"],
-        'bool': ["bool"],
+        'bool': ["bool", "bool_"],
```
rly:
```suggestion
        'bool': ["bool"],
```
This change is not necessary unless we want to add a new alias "bool_" for "bool" for dtype specs in the spec yaml.

rly (dismissed):

rly (dismissed):

src/hdmf/spec/spec.py
dsleiter: Ah, ok, that makes the purpose of `primary_dtype_synonyms` more clear, thanks for clarifying.

I added `bool_` here in order to have it added to `validator.__allowable` (see https://github.com/agencyenterprise/hdmf/blob/enh/np_bool_in_validator/src/hdmf/validate/validator.py#L18-L40), since that seemed like the most minimally invasive way to resolve the issue, but I didn't know the side-effects of updating `primary_dtype_synonyms`. Removing `bool_` here will make the new test fail, so we need to find another solution.

I originally wanted to do something like this: https://github.com/hdmf-dev/hdmf/pull/203 but that would either require a more significant change to how the validator checks types in `check_type` (currently by matching against lists in `__allowable`), or make `bool` a unique case.

I've found another implementation here which I think is a good middle-ground and relatively consistent with the existing implementation. Let me know what you think.

For the longer term - do you think this solution is acceptable, or should we add an issue to try and bring the type checking in the validator in line with the method used for type checking in docval (https://github.com/hdmf-dev/hdmf/pull/203)?

rly (approved):

rly:
Your change looks good and is where I would put it as well. Thanks!

dsleiter:
Awesome, thanks @rly!

PR #517, opened 2021-02-18
Updated External Resources

mavaylon1:
## Motivation

This update changes the functionality of the resource table and adds the entity table to be used in the ExternalResources class. With these updates came changes to some of the class methods in ExternalResources, the yaml schema, and the unit tests.

## How to test the behavior?
```
Create an instance of ExternalResources:
table=ExternalResources(name='test')

**add_resource:**
r=table.add_resource(name='resource_name',uri='resource_uri')

**add_entity:**
You have the option to either put in the index manually via an integer or you can refer to the resource itself to create the index.

table.add_entity(key='key1',resource_table_idx=0,entity_id='123',entity_uri='uri') 
or
table.add_entity(key='key1',resource_table_idx=r,entity_id='123',entity_uri='uri') 

**add_ref**
This class method is not new, but it has new parameters. You now have a column for the resource id, entity id, and entity uri. 

table.add_ref(container = some_container, field="", key='key1',resource_table_idx=r, entity_id='456',entity_uri='uri2')

This method also has the flexibility to create a new resource if the user has not made the resource they want to refer to. Simply leave resource_table_idx empty and instead use resource_name with resource_uri:

table.add_ref(container = some_container, field='', key='key2', resource_name='new_resource',resource_uri='new_uri',entity_id='112',entity_uri='entity_uri2')

Verify that all the new unit tests pass.
```

## Checklist


oruebel (commented):
Please remove the empty ``.DS_Store`` files

rly:
Please remove `.DS_Store` files and add `.DS_Store` to `.gitignore` to keep it from being added in the future.

src/hdmf/common/resources.py
```diff
@@ -253,15 +292,21 @@ def add_ref(self, **kwargs):
...
+            resource_name = kwargs['resource_name']
+            resource_uri = kwargs['resource_uri']
+            resource_table_idx = self.add_resource(resource_name, resource_uri)
 
-        if resource_name is not None and resource_id is not None and resource_uri is not None:
+        if ((resource_table_idx or resource_name) is not None and entity_id is not None and entity_uri is not None):
             add_rsc = True
-        elif not (resource_name is None and resource_id is None and resource_uri is None):
-            msg = ("Specify all or none of resource_name, entity_id, and entity_uri arguments. "
+        elif not ((resource_table_idx and resource_name is None) and entity_id is None and resource_uri is None):
```
rly: I believe based on the new logic above, `resource_table_idx` will never be None at this point. If `resource_table_idx` is None and `resource_name` is None, then the call to `self.add_resource` on line 304 will raise a error from the docval check saying that the name argument is missing.

I believe these checks should be split up in the following way: Above line 299, there should be a check that says if `resource_table_idx` is not None, then if `resource_name` is not None or `resource_table_idx` is not None, then raise an error. And if `resource_table_idx` is None, then either both `resource_name` and `resource_table_idx` are provided (if only one is provided, raise an error) or neither is provided. 

Then after resource_table_idx is possibly set, you can change line 306 to be:
`if (resource_table_idx is not None and entity_id is not None and entity_uri is not None):`
and line 308 to be:
`elif not (resource_table_idx is None and entity_id is None and resource_uri is None):`

The `add_rsc` flag variable should be renamed `add_entity` to reflect its new meaning (see line 339)

Please also add a test to ensure it is possible to run `add_ref` without any of the last five arguments.

src/hdmf/common/resources.py
```diff
@@ -140,24 +167,34 @@ def add_key(self, key):
...
-        resource_entity = ResourceEntity(key, resource_name, resource_entity_id, resource_entity_uri,
-                                         table=self.resources)
+        resource_entity = Entity(key, resource_table_idx, entity_id, entity_uri, table=self.entities)
         return resource_entity
 
+    @docval({'name': 'name', 'type': str, 'doc': 'the name of the ontology resource'},
+            {'name': 'uri', 'type': str, 'doc': 'uri associated with ontology resource'})
+    def add_resource(self, **kwargs):
+        """
+        Add onotology name and uri to ResourceTable that will be referenced by the ResourceTable idx.
```
```suggestion
        Add ontology name and uri to ResourceTable that will be referenced by the ResourceTable idx.
```

src/hdmf/common/resources.py
```diff
@@ -29,30 +29,53 @@ class Key(Row):
...
+    __defaultname__ = 'entities'
+
     __columns__ = (
         {'name': 'keytable_idx', 'type': (int, Key),
          'doc': ('The index into the keys table for the user key that '
                  'maps to the resource term / registry symbol.')},
-        {'name': 'resource_name', 'type': str,
-         'doc': 'The resource/registry that the term/symbol comes from.'},
-        {'name': 'resource_entity_id', 'type': str,
+        {'name': 'resource_table_idx', 'type': (int, Resource),
```
rly: We should be consistent with line 61. Either have an underscore before "table" or not.

rly: This should also be changed in the schema. Sorry I didn't catch this inconsistency there.

src/hdmf/common/resources.py
```diff
@@ -126,6 +152,7 @@ def __init__(self, **kwargs):
         super().__init__(name)
         self.keys = kwargs['keys'] or KeyTable()
         self.resources = kwargs['resources'] or ResourceTable()
+        self.entities = kwargs['entities'] or EntityTable()
         self.objects = kwargs['objects'] or ObjectTable()
         self.object_keys = kwargs['object_keys'] or ObjectKeyTable()
```
rly, line 158: Reviewing this code again, can you docval-ize the `add_key` method on line 159 -- change key to **kwargs and add docval to it, like the other methods. Then the `key` argument which maybe should be renamed `key_name` can have a docstring and type associated with it.

src/hdmf/common/resources.py
rly: I personally prefer an underscore before "table" which means this needs to be changed throughout the code and schema.

rly (changes requested):

docs/gallery/External_Resources.py
```diff
@@ -0,0 +1,144 @@
...
+
+###############################################################################
+# Introduction
+# ------------
+# The :py:class:`~hdmf.common.resources.ExternalResources` class provides a way
+# to organize and map user terms (keys) to multiple resources and entities
+# from the resources.
+
+###############################################################################
+# Creating an instance of the ExternalResources class_name
```
rly:
```suggestion
# Creating an instance of the ``ExternalResources`` class
```

docs/gallery/External_Resources.py
```diff
@@ -0,0 +1,144 @@
...
+from hdmf.common import ExternalResources
+from hdmf import Data
+import pandas as pd
+
+er = ExternalResources(name='test')
+
+###############################################################################
+# Using the add_ref method
+# ------------
+# add_ref will be one of the ExternalResources class methods that simplifies
```
rly, line 27:
```suggestion
# The :py:meth:`ExternalResources.add_ref <hdmf.common.resources.ExternalResources.add_ref>` method simplifies
```

docs/gallery/External_Resources.py
```diff
@@ -0,0 +1,144 @@
...
+# how to add new data to KeyTable, ObjectTable, EntityTable, and ResourceTable.
```
rly: End users will not know what these tables are yet and I think they need not know at this point. Can you introduce them after `add_ref` is called around line 44?

docs/gallery/External_Resources.py
rly: It's useful to end users for this tutorial to link to the API documentation. No need to do it now before the text is finalized, but just putting it here so that we remember.

docs/gallery/External_Resources.py
```diff
@@ -0,0 +1,144 @@
...
+
+# This method below returns an Object object.
+object_ = er._check_object_field(data, field="")
+
+key_object = er.get_key(key_name='one')
+
+er.add_external_reference(object_, key_object)
+
+# This return ObjectKeyTable
+er.object_keys.to_dataframe()
```
rly, line 155: This is a good start! I would suggest tweaking the wording to be more user focused -- what do these fields represent in the real world, how are they related to each other, and when would a user want to use ExternalResources. If you can also use real-world examples -- `Homo sapien` is good, but please use the IDs and URIs from the NCBI species ontology

rly (changes requested):

src/hdmf/common/resources.py
```diff
@@ -304,22 +358,28 @@ def add_keys(self, **kwargs):
...
             mask = keys == key
-            ret[key] = self.add_key(key)
-            for row in res_df[mask][[d['name'] for d in self.resources.__columns__[1:]]].to_dict('records'):
-                self.add_resource(ret[key], *row.values())
+            try:
+                self.get_key(key_name=key)
+            except Exception:
+                ret[key] = self.add_key(key)
+            else:
+                ret[key] = self.get_key(key_name=key)
```
rly: This seems inefficient and a catch-all `except Exception` is generally discouraged. I think this would also work:
```suggestion
            if key_name in self.keys:
                ret[key] = self.get_key(key_name=key)
            else:
                ret[key] = self.add_key(key)
```

rly (changes requested):

docs/gallery/External_Resources.py
rly: This will be updated in #539.

docs/gallery/External_Resources.py
rly: This will be updated in #539.

rly (approved):

rly:
Since the changes to the HDMF Common schema have not yet been released, I changed the base (target) for this PR to be a new branch that will track dev and hold changes related to this alpha version of the schema until it is released. 

@mavaylon1 please Squash and Merge this PR so that you get the commit credit. Thanks.

Issue #525, opened 2021-02-24
Validating links across extensions can fail


rly:
## Description

An IndexSeries type links to an ImageSeries. If an extension defines a new type ExtImageSeries that extends ImageSeries, then it should be valid for the IndexSeries to link to ExtImageSeries. However, validation fails.

This seems to be because the validator looks only at the current namespace and validates the types within that namespace, so ExtImageSeries is not known to the nwb.core namespace validator.

This is blocking correct validation of NWB files from the Allen Institute.

## Environment

    Python Executable: Conda
    Python Version: Python 3.8
    Operating System: Windows
    HDMF Version: dev

## Checklist

- [x] Have you ensured the bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?
- [x] Have you included a brief and descriptive title?
- [x] Have you included a clear description of the problem you are trying to solve?
- [x] Have you included a minimal code snippet that reproduces the issue you are encountering?
- [x] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?

t-b:
@rly 

> This is blocking correct validation of NWB files from the Allen Institute.

My turf (icephys data)?

rly:
@t-b This is actually for non-icephys data. 

Further investigation reveals that the validator is correctly not validating against the nwb.core namespace directly. Instead, what happens is that the file contains four extensions which include the nwb.core namespace and only one of them knows about ExtImageSeries. So when validating the file against any of the other three namespaces that do not know about ExtImageSeries, validation of the IndexSeries fails.

mavaylon1:
Hey @rly is this something I can mark for the next release on June 11?

PR #526, opened 2021-02-25
Ensure all specs are validated

### LINKED ISSUE #507: Ensure all specs are validated
## Description

Currently, there are a few edge cases where specs do not get validated, as raised in https://github.com/hdmf-dev/hdmf/pull/500. We want to ensure that all specs get correctly validated.

1. if a group includes two or more children with the same defined data type but a different name (independent of the quantity), then only the last instance will be validated. Eg. for an electrodes group that has VectorData X, Y, and Z - only Z will be validated. The validator should validate each of these children.
2. if a child link has the same data type as a child group or dataset, then the link will be validated but the group or dataset with the same data type will not be validated. The validator should validate both the link and the group or dataset.
3. if a group specifies two unnamed child groups of type A and B, where B inherits from A, then a single child of type B satisfies the spec for both B and A. The validator should require distinct children to match each spec. Both of the following cases should satisfy the spec: a) one child of type A and one of type B, or b) two children of type B.

## Steps to Reproduce

Create an hdmf object and spec which should be invalid according to one of the three points above, and see that it passes validation.

## Environment

    Python Executable: Conda or Python
    Python Version: Python 3.5, 3.6, 3.7, or 3.8
    Operating System: Windows, macOS or Linux
    HDMF Version: 2.3.0

## Checklist

- [x] Have you ensured the bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?

dsleiter:
## Motivation

* Fix #507
* Significant refactoring of the GroupValidator was done in an attempt to make it more easily understandable and facilitate future development
* The logic for matching group builder children against group spec children was extracted to a new SpecMatcher class in order to isolate the matching algorithm, which will need to be improved in future iterations
* New unit tests were added for the improvements listed in #507
* Refactoring was done with future validator issues in mind, such as raising a warning for superfluous builders attached to a group (https://github.com/NeurodataWithoutBorders/pynwb/issues/1332 and https://github.com/NeurodataWithoutBorders/pynwb/issues/942)

I realize that this is a pretty significant change to the GroupValidator code, and am very open to critical comments. Do you think the changes improve readability enough to warrant the change? Is there any clear misunderstanding of how validation should be done?

## How to test the behavior?

Execute the unit tests. All previously existing tests still pass after refactoring, in addition to the new tests.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Bug fixes
 - Fix CI testing on Python 3.9. @rly (#523)
+- Fix certain edge cases where `GroupValidator` would not validate all of the child groups or datasets
+  attached to a `GroupBuilder`. @dsleiter (#526)
 
 ## HDMF 2.4.0 (February 23, 2021)
 
--- a/src/hdmf/spec/spec.py
+++ b/src/hdmf/spec/spec.py
@@ -483,13 +483,18 @@ class BaseStorageSpec(Spec):
 
     @property
     def data_type_inc(self):
-        ''' The data type of this specification '''
+        ''' The data type this specification inherits '''
         return self.get(self.inc_key())
 
     @property
     def data_type_def(self):
         ''' The data type this specification defines '''
         return self.get(self.def_key(), None)
+
+    @property
+    def data_type(self):
+        ''' The data type of this specification '''
+        return self.data_type_def or self.data_type_inc
 
     @property
     def quantity(self):
--- a/src/hdmf/validate/validator.py
+++ b/src/hdmf/validate/validator.py
@@ -2,7 +2,7 @@ import re
 from abc import ABCMeta, abstractmethod
 from copy import copy
 from itertools import chain
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 
 import numpy as np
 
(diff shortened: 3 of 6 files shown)


dsleiter:
I marked the PR as WIP because there are a few questions to resolve, and I wanted to get some feedback from you all if this is on the right track or not.

rly:
This is great. I was diving into the `GroupValidator` code earlier today to try to handle #525 and I ended up refactoring a few things, but did not clean up as much as you did (I'll hold off until this is merged). 

On first pass, the changes all look good to me. I'll take a deeper look tomorrow.

dsleiter:
Sounds good @rly, sorry this one took so long to complete! I expect to be able to make smaller, more frequent contributions going forwards.

rly (approved):

rly:
This looks great to me. Can you create a new issue ticket documenting the edge cases where the validator does not work correctly. For example in your last test case that is skipped, when a builder can match multiple specs, it matches against the first spec rather than the most specific spec (in terms of type hierarchy). And if there are any others that you thought of while going through this. Thanks for all your work on this!

dsleiter:
Thanks @rly, glad it looks good!

Here are some new issues based on the discussion here:
* https://github.com/hdmf-dev/hdmf/issues/528
* https://github.com/hdmf-dev/hdmf/issues/529
* https://github.com/hdmf-dev/hdmf/issues/530
* https://github.com/hdmf-dev/hdmf/issues/531

There are a few items from my code comments above where I'm not sure if an issue is required:
1. should we ideally use docval on all classes no matter what, or is it fine to leave it off of module-level utility classes that are not intended or likely to be used outside of the module (eg. `SpecMatcher` and `SpecMatches` in the validator)? Happy to add docval if that is ideal.
2. is the new property `data_type` on `BaseStorageSpec` something that fits well conceptually with hdmf and the schema, or is it just something that has utility within the codebase? If so, then we could add a `neurodata_type` property to nwb specs, and add a clarification in the `hdmf ` documentation about `data_type` and its relationship to `data_type_def` and `data_type_inc`. In addition, it might be beneficial to refactor out the multiple checks for `data_type_def` and `data_type_inc` throughout the code, although that doesn't make any functional improvements.

oruebel:
>     1. should we ideally use docval on all classes no matter what

Use of docval is primarily required when type-validation is important, e.g. for Container classes that are relevant for I/O. Use of docval can also help to ensure consisting formatting of doc-strings, but if type validation is not needed, then just using regular doc-strings is fine (but we should make sure the docstrings are formatted correctly). In some rare cases, when a function is called a lot (e.g., 1000s of times) it can be useful to omit docval to avoid the overhead of validation in lieu of performance.

dsleiter:
@oruebel thanks, makes sense. In this case, I don't think performance is a limiting factor here, and I guess there isn't a good reason not to use docval, so I'll go ahead and add it in (probably with a PR for another validator issue, unless you think it's better as a separate issue)

oruebel:
> , unless you think it's better as a separate issue

Doing this as part of another PR is perfectly fine

PR #522, opened 2021-02-23
automatically create dynamic tables

bendichter:
## Motivation

fix #234

be able to use `get_class` on dynamic tables.

Works and tested for:
* regular columns
* DynamicTableRegion columns
* indexed columns

PR #533, opened 2021-03-03
Refactor MCI to remove need for default init

rly:
## Motivation

This PR refactors `MultiContainerInterface` to remove the `__init__` method which is a hurdle in refactoring dynamic class generation (see #522). To dynamically generate a `MultiContainerInterface`, we have to call both `super().__init__` of the new class and `MultiContainerInterface.__init__`. The only thing that `MultiContainerInterface.__init__` does is initialize the attributes to LabelledDict. Since this is an object property as opposed to a class property it cannot be done at the class level. This PR moves initialization of the attributes to the getter property if it is not already defined. This might not be the best place to do so, so alternative solutions are welcome.

## Checklist

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -575,7 +575,6 @@ class MultiContainerInterface(Container):
         if cls is MultiContainerInterface:
             raise TypeError("Can't instantiate class MultiContainerInterface.")
         if not hasattr(cls, '__clsconf__'):
-            # either the API was incorrectly defined or only a subclass with __clsconf__ can be initialized
             raise TypeError("MultiContainerInterface subclass %s is missing __clsconf__ attribute. Please check that "
                             "the class is properly defined." % cls.__name__)
         return super().__new__(cls, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


src/hdmf/container.py
```diff
@@ -759,7 +745,26 @@ def _func(self, **kwargs):
         return _func
 
     @classmethod
-    def __make_setter(cls, nwbfield, add_name):
+    def __make_getter(cls, attr):
```
ajtritt, line 748: Just to make sure I understand/remember correctly.  We need this to set the parent/child stuff on the multiple containers in this MCI. i.e. we can't use Container.__make_setter (or whatever it's called) because it's just not aware of the multiple containers being in a LabelledDict.  

Is that correct?

rly: We can't use __make_setter because other code assumes that `getattr(self, attr_name)` is a LabelledDict. For example in the add, get, and getitem methods (and add/get are called by many things). I could move the initialization of `fields['attr']` there.

ajtritt: I think it's fine the way it is. I was just trying to understand what was going on.

ajtritt (approved):

PR #534, opened 2021-03-04
[WIP] Extract class generation code out of TypeMap

rly:
## Motivation

@ajtritt I would appreciate if you could look over this approach. It feels a bit clunky. Is this better than the current approach?

This PR extracts class generation code (`get_class`) code out of `TypeMap` into `ClassGenerator`, `CustomClassGenerator`, and `MCIClassGenerator`. `TypeMap` contains a `ClassGenerator` object and `type_map.get_container_cls(...)` calls `class_generator.generate_class(...)`.

`ClassGenerator` contains a list of `CustomClassGenerator` objects which allow an API to add specialized behavior during class generation, such as generating `__fields__` and `__init__` for all `Container` classes (which is all classes), `__clsconf__` for new classes that should inherit from `MultiContainerInterface`, and `__columns__` for new classes that inherit from `DynamicTable`. 

How this works: during class generation, `type_map.get_container_cls(...)` calls `class_generator.generate_class(...)` which iterates through the list of registered custom class generators and passes the mutable `clsdict` dict and `bases` list. Each registered generator will modify `clsdict` and `bases` as it sees fit, which might involve overriding what a previous generator does. These will ultimately be passed to `ExtenderMeta` to generate the new class. By default, `CustomClassGenerator` and `MCIClassGenerator` are registered in that order (order matters). In hdmf-common `__init__.py`, a `DynamicTableClassGenerator` class is registered. These classes contain a class method `update_cls_args(...)` which is called by `class_generator.generate_class(...)`. The classes cannot be instantiated.

This is a WIP. `DynamicTableClassGenerator` is not yet fully implemented, so tests associated with it fail.

I also refactored dynamic class generation tests and ObjectMapper.convert_dtype tests to their own files to improve code organization.

DynamicTableClassGenerator code and tests are adapted from @bendichter's #522.

## Checklist


bendichter:
I agree the get_class code is complicated but I'm not sure this is much better than what we currently have

src/hdmf/build/classgenerator.py
```diff
@@ -0,0 +1,336 @@
...
+
+import numpy as np
+
+from ..container import Container, Data, DataRegion, MultiContainerInterface
+from ..spec import AttributeSpec, LinkSpec, RefSpec
+from ..spec.spec import BaseStorageSpec, ZERO_OR_MANY, ONE_OR_MANY
+from ..utils import docval, getargs, ExtenderMeta, get_docval, fmt_docval_args
+
+
+class ClassGenerator:
```
ajtritt, line 12: could this functionality just be part of TypeMap?

rly: Yes. The only benefit of having it outside TypeMap is to organize it as separate, isolated functionality, but there is so little code, that I'd be fine moving it to TypeMap.

src/hdmf/common/io/table.py
```diff
@@ -43,3 +44,63 @@ def get_attr_value(self, **kwargs):
...
+
+
+class DynamicTableGenerator(CustomClassGenerator):
+
+    @classmethod
+    def update_cls_args(cls, classdict, bases, not_inherited_fields, spec, type_map):
+        """Update the given class dict and base classes if there is a field spec with quantity * or +
+        :param classdict: The dict to update with __clsconf__ if applicable
+        :param bases: The list of base classes to update if applicable
+        :param not_inherited_fields: Dict of additional fields that are not in the base class
```
ajtritt, line 56: by `base class` do you mean the base data_type/included spec?

rly: I mean the parent type. For DynamicTable, that is Container. So `not_inherited_fields` contains specs for attributes 'description' and 'colnames', and datasets 'id' and included 'VectorData'.

src/hdmf/build/classgenerator.py
```diff
@@ -0,0 +1,336 @@
...
+                    create='create_{}'.format(f)
+                ))
+                # remove the attribute from __fields__ -- needed so that MCI initializes the attribute to LabelledDict
+                classdict['__fields__'] = tuple(filter(lambda x: x['name'] != f, classdict['__fields__']))
+
+        if len(clsconf):
+            classdict.update(__clsconf__=clsconf)
+
+            # add MultiContainerInterface to the type hierarchy if not already there
+            if not isinstance(bases[0], MultiContainerInterface):
```
ajtritt, line 335: when does this happen?

ajtritt:
This looks like a great start. The only thing that really concerns me is the importance of order.

src/hdmf/build/classgenerator.py
rly: Good point. I can't think of a scenario.

rly:
Closed in favor of the approach in #535

PR #535, opened 2021-03-05
Extract class generation code out of TypeMap

rly:
## Motivation

Similar to #534, but after trying to adapt the code to handle DynamicTable class generation, it got a bit messy and was not as much of an improvement as I would like.

This is an attempt to improve #534. 

How this works: during class generation, `type_map.get_container_cls(...)` calls `class_generator.generate_class(...)` which iterates through each **field spec** and iterates through each custom class generator. The generator can then choose to process the field spec or not. If so, no other generator can process it. If not, the next generator in the list can choose to process it. Processing a field spec involves updating the mutable `classdict` dict and mutable `docval_args` list and adding the field to either `__fields__`, `__clsconf__`, or `__columns__`. **Importantly, this approach removes the need to undo the actions of other generators -- only one generator can act on a field.** It also handles setting up docval and init better than in #534. Afterward, `post_process` of each generator is called to change lists to tuple and any other actions before `__init__` is generated.

By default, `DynamicTableClassGenerator` is called first, then `MCIClassGenerator`, then `CustomClassGenerator`. The classes contain only class methods and cannot be instantiated.

@ajtritt What do you think? Is this better? The downside I see is that I separated the singular `update_cls_dict` from before into multiple methods so that each `CustomClassGenerator` can hook into different parts of the logic.

`DynamicTableClassGenerator` works AFAIK. Remaining `_build_docval` tests currently fail and need to be updated. 

## Checklist


src/hdmf/build/classgenerator.py
```diff
@@ -0,0 +1,327 @@
...
+            bases = [parent_cls]
+            docval_args = list(deepcopy(get_docval(parent_cls.__init__)))
+            for f, field_spec in not_inherited_fields.items():
+                for class_generator in self.__custom_generators:
+                    # each generator can update classdict and docval_args
+                    if class_generator.apply_generator_to_field(field_spec, bases, type_map):
+                        class_generator.process_field_spec(classdict, parent_cls, f, field_spec, not_inherited_fields,
+                                                           type_map)
+                        class_generator.update_docval_args(docval_args, f, field_spec, type_map)
+                        break  # each field_spec should be processed by only one generator
```
rly: This could be combined into a single method
`class_generator.process_field_spec(...)` that returns True/False. If return True, then it has been processed, and break the loop.

rly:
This is now good to review @ajtritt

rly:
And FWIW, I added a bunch of tests so the new classgenerator.py has 96% coverage. The remaining 4% comes from resolving the type of attribute references. This functionality needs to be improved so that it doesn't just use generic values (Container, Data) when the container class cannot be found in the TypeMap.

src/hdmf/build/classgenerator.py
```diff
@@ -0,0 +1,344 @@
...
+    def __new__(cls, *args, **kwargs):  # pragma: no cover
+        raise TypeError('Cannot instantiate class %s' % cls.__name__)
+
+    # mapping from spec types to allowable python types for docval for fields during dynamic class generation
+    # e.g., if a dataset/attribute spec has dtype int32, then get_class should generate a docval for the class'
+    # __init__ method that allows the types (int, np.int32, np.int64) for the corresponding field.
+    # passing an np.int16 would raise a docval error.
+    # passing an int64 to __init__ would result in the field storing the value as an int64 (and subsequently written
+    # as an int64). no upconversion or downconversion happens as a result of this map
+    _spec_dtype_map = {
```
ajtritt, line 122: Does this exist anywhere else? Or was it moved here from objectmapper.py?

src/hdmf/build/classgenerator.py
```diff
@@ -0,0 +1,344 @@
...
+        'utf-8': str,
+        'ascii': bytes,
+        'bytes': bytes,
+        'bool': (bool, np.bool_),
+        'isodatetime': datetime,
+        'datetime': datetime
+    }
+
+    @classmethod
+    def _get_scalar_type_map(cls, spec_dtype):
```
ajtritt: This function name is kind of confusing. Is it exposed for overwriting?

ajtritt (dismissed):
Looks good to me.... Nice work, kind of funny that we tried solving this with just a single method at one point.

src/hdmf/build/classgenerator.py
ajtritt: If this exists somewhere else, maybe we should move it to central location and import it from there. No need to do that in this PR though.

rly: It was moved from `manager.py`. It's used only in docval generation, so I feel like it makes sense to live in this module if not in this class. There are related but separate mappings in:
1. ObjectMapper: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/build/objectmapper.py#L100
2. spec.DtypeHelper: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/spec/spec.py#L27
3. Validator: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/validate/validator.py#L20

that perhaps we should merge into a central utility module for dtype handling.

src/hdmf/build/classgenerator.py
rly: This used to be `TypeMap.__get_scalar_type_map` so we can feel free to change it. I made it single underscore in case a subclass of `CustomClassGenerator` wants to customize it but maybe that is not necessary.

rly: How about renaming it to `_get_type_from_spec_dtype`

ajtritt: I think it's fine for now. as long as it's not out in the wild and it's documented, I think it's fine.

src/hdmf/build/classgenerator.py
ajtritt: > perhaps we should merge into a central utility module for dtype handling

Yeah, I think at some point we should do that.

ajtritt (approved):

PR #536, opened 2021-03-10
Generate DynamicTable using new get_class structure

rly:
## Motivation

This PR demonstrates the use of #535 to allow generation of `DynamicTable` using `get_class`. More tests should be added before merging. This is based on @bendichter 's #522.

## Checklist


tests/unit/common/test_generate_table.py
```diff
@@ -0,0 +1,115 @@
...
+                    dtype='float'
+                ),
+                DatasetSpec(
+                    data_type_inc='VectorData',
+                    name='indexed_col',
+                    doc='a test column',
+                    dtype='float'
+                ),
+                DatasetSpec(
+                    data_type_inc='VectorData',
```
rly: This should be a VectorIndex

tests/unit/common/test_generate_table.py
```diff
@@ -0,0 +1,115 @@
...
+            ]
+        )
+
+        self.dt_spec2 = GroupSpec(
+            'A test extension that contains a dynamic table',
+            data_type_def='TestDTRTable',
+            data_type_inc='DynamicTable',
+            datasets=[
+                DatasetSpec(
+                    data_type_inc='DynamicTableRegion',
```
rly, line 64: need to test indexed DTR

tests/unit/common/test_generate_table.py
```diff
@@ -0,0 +1,115 @@
...
+                    name='indexed_col',
+                    doc='a test column',
+                    dtype='float'
+                ),
+                DatasetSpec(
+                    data_type_inc='VectorData',
+                    name='indexed_col_index',
+                    doc='a test column',
+                    dtype='float'
+                ),
```
rly, line 40: test optional columns

rly:
And please add changelog and remove print statements from test

bendichter:
@rly can you review? I can't list you as a reviewer, since you opened this PR

src/hdmf/common/io/table.py
```diff
@@ -43,3 +43,59 @@ def get_attr_value(self, **kwargs):
...
+        """
+        if attr_name.endswith('_index'):  # do not add index columns to __columns__
+            return
+        field_spec = not_inherited_fields[attr_name]
+        column_conf = dict(name=attr_name, description=field_spec['doc'])
+        if not field_spec.required:
+            column_conf['required'] = False
+        dtype = cls._get_type(field_spec, type_map)
+        if issubclass(dtype, DynamicTableRegion):
+            column_conf['table'] = True
```
rly, line 78:
```suggestion
            # the spec does not know which table this DTR points to
            # the user must specify the table attribute on the DTR after it is generated
            column_conf['table'] = True
```

tests/unit/common/test_generate_table.py
```diff
@@ -0,0 +1,181 @@
...
+        TestDTRTable = self.TestDTRTable
+
+        @docval({'name': 'ref_col', 'type': int, 'doc': 'references table'},
+                allow_extra=True)
+        def add_row(self, **kwargs):
+            super(TestDTRTable, self).add_row(**kwargs)
+            for arg, table in zip(['ref_col', 'indexed_ref_col'], [test_table, test_table]):
+                col = self[arg].target if isinstance(self[arg], VectorIndex) else self[arg]
+                if col.table is None:
+                    col.table = table
```
rly: I think the following, inserted at line 175, would accomplish the same while being easier to read:
```python
test_dtr_table = self.TestDTRTable(name='test_dtr_table', description='my table')
test_dtr_table['ref_col'].table = test_table
test_dtr_table['indexed_ref_col'].target.table = test_table
```

It looks like a dynamically generated DTR is missing the `table` attribute indicating which DynamicTable it points to, so it needs to be added after generation. If so, this should be documented. For now, documentation in the changelog would do, and we can add this as a TODO
item in the extensions documentation.

tests/unit/common/test_generate_table.py
```diff
@@ -0,0 +1,181 @@
...
+        test_table.add_row(my_col=3.0, indexed_col=[1.0, 3.0], optional_col2=.5)
+        test_table.add_row(my_col=4.0, indexed_col=[2.0, 4.0], optional_col2=.5)
+
+        test_dtr_table = self.TestDTRTable(name='test_dtr_table', description='my table')
+
+        test_dtr_table.add_row(ref_col=0, indexed_ref_col=[0, 1])
+        test_dtr_table.add_row(ref_col=0, indexed_ref_col=[0, 1])
+
+        np.testing.assert_array_equal(test_dtr_table['indexed_ref_col'].target.data, [0, 1, 0, 1])
+        np.testing.assert_array_equal(test_dtr_table['ref_col'].data, [0, 0])
```
rly, line 198: The DynamicTable code is so large and complex that I worry that something would not be set correctly on write or read. Can you add a roundtrip test?

tests/unit/common/test_generate_table.py
bendichter: @rly The thing is you can't do that before adding a row. I can put those lines at the end though:

```python
        test_dtr_table = self.TestDTRTable(name='test_dtr_table', description='my table')

        test_dtr_table.add_row(ref_col=0, indexed_ref_col=[0, 1])
        test_dtr_table.add_row(ref_col=0, indexed_ref_col=[0, 1])

        test_dtr_table['ref_col'].table = test_table
        test_dtr_table['indexed_ref_col'].target.table = test_table
```

is that better?

bendichter: Ideally I'd like to specify what the target tables are in the init. Something like:

```python
test_dtr_table = self.TestDTRTable(name='test_dtr_table', description='my table',
                                   target_tables={'ref_col': test_table, 'indexed_ref_col': test_table})
```

tests/unit/common/test_generate_table.py
bendichter: ok

bendichter: @rly actually can you do that part? I have figure out how to get the H5RoundTripMixin to work

rly: Sure, happy to do this.

tests/unit/common/test_generate_table.py
rly: Yes, that would be ideal and we can do that. I'll give that a try.

rly:
@bendichter I added a roundtrip test but left off testing validation of the file because I could not figure out why the test namespace did not contain its included namespaces, an issue that is not related to this PR.

tests/unit/common/test_generate_table.py
rly: Implementing the ideal case proved harder than I thought and will require some more work

bendichter (dismissed):
Looks great!

tests/unit/common/test_generate_table.py
bendichter: OK, let's hold off on this feature for now so we can get this through. What we have now works quite well for reading. Even though writing could be improves, it's still much better than what we have now.

tests/unit/common/test_generate_table.py
```diff
@@ -159,18 +185,48 @@ def test_dynamic_table_region(self):
...
 
+    def test_dynamic_table_region_optional(self):
+        test_table = self.TestTable(name='test_table', description='my test table')
+        test_table.add_row(my_col=3.0, indexed_col=[1.0, 3.0], optional_col2=.5)
+        test_table.add_row(my_col=4.0, indexed_col=[2.0, 4.0], optional_col2=.5)
+
+        test_dtr_table = self.TestDTRTable(name='test_dtr_table', description='my table',
+                                           target_tables={'optional_ref_col': test_table,
+                                                          'optional_indexed_ref_col': test_table})
+        self.assertIs(test_dtr_table['optional_ref_col'].table, test_table)
```
bendichter, line 208: @rly Could you try adding a row to `test_dtr_table`? I think it will fail because we need to make sure that unindexed `DynamicTableRegion` dtypes get mapped to `int` dtypes for the corresponding docval arg

bendichter: oh, you did, on line 191 and 192. OK, yeah this looks great!

rly: I added another test there just in case

bendichter (dismissed):

bendichter (approved):

PR #537, opened 2021-03-16
add EnumData API code

ajtritt:
## Motivation

Provides API functionality for the EnumData type. EnumData behaves like VocabData, except the elements/vocabulary is stored as a dataset rather than an attribute. This gets around size limits of attributes imposed by HDF5 size limits on file headers. 

**EnumData is experiment**

Corresponding schema change here: https://github.com/hdmf-dev/hdmf-common-schema/pull/51

## Checklist


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -918,26 +918,30 @@ def __resolve_dtype_helper__(cls, dtype):
...
+                    self.logger.debug("Setting %s '%s' attribute '%s' to %s"
+                                      % (obj.__class__.__name__, obj.name, key, value.__class__.__name__))
+                    obj.attrs[key] = value
+                elif isinstance(value, (Container, Builder, ReferenceBuilder)):           # a reference
+                    self.__queue_ref(self._make_attr_ref_filler(obj, key, value))
+                else:
+                    self.logger.debug("Setting %s '%s' attribute '%s' to %s"
+                                      % (obj.__class__.__name__, obj.name, key, value.__class__.__name__))
+                    obj.attrs[key] = value                   # a regular scalar
+            except Exception as e:
```
rly, line 942: Why was this try/except added?

ajtritt: To produce a better error message when it fails. It's not expected as a result of any common misuse, but just in case. It happened to me when I was working on this feature, and it was the only way I could figure out what was going on.

tests/unit/common/test_table.py
```diff
@@ -912,7 +912,8 @@ class SubTable(DynamicTable):
         {'name': 'col6', 'description': 'optional region', 'table': True},
         {'name': 'col7', 'description': 'required, indexed region', 'required': True, 'index': True, 'table': True},
         {'name': 'col8', 'description': 'optional, indexed region', 'index': True, 'table': True},
-        {'name': 'col10', 'description': 'required, indexed vocab column', 'index': True, 'class': VocabData},
+        {'name': 'col10', 'description': 'required, indexed enum column', 'index': True, 'class': EnumData},
+        {'name': 'col11', 'description': 'required, enumerable column', 'enum': True, 'index': True},
```
rly: What is the difference (if any) between these two lines?

tests/unit/common/test_table.py
```diff
@@ -996,8 +997,8 @@ def test_add_opt_column(self):
         table.add_column(name='col8', description='column #8', index=True, table=True)
         self.assertEqual(table.col8.description, 'column #8')
 
-        table.add_column(name='col10', description='column #10', index=True, col_cls=VocabData)
-        self.assertIsInstance(table.col10, VocabData)
+        table.add_column(name='col11', description='column #11', enum=True)
+        self.assertIsInstance(table.col11, EnumData)
```
rly: FYI I think this should raise a warning because `add_column` is missing the index argument whereas the configuration on line 916 has the index argument. The warning is tested on line 1039 already. 
```suggestion
        table.add_column(name='col11', description='column #11', enum=True, index=True)
        self.assertIsInstance(table.col11, EnumData)
```

tests/unit/common/test_table.py
```diff
@@ -1052,14 +1053,14 @@ def test_add_opt_column_mismatched_index_data(self):
...
+        msg = ("Column 'col10' is predefined in SubTable with class=<class 'hdmf.common.table.EnumData'> "
                "which does not match the entered col_cls "
                "argument. The predefined class spec will be ignored. "
                "Please ensure the new column complies with the spec. "
                "This will raise an error in a future version of HDMF.")
         with self.assertWarnsWith(UserWarning, msg):
-            table.add_column(name='col10', description='column #10', index=True)
-        self.assertEqual(table.col10.description, 'column #10')
+            table.add_column(name='col10', description='column #11', index=True)
+        self.assertEqual(table.col10.description, 'column #11')
```
rly:
```suggestion
            table.add_column(name='col10', description='column #10', index=True)
        self.assertEqual(table.col10.description, 'column #10')
```
This change doesn't seem necessary.

tests/unit/common/test_table.py
ajtritt: The first should test passing in the class directly. The second should test just doing it with the keyword.

ajtritt:
@rly just a heads up, the schema now points to tip of the schema master branch.

tests/unit/common/test_table.py
```diff
@@ -1401,6 +1402,18 @@ def setUpContainer(self):
...
 
+class TestDynamicTableAddEnumRoundTrip(H5RoundTripMixin, TestCase):
+
+    def setUpContainer(self):
+        table = DynamicTable('table0', 'an example table')
+        table.add_column('bar', 'an enumerable column', enum=True)
+        table.add_row(bar='a')
+        table.add_row(bar='b')
+        table.add_row(bar='a')
+        table.add_row(bar='c')
```
rly, line 1456: Can you make a separate non-round-trip test with this code and test for the expected result? And do the same for an indexed enum column? What is the expected result? I ran:
```
>>> table = DynamicTable('table0', 'an example table')
>>> table.add_column('foo', 'an indexed enumerable column', enum=True, index=True)
>>> table.add_row(foo=['a', 'b'])
>>> table.add_row(foo=['a', 'b'])
>>> table.add_row(foo=['a', 'b', 'c'])
>>> table.foo
<hdmf.common.table.EnumData object at 0x0000016EA474E708>
>>> table.foo.data
['a', 'b', 'a', 'b', 'a', 'b', 'c']
>>> table.foo.elements
<hdmf.common.table.VectorData object at 0x0000016EA474E948>
>>> table.foo.elements.data
[]
>>> table['foo']
<hdmf.common.table.VectorIndex object at 0x0000016EA474EA48>
>>> table['foo'].data
[2, 4, 7]
```
which seems a little counter intuitive. I would think either each element of the row should be enumerated, or the array for the row should be enumerated, but here, nothing is enumerated.

This can be done in a separate PR and since it is experimental, it is not critical to get it in the next release.

rly (dismissed):
Aside from the unclear behavior for indexed enum columns, which can be handled in a separate PR, this looks good to me.

src/hdmf/data_utils.py
```diff
@@ -10,11 +10,13 @@
 
 
 def append_data(data, arg):
+    if isinstance(data, DataIO):
+        pass
```
rly:
```suggestion
```

tests/unit/common/test_table.py
```diff
@@ -946,8 +947,8 @@ def test_init(self):
         self.assertIsNone(table.col6)
         self.assertIsNone(table.col8)
         self.assertIsNone(table.col8_index)
-        self.assertIsNone(table.col10)
-        self.assertIsNone(table.col10_index)
+        self.assertIsNone(table.col11)
+        self.assertIsNone(table.col11_index)
```
rly, line 951: I'm confused -- columns 2, 4, 6, and 8 are optional so they are not generated by the class when there is no data. But columns 10 and 11 are required, so they should not be None, right?

tests/unit/common/test_table.py
```diff
@@ -996,8 +997,8 @@ def test_add_opt_column(self):
         table.add_column(name='col8', description='column #8', index=True, table=True)
         self.assertEqual(table.col8.description, 'column #8')
 
-        table.add_column(name='col10', description='column #10', index=True, col_cls=VocabData)
-        self.assertIsInstance(table.col10, VocabData)
+        table.add_column(name='col11', description='column #11', enum=True, index=True)
+        self.assertIsInstance(table.col11, EnumData)
```
rly, line 1004: Can we keep the previous code regarding col10 to test add_column with col_cls=EnumData?

```suggestion
        table.add_column(name='col10', description='column #10', index=True, col_cls=EnumData)
        self.assertIsInstance(table.col10, EnumData)

        table.add_column(name='col11', description='column #11', enum=True, index=True)
        self.assertIsInstance(table.col11, EnumData)
```

tests/unit/common/test_table.py
rly: Can you add non-round-trip tests for a non-indexed enum column and an indexed enum column? I'm also happy to do it.

tests/unit/common/test_table.py
ajtritt: 10 and 11 are not required. The description is misleading.

rly: Oops, you're right

tests/unit/common/test_table.py
ajtritt: I'm not sure what you're asking for... Does the TestCase starting at line 1153 work for the non-index enum column?

rly: Yeah, the TestCase at line 1153 works for the non-index enum column because `DynamicTable.add_row` just calls `EnumData.add_row`. So that's fine. My bad. The indexed enum column is more complicated though and should be tested separately.

rly (approved):

PR #551, opened 2021-04-05
Add/aligned dynamic table

oruebel:
## Motivation

Add support for new AlignedDynamicTable type originally developed as part of ndx-icephys-meta to hdmf.common

## Checklist

- [X] Did you update CHANGELOG.md with your changes?


src/hdmf/common/table.py
```diff
@@ -1267,3 +1267,434 @@ def add_row(self, **kwargs):
         if not index:
             val = self.__add_term(val)
         super().append(val)
+
```
ajtritt: At the very least, I think this should be moved into it's own file. This feature alone adds ~33% more lines of code.

ajtritt: What does this change about the user experience? From my perspective, a group with a three DynamicTables would suffice for the use you outlined [here](https://github.com/hdmf-dev/hdmf-common-schema/pull/57#discussion_r608082072). That means, the user would access categories like so:

```
dt_group.stimulus['col1']
```

Instead, with this this addition, users access categories like this:

```
hier_dt.get_category('stimulus')['col1']
```

Am I missing something, or is there some other reason for this?

oruebel: Moving it to a separate module is fine with me. What should the module be called?

> What does this change about the user experience?

If you only want to interact with one-category at a time just having a collection of tables is fine, but with this you can do both, i.e., treat it as one large coherent table as well as interact with each category independently. In practice users need to do both, i.e., extract all metadata about a recording at once (i.e., extract full rows across all tables at once) as well as just look at a particular sub-category (e.g., settings of electrodes or stimuli). Inheriting from DynamicTable further has the advantage that we can more easily build of all the features.

ajtritt: Gotcha. That makes sense. I have no preference on module name.... aligned_table.py?

oruebel: I just committed the changes to move it to its on module including separate io and test modules.

oruebel:
@rly @ajtritt this PR is ready for review

ajtritt (dismissed):

oruebel:
@ajtritt can you re-approve. There was a merge conflict on the CHANGELOG that blocked merging

ajtritt (approved):

rly (approved):

Issue #554, opened 2021-04-06
Autogmatic class generation for AlignedDynamicTable fails


oruebel:
## Description

This issue has come up as part of https://github.com/hdmf-dev/hdmf/pull/551 to integrated ``AlignedDynamicTable`` which is a ``DynamicTable`` that itself contains other ``DynamicTable`` objects. Due https://github.com/hdmf-dev/hdmf/issues/553 integrating ``AlignedDynamicTable`` triggered the automatic class generation for that type. The automatic class generation for ``AlignedDynamicTable`` , however, fails on the current HDMF dev branch. It appears that the automatic class generation attempts to use both ``DynamicTable`` and ``MultiContainerInterface`` as a base type for ``AlignedDynamicTable`` which appear to result into a conflict with regard to method resolution order. Specifically the type of error I am seeing is:

```
ImportError: Failed to import test module: validator_tests.test_validate
Traceback (most recent call last):
  File "/Users/oruebel/anaconda3/envs/icephys-port/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/Users/oruebel/anaconda3/envs/icephys-port/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/Users/oruebel/Devel/nwb/icephys_extensions/port_to_nwb/hdmf/tests/unit/validator_tests/test_validate.py", line 10, in <module>
    from hdmf.testing import TestCase
  File "/Users/oruebel/Devel/nwb/icephys_extensions/port_to_nwb/hdmf/src/hdmf/testing/__init__.py", line 1, in <module>
    from .testcase import TestCase, H5RoundTripMixin  # noqa: F401
  File "/Users/oruebel/Devel/nwb/icephys_extensions/port_to_nwb/hdmf/src/hdmf/testing/testcase.py", line 11, in <module>
    from ..common import validate as common_validate, get_manager
  File "/Users/oruebel/Devel/nwb/icephys_extensions/port_to_nwb/hdmf/src/hdmf/common/__init__.py", line 109, in <module>
    load_namespaces(__resources['namespace_path'])
  File "/Users/oruebel/Devel/nwb/icephys_extensions/port_to_nwb/hdmf/src/hdmf/utils.py", line 584, in func_call
    return func(**pargs)
[...]

PR #555, opened 2021-04-07
do not autogenerate  classes when loading namespace

### LINKED ISSUE #553: Extra auto classgeneration on import of hdmf.common types
## Description

In the function ``TypeMap.get_container_cls`` function in ``hdmf.build.manager`` the internal helper function ``TypeMap.__get_container_cls`` call initially returns ``None`` for classes from ``hdmf.common``.  When importing a container class from ``hdmf.common`` this results in a sequence of calls where first Container classes are autogenerated and the later on the correct custom classes from ``hdmf.common`` are detected.  When I add a ``print(cls, namespace, data_type)`` in ``TypeMap.get_container_cls`` and import ``DynamicTable`` I get the following:

```
from hdmf.common.table import DynamicTable
None hdmf-common Data
None hdmf-common Container
None hdmf-common SimpleMultiContainer
None hdmf-common VectorData
None hdmf-common VectorIndex
None hdmf-common ElementIdentifiers
None hdmf-common DynamicTableRegion
None hdmf-common DynamicTable
None hdmf-common AlignedDynamicTable
None hdmf-common CSRMatrix
<class 'hdmf.common.table.DynamicTable'> hdmf-common DynamicTable
<class 'hdmf.common.table.VectorData'> hdmf-common VectorData
<class 'hdmf.common.table.VectorIndex'> hdmf-common VectorIndex
<class 'hdmf.common.table.ElementIdentifiers'> hdmf-common ElementIdentifiers
<class 'hdmf.common.table.DynamicTableRegion'> hdmf-common DynamicTableRegion
<class 'hdmf.common.table.EnumData'> hdmf-experimental EnumData
<class 'hdmf.common.sparse.CSRMatrix'> hdmf-common CSRMatrix
<class 'hdmf.common.resources.ExternalResources'> hdmf-experimental ExternalResources
<class 'hdmf.common.multi.SimpleMultiContainer'> hdmf-common SimpleMultiContainer
```

I.e., there is a first round where ``TypeMap.__get_container_cls`` returns ``None`` which in turn triggers the automatic class generation and then in a second round the correct classes are found. 

While the end-behavior for the user is correct but having classes autogenerated for all types seems like an unnecessary that could slow down import of HDMF. I found this behavior as part of PR https://github.com/hdmf-dev/hdmf/pull/551 to integrate ``AlignedDynamicTable`` , because automatic class generation currently fails for this type (see Issue https://github.com/hdmf-dev/hdmf/issues/554) and the behavior described in this issues triggers that behavior.

## Steps to Reproduce

```from hdmf.common.table import DynamicTable```

## Environment

    Python Executable: Installed HDMF from latest dev
    Python Version: Python  3.8
    Operating System:  macOS
    HDMF Version: dev

## Checklist

- [X] Have you ensured the bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?

### PR DESCRIPTION
**Pull request by ajtritt**
## Motivation

Fix #553 

## How to test the behavior?
Set breakpoints or print statements deep in TypeMap.get_container_cls. 

## Checklist

### DIFF
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -474,14 +474,14 @@ class TypeMap:
         for new_ns, ns_deps in deps.items():
             for src_ns, types in ns_deps.items():
                 for dt in types:
-                    container_cls = self.get_container_cls(src_ns, dt, autogen=False)
+                    container_cls = self.get_container_cls(dt, src_ns, autogen=False)
                     if container_cls is None:
                         container_cls = TypeSource(src_ns, dt)
                     self.register_container_type(new_ns, dt, container_cls)
         return deps
 
-    @docval({"name": "namespace", "type": str, "doc": "the namespace containing the data_type"},
-            {"name": "data_type", "type": str, "doc": "the data type to create a AbstractContainer class for"},
+    @docval({"name": "data_type", "type": str, "doc": "the data type to create a AbstractContainer class for"},
+            {"name": "namespace", "type": str, "doc": "the namespace containing the data_type", "default": None},
             {"name": "autogen", "type": bool, "doc": "autogenerate class if one does not exist", "default": True},
             returns='the class for the given namespace and data_type', rtype=type)
     def get_container_cls(self, **kwargs):
@@ -492,7 +492,7 @@ class TypeMap:
         namespace, data_type, autogen = getargs('namespace', 'data_type', 'autogen', kwargs)
 
         # namespace is unknown, so look it up
-        if namespace == 'UNKNOWN':
+        if namespace is None:
             for key, val in self.__container_types.items():
                 # NOTE that the type_name may appear in multiple namespaces based on how they were resolved
                 # but the same type_name should point to the same class
@@ -514,16 +514,16 @@ class TypeMap:
         """Ensure that classes for all types used by this type exist and generate them if not.
         """
         if spec.data_type_inc is not None:
-            self.get_container_cls(namespace, spec.data_type_inc)
+            self.get_container_cls(spec.data_type_inc, namespace)
         if isinstance(spec, GroupSpec):
             for child_spec in (spec.groups + spec.datasets):
                 if child_spec.data_type_inc is not None:
-                    self.get_container_cls(namespace, child_spec.data_type_inc)
+                    self.get_container_cls(child_spec.data_type_inc, namespace)
                 if child_spec.data_type_def is not None:
-                    self.get_container_cls(namespace, child_spec.data_type_def)
+                    self.get_container_cls(child_spec.data_type_def, namespace)
             for child_spec in spec.links:
                 if child_spec.target_type is not None:
-                    self.get_container_cls(namespace, child_spec.target_type)
+                    self.get_container_cls(child_spec.target_type, namespace)
 
     def __get_parent_cls(self, namespace, data_type, spec):
         dt_hier = self.__ns_catalog.get_hierarchy(namespace, data_type)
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by rly** [comment]
Looks good, though one test is failing. IIRC that test was the following:
'test-core' namespace contains type Bar
'ndx-qux' namespace contains types Qux, Spam
'ndx-test' namespace contains type Baz which extends Bar and includes types Qux and Bar

I think I figured out the core issue. 

1) some dependencies (data_type_inc and links) were not being resolved. That can be resolved with the following new `get_container_cls` and renamed `__resolve_child_container_classes`:
```python
    def get_container_cls(self, **kwargs):
        """Get the container class from data type specification.
        If no class has been associated with the ``data_type`` from ``namespace``, a class will be dynamically
        created and returned.
        """
        namespace, data_type, autogen = getargs('namespace', 'data_type', 'autogen', kwargs)
        cls = self.__get_container_cls(namespace, data_type)
        if cls is None and autogen:  # dynamically generate a class
            spec = self.__ns_catalog.get_spec(namespace, data_type)
            self.__resolve_dependent_container_classes(spec, namespace)
            parent_cls = self.__get_parent_cls(namespace, data_type, spec)
            attr_names = self.__default_mapper_cls.get_attr_names(spec)
            cls = self.__class_generator.generate_class(data_type, spec, parent_cls, attr_names, self)
            self.register_container_type(namespace, data_type, cls)
        return cls

    def __resolve_dependent_container_classes(self, spec, namespace):
        """Trigger class generation and registration for dependent data types."""
        if spec.data_type_inc is not None:
            self.get_container_cls(namespace, spec.data_type_inc)
        if isinstance(spec, GroupSpec):
            for child_spec in (spec.groups + spec.datasets):
                if child_spec.data_type_inc is not None:
                    self.get_container_cls(namespace, child_spec.data_type_inc)
                elif child_spec.data_type_def is not None:
                    self.get_container_cls(namespace, child_spec.data_type_def)
            for child_spec in spec.links:
                if child_spec.target_type is not None:
                    self.get_container_cls(namespace, child_spec.target_type)
```

2) The test generates classes from different namespaces and expects the TypeMap to resolve between them, BUT when a data type from one namespace references types from another namespace, those type classes should be included/added in the first namespace and it is those generated classes with the first namespace that will be referred to in the docval and other properties of the original generated class.

So the test can be modified to generate classes all from the ndx-test namespace
```python
baz_cls = self.type_map.get_container_cls('ndx-test', 'Baz')
bar_cls = self.type_map.get_container_cls('ndx-test', 'Bar')
qux_cls = self.type_map.get_container_cls('ndx-test', 'Qux')
```

This cross-namespace strategy might not work fully though. 
Let's say:
Namespace N1 contains group type A
Namespace N2 contains group type B extends A. And so N2 will also include its own version of type A.
Namespace N3 contains group type C which includes subgroup type A. And so N3 will also include its own version of type A.

Type B inherits from type A built with namespace N2
Type C allows inclusion of classes of type A built with namespace N3. So the constructor for type C will not allow Type B to be passed in. 

I'm not sure what the solution to this is yet, but this situation is pretty rare. It has happened already with the Allen Institute so we should test that before releasing this (but we can merge this).

rly:
This change will fix #554 as well.

src/hdmf/build/manager.py
```diff
@@ -474,23 +474,25 @@ def load_namespaces(self, **kwargs):
         for new_ns, ns_deps in deps.items():
             for src_ns, types in ns_deps.items():
                 for dt in types:
-                    container_cls = self.get_container_cls(src_ns, dt)
+                    container_cls = self.get_container_cls(src_ns, dt, autogen=False)
```
rly: I think the autogen flag is not necessary. You could just do:
```suggestion
                    container_cls = self.__get_container_cls(src_ns, dt)
```

rly: Never mind - this is useful since I now propose to change `__get_conainer_cls`

rly:
Never mind my comments above. I just realized the purpose of TypeSource and made a proposed change in #557 to fix the failing test.

oruebel (dismissed):
Looks good to me. I'm wondering whether this will also help speed-up importing HDMF.

oruebel (approved):

ajtritt:
@rly @oruebel This change will require a change in `pynwb` to use the new form of `TypeMap.get_container_cls`

PR #557, opened 2021-04-08
Improve class generation across namespace dependencies

rly:
## Motivation

Proposed modification of #555 to handle the failing test and improve cross-namespace class generation. Merging would be into the enh/get_class_fix branch used by #555.

Scenario:
Namespace N1 contains group type A. 
Namespace N2 contains group type B extends A. If class A for type A in namespace N1 does not exist, then the TypeMap now generates this class A, maps type A in namespace N1 to this class A, and maps type A in namespace N2 to *the same class A*.
Namespace N3 contains group type C which includes subgroup type A. The TypeMap now uses the same class A for all three namespaces.

The order of calling get_class on classes and their dependencies thus does not matter.

Note: If get_class is called such that a class is dynamically generated for class A, and later a custom API class is mapped to class A, then the two class As will not be the same, which could lead to unexpected behavior. This is not new behavior -- I just mention that this could happen.

## Checklist

### DIFF
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -542,10 +542,10 @@ class TypeMap:
             return None
         ret = self.__container_types[namespace][data_type]
         if isinstance(ret, TypeSource):  # data_type is a dependency from ret.namespace
-            klass = self.get_container_cls(ret.namespace, ret.data_type)  # get class / generate class
+            cls = self.get_container_cls(ret.namespace, ret.data_type)  # get class / generate class
             # register the same class into this namespace (replaces TypeSource)
-            self.register_container_type(namespace, data_type, klass)
-            ret = klass
+            self.register_container_type(namespace, data_type, cls)
+            ret = cls
         return ret
 
     @docval({'name': 'obj', 'type': (GroupBuilder, DatasetBuilder, LinkBuilder, GroupSpec, DatasetSpec),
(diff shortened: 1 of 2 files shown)


oruebel (approved):
Looks good to me.

src/hdmf/build/manager.py
```diff
@@ -529,15 +535,17 @@ def __get_parent_cls(self, namespace, data_type, spec):
...
         ret = self.__container_types[namespace][data_type]
-        if isinstance(ret, TypeSource):
-            ret = self.__get_container_cls(ret.namespace, ret.data_type)
-            if ret is not None:
-                self.register_container_type(namespace, data_type, ret)
+        if isinstance(ret, TypeSource):  # data_type is a dependency from ret.namespace
+            klass = self.get_container_cls(ret.namespace, ret.data_type)  # get class / generate class
+            # register the same class into this namespace (replaces TypeSource)
+            self.register_container_type(namespace, data_type, klass)
+            ret = klass
```
ajtritt:
```suggestion
            cls = self.get_container_cls(ret.namespace, ret.data_type)  # get class / generate class
            # register the same class into this namespace (replaces TypeSource)
            self.register_container_type(namespace, data_type, cls)
            ret = cls
```

tests/unit/build_tests/test_classgenerator.py
```diff
@@ -387,12 +387,17 @@ def test_get_class_include_from_separate_ns(self):
             type_map=self.type_map
         )
 
+        # order of generation should not matter as long as dependencies are listed correctly in namespaces
```
ajtritt, line 390: When I was working on this, order was relevant, so I think we should test that.

ajtritt (changes requested):

tests/unit/build_tests/test_classgenerator.py
ajtritt: I will add that on `enh/get_class_fix`

PR #558, opened 2021-04-08
Hack er

mavaylon1:
## Motivation

Expand ExternalResources methods and fix bugs in existing methods. 

## How to test the behavior?
```

Run the unit test for resources.py
```

## Checklist


docs/gallery/External_Resources.py
```diff
@@ -14,7 +14,38 @@
...
-# from the resources.
+# from the resources. A typical use case for external resources is to link data
+# stored in datasets or attributes to ontologies. For example, you may have a
+# dataset `country` storing locations. Using
+# :py:class:`~hdmf.common.resources.ExternalResources` allows us to link the
+# country names stored in the dataset to an ontology of all countries, enabling
+# more rigid standardization of the data and facilitating data query and
+# introspection.
+#
+# From a users perspective, one can think of the ``ExternalResources`` as a
```
rly:
```suggestion
# From a user's perspective, one can think of the ``ExternalResources`` as a
```

docs/gallery/External_Resources.py
```diff
@@ -14,7 +14,38 @@
...
+# simple table, in which each row associates a particular ``key`` stored in a
+# particular ``object`` (i.e., Attribute or Dataset in a file) with a particular
+# ``entity`` (e.g., a term)  of an online ``resource`` (e.g., an ontology).
+# That is, ``(object, key)`` refer to parts inside a file and ``(resource, entity)``
+# refer to an external resource outside of the file, and ``ExternalResources``
+# allows us to link the two. To reduce data redundancy and improve data integrity
```
rly:
```suggestion
# allows us to link the two. To reduce data redundancy and improve data integrity,
```

docs/gallery/External_Resources.py
```diff
@@ -14,7 +14,38 @@
...
+# * :py:class:`~hdmf.common.resources.EntityTable`  where each row describes an
+#   :py:class:`~hdmf.common.resources.Entity`
+# * :py:class:`~hdmf.common.resources.ObjectTable` where each row descibes an
+#   :py:class:`~hdmf.common.resources.Object`
+# * :py:class:`~hdmf.common.resources.ObjectKeyTable` where each row describes an
+#  :py:class:`~hdmf.common.resources.ObjectKey` pair identifying which keys
+#   are used by which objects.
+#
+# The :py:class:`~hdmf.common.resources.ExternalResources` class then provides
+# convenienve functions to simply interation with these tables, allowing users
```
rly:
```suggestion
# convenience functions to simplify interaction with these tables, allowing users
```

docs/gallery/External_Resources.py
```diff
@@ -27,9 +58,11 @@
 ###############################################################################
 # Using the add_ref method
 # ------------------------------------------------------
-# add_ref takes inputs of new data via keyword arguments
-# and adds to ExternalResources.
-# You can think of add_ref as a wrapper function.
+# :py:func:`~hdmf.common.resources.ExternalResources.add_ref`
+# is a wrapper function provided by the ExternalResources class, that
```
rly:
```suggestion
# is a wrapper function provided by the ``ExternalResources`` class, that
```

docs/gallery/External_Resources.py
```diff
@@ -27,9 +58,11 @@
...
+# simplifies adding data. Using add_ref allows us to treat new entries similar
```
rly:
```suggestion
# simplifies adding data. Using ``add_ref`` allows us to treat new entries similar
```

docs/gallery/External_Resources.py
```diff
@@ -27,9 +58,11 @@
...
+# to adding a new row to a flat table, with add_ref taking care of populating
```
rly:
```suggestion
# to adding a new row to a flat table, with ``add_ref`` taking care of populating
```

docs/gallery/External_Resources.py
```diff
@@ -63,14 +96,12 @@
...
-# you would like to associate with resources.
-# To disambiguate between these different fields, you can set the 'field' keyword
-# argument corresponding to the name of the dataset or attribute containing the
-# values being associated with a resource. For example:
+# In the above example, ``field`` was empty because the data of the ``container``
+# data object itself was being associated with a resource. However, rather than
+# the ``container`` data object itself, you may instead want to associated external
+# resources with datasets or attributes contained in the data object.
+# that you would like to associate with resources. To disambiguate between
+# these different fields, you can set the 'field' keyword.
```
rly:
```suggestion
# In the above example, the ``field`` keyword argument was empty because the data
# of the :py:class:`~hdmf.container.Data` object passed in for the ``container``
# argument was being associated with a resource. However, you may want to associate
# an attribute of a :py:class:`~hdmf.container.Data` object with a resource or
# a dataset or attribute of a :py:class:`~hdmf.container.Container` object with
# a resource. To disambiguate between these different fields, you can set the
# 'field' keyword.
```

docs/gallery/External_Resources.py
```diff
@@ -50,21 +73,40 @@
            resource_uri='https://www.ncbi.nlm.nih.gov/taxonomy', entity_id='NCBI:txid10090',
            entity_uri='https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=10090')
 
+###############################################################################
+# Using the add_ref method with get_resource
+# ------------------------------------------------------
+# We can use the get_resource method to reference an existing resource within
+# ExternalResources when adding data with add_ref.
```
rly: I suggest starting with the use case, then explaining how this function solves the use case.

```suggestion
# When adding references to resources, you may want to refer to multiple entities
# within the same resource. Resource names are unique, so if you call ``add_ref``
# with the name of an existing resource, then that resource will be reused. You 
# can also use the :py:func:`~hdmf.common.resources.ExternalResources.get_resource`
# method to get the ``Resource`` object and pass that in to ``add_ref`` to 
# reuse an existing resource.
```

src/hdmf/common/resources.py
```diff
@@ -291,14 +308,29 @@ def add_ref(self, **kwargs):
         It is possible to use the same name of the key to refer to different resources
         so long as the name of the key is not used within the same object and field.
         This method does not support such functionality by default. The different
-        keys must be added separately using *add_key* and passed to the *key* argument in separate calls of this method.
+        keys must be added separately using _add_key and passed to the *key* argument in separate calls of this method.
```
```suggestion
        keys must be added separately using _add_key and passed to the *key* argument 
        in separate calls of this method. If a resource with the same name already
        exists, then it will be used and the resource_uri will be ignored.
```

docs/gallery/External_Resources.py
```diff
@@ -85,112 +127,24 @@
...
 # ------------------------------------------------------
-# This method creates new resource objects and adds them to the ResourceTable.
-# The ResourceTable holds the name and uri for the resource.
-# For example, you could add NCBI Taxonomy with its uri.
-
-er = ExternalResources(name='new_example')
-
-
-er.add_resource(resource='NCBI_Taxonomy', uri='https://www.ncbi.nlm.nih.gov/taxonomy')
+# This method will return a key_object. In the current version of ExternalResources
```
rly:
```suggestion
# This method will return a ``Key`` object. In the current version of ``ExternalResources``,
```

docs/gallery/External_Resources.py
```diff
@@ -85,112 +127,24 @@
...
+# duplicate keys are allowed; however, each key needs a unique linking Object.
+# i.e (container, field)
```
rly:
```suggestion
# In other words, each combination of (container, field, key) can exist only once in
# ``ExternalResources``.
```

docs/gallery/External_Resources.py
```diff
@@ -85,112 +127,24 @@
...
 
 ###############################################################################
-# Using the add_entity method
+# Using the add_ref method with a key_object
 # ------------------------------------------------------
-# This method creates a new entity object and adds it to the EntityTable.
-# Keeping with the NCBI Taxonomy example, the entity would be a specfic item
-# or "search" in NCBI Taxonomy, whereas the resource is NCBI Taxonomy itself.
+# Sometimes you want to reference a specific key that already exists when adding
+# new ontology data into ExternalResources.
```
rly:
```suggestion
# new ontology data into ``ExternalResources``.
```

rly (approved):

oruebel (approved):
Looks good to me

PR #566, opened 2021-04-12
Improve DynamicTable equals

rly:
## Motivation

When iterating through the `Container` objects in a `MultiContainerInterface` or testing whether a Container is in a `MultiContainerInterface`, the equality method `__eq__` is useful (e.g., `if container in proc_module.data_interfaces`). 

`DynamicTable.__eq__` assumes that the other object is a `DynamicTable` and calls `.to_dataframe()` on it. But if the other object is not a `DynamicTable`, an error is raised, saying the method doesn't exist. 

`DynamicTable.__eq__` should first check that the other object is a `DynamicTable`.

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -593,6 +593,8 @@ class DynamicTable(Container):
 
         :return: Bool indicating whether the two DynamicTables contain the same data
         """
+        if other is self:
+            return True
         if not isinstance(other, DynamicTable):
             return False
         if self.name != other.name or self.description != other.description:
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -583,18 +583,20 @@ class DynamicTable(Container):
                 c.add_row(data[colname])
 
     def __eq__(self, other):
-        """
-        Compare if the two DynamicTables contain the same data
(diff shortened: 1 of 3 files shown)


ajtritt (approved):

src/hdmf/common/table.py
```diff
@@ -583,18 +583,20 @@ def add_row(self, **kwargs):
...
+        description. Then, this table and the other table are converted to pandas
+        dataframes and the equality of the two tables is returned.
 
         :param other: DynamicTable to compare to
 
-        :raises: An error will be raised with to_dataframe is not defined or other
-
         :return: Bool indicating whether the two DynamicTables contain the same data
         """
+        if not isinstance(other, DynamicTable):
```
ajtritt, line 598:
```suggestion
        if other is self:
            return True
        if not isinstance(other, DynamicTable):
```

ajtritt: Avoid conversion to dataframes when possible.

ajtritt (changes requested):

src/hdmf/common/table.py
rly: Very sensible

ajtritt (approved):

PR #567, opened 2021-04-13
Do not add MCI as a base if base is already subclass of MCI

rly:
## Motivation

When generating a class from a spec that is a `MultiContainerInterface` (because it contains quantity * or + of a group or dataset), `MultiContainerInterface` should be added as a base class. However, sometimes the other base class (X) is already a subclass of `MultiContainerInterface`. This results in:

```TypeError: Cannot create a consistent method resolution order (MRO) for bases MultiContainerInterface, X```

## How to test the behavior?
```
```

## Checklist


oruebel (dismissed):

rly:
Thanks for the review @oruebel . I realized that the line `if MultiContainerInterface in b.__mro__` was a bit odd and that `if issubclass(b, MultiContainerInterface)` would be clearer and possibly more efficient, so I updated that line.

oruebel (approved):

ajtritt (approved):

PR #583, opened 2021-04-22
Revert "Validation finds unexpected elements (#542)"

rly:
#542 introduced warnings for extra fields but treats attributes added to a spec as extra fields. This is problematic for PyNWB because in the NWB schema, the `Units` table contains the column with spec `name: spike_times; neurodata_type_inc: VectorData; attributes: [new attribute for resolution]`. The resolution attribute is flagged as an extra field. Upon further discussion at https://github.com/hdmf-dev/hdmf-schema-language/issues/13 , we decided to continue allowing this case (this use case may also appear in extensions).

This means that #542 as is will break validation of NWB files with the Units/spike_times.resolution field (and files with extensions that have this case). @dsleiter has graciously agreed to work on resolving this issue in a separate PR. In the meantime, in order to prepare for the upcoming 2.5 minor release, this PR reverts commit 67cffecad50fbc31f8406367f9442b6e3bd87a4a (#542).

We will reintroduce this commit in a separate PR post-release.


rly:
Replaced by #583.

PR #550, opened 2021-03-30
Prepare for version 2.5.0 release

rly:
- Updated changelog
- Updated requirements
- Updated CI
- Add additional cache spec test and clean up some other tests

~~Some critical bugs were identified with the experimental namespace and class generation. Ideally, these will be resolved before this release.~~

CI notes:
- h5py 2.10 did not release wheels for Python 3.9, so it has to be built from source using the HDF5 library. This is doable in Linux, but I cannot figure it out for Windows and MacOS on Azure.
- On Python 3.9, we have to build h5py 2.10. If numpy 1.20 is installed, then the build of h5py 2.10 appears to be incompatible with numpy 1.19: `ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject` . Upgrading to numpy 1.20 solves the issue but numpy 1.20 does not support Python 3.6, which we want to continue to support. For this reason, the python3.9 job fails. So users should note that when using Python 3.9 (which is not fully supported by HDMF), they should install a version of numpy first, and then build the wheels for h5py 2.10. 
- The pynwb-dev-python38 job fails because of a single test in pynwb that does not expect the hdmf-experimental namespace.

PR #503, opened 2020-12-23
Fix classification of attributes as new/overridden and issue with custom spec keys

rly:
## Motivation

Fix #501, address part of #502, and fix another related bug where cached HDMF schema loaded from a NWB file were not being resolved correctly.

When extracting HDMF out of PyNWB, a few keys were renamed to be more general, e.g., 'neurodata_type_def' -> 'data_type_def', and subclasses of `GroupSpec`, `DatasetSpec`, and `SpecNamespace` in HDMF were created in PyNWB. These NWB-specific subclasses are used when reading the hdmf-common schema cached within a file, but the cached schema uses 'data_type_def' instead of 'neurodata_type_def'. This prevented hdmf-common specs from being resolved correctly so that attributes, datasets, groups, and links defined by a base type within hdmf-common were in some  cases not inherited by a type that extends the base type.

I also refactored parts of spec.py, added comments, and added additional tests to push code coverage of spec.py from 79% to 83% and namespace.py from 79% to 90%.

## Checklist

- [x] Did you update CHANGELOG.md with your changes?
- [x] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [x] Have you ensured the PR clearly describes the problem and the solution?
- [x] Is your contribution compliant with our coding style? This can be checked running `flake8` from the source directory.
- [x] Have you checked to ensure that there aren't other open [Pull Requests](https://github.com/hdmf-dev/hdmf/pulls) for the same change?
- [x] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number? By including "Fix #XXX" you allow GitHub to close issue #XXX when the PR is merged.

PR #586, opened 2021-04-22
Fix wrong data_type_inc key being used in type resolution

rly:
## Motivation

PyNWB loads the spec for `AlignedDynamicTable` and resolves the included datasets from `DynamicTable`, namely:
```
(
    {'shape': [None], 'dims': ['num_rows'], 'dtype': 'int', 'name': 'id', 'doc': 'Array of unique identifiers for the rows of this dynamic table.', 'neurodata_type_inc': 'ElementIdentifiers'}, 
    {'doc': 'Vector columns, including index columns, of this dynamic table.', 'quantity': '*', 'neurodata_type_inc': 'VectorData'}
)
```

However, when the cached spec is read from a file, the included datasets are not resolved. `AlignedDynamicTable` has no datasets.

This only happens when PyNWB loads hdmf-common. In HDMF, both the original spec and the cached spec contain resolved datasets. 

The root error seems to be that the wrong data type inc key is being updated when the spec is resolved. This PR fixes that and cleans up some of the surrounding code.

See https://github.com/hdmf-dev/hdmf/pull/550#issuecomment-824472387 for error message.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
   `MultiContainerInterface`. @rly (#567)
 - Fix `make clean` command for docs to clean up sphinx-gallery tutorial files. @oruebel (#571)
 - Make sure we cannot set ``AlignedDynamicTable`` as a category on an ``AlignedDynamicTable``. @oruebel (#571)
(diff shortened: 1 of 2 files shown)


rly:
Closed in favor of #503 which is more comprehensive.

PR #596, opened 2021-05-03
Require setuptools

hrnciar:
## Motivation

pkg_resources needs setuptools as a runtime dependency. Since it wasn't specified so far, there was a possibility of `ModuleNotFoundError: No module named 'pkg_resources'` in case of setuptools not being present. There is a way how to install packages without setuptools or setuptools can be uninstalled after hdmf package was installed. My PR prevents this error.

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## Upcoming
+
+### Bug fix
+- Add explicit `setuptools` requirement. @hrnciar (#596)
+
 ## HDMF 2.5.1 (April 23, 2021)
 
 ### Bug fix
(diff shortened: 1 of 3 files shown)


rly:
This looks fine to me. I think it would be rare that `setuptools` would be uninstalled, but it would be good to specify this just in case.

Also, the use of `pkg_resources.resource_filename` is now [discouraged](https://setuptools.readthedocs.io/en/latest/pkg_resources.html) in favor or `importlib.resources` (python >=3.7) or its backport `importlib_resources`. However, `importlib.resources` does not support finding a resource in a non-package subdirectory such as what we do with `hdmf-common-schema/common/namespace.yaml`. `importlib_resources` recently added support for this: https://gitlab.com/python-devs/importlib_resources/-/issues/58 , but it does not seem necessary to change our code and add a new dependency right now.

rly (approved):

rly:
Thanks @hrnciar !

PR #613, opened 2021-05-17
Load dependent/included types in namespace/extension

### LINKED ISSUE #356: Cannot construct some extension types if type ancestry is not included in namespace
## Description

I am deep in the world of extensions for several groups. 

If an extension:
1) defines a type that does not directly extend `NWBDataInterface` (e.g., it extends `ElectricalSeries`)
2) the namespace includes the extended type (e.g., `ElectricalSeries`)
3) the namespace does not include the ancestry of the extended type all the way back to `NWBDataInterface`,

then when I add an instance of that type to `nwbfile.acquisition` (which allows `NWBDataInterface` objects), write the NWB file, and then read the file, the build process does not construct the new type. 

For example, if I define a new type `TetrodeSeries` that extends `ElectricalSeries` and the namespace file includes only the `ElectricalSeries` type from pynwb core, then the `TetrodeSeries` objects in `nwbfile.acquisition` will not be constructed from file.

However, if the namespace DOES include the ancestry of the extended type all the way back to `NWBDataInterface`, e.g., the extension namespace file includes the following:
```yaml
  schema:
  - namespace: core
    neurodata_types:
    - ElectricalSeries
    - TimeSeries
    - NWBDataInterface
```

then the build process works correctly.

I think including JUST the extended type, e.g., `ElectricalSeries`, should be sufficient to achieve the desired behavior, and this is a bug. Thoughts @ajtritt @oruebel ?

rly:
## Motivation

Fix #356 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 2.5.4 (May 17, 2021)
+
+### Bug fix
+- Fix issue where dependencies of included types were not being loaded in namespaces / extensions. @rly (#613)
+
 ## HDMF 2.5.3 (May 12, 2021)
 
 ### Bug fix
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -537,8 +537,7 @@ class TypeMap:
                 if spec.data_type_def is not None:
                     self.get_dt_container_cls(spec.data_type_def, namespace)
             elif isinstance(spec, LinkSpec):
-                if spec.target_type is not None:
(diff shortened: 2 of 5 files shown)


rly:
This is blocking the release of PyNWB 1.5.0 which uses nwb-schema 2.3.0 because the new schema adds in `ImageSeries` an optional link to `Device`, and extensions with data types that extend `ImageSeries` fail to load unless the extension includes `Device` from the NWB core schema. Example: aibs-stimulus-template extension.

ajtritt (approved):

PR #620, opened 2021-05-26
Drop python 3.6 support, add python 3.9 support

rly:
## Motivation

Python 3.6 end of life is 23 Dec 2021. The latest versions of h5py, numpy, pandas, scipy do not provide wheels for python 3.6. It seems that generally, if a package added 3.9 support, they dropped 3.6 support. This makes it difficult to test against. 

This PR will merge into the `enh/h5py3` branch to drop python 3.6 support and add python 3.9 support. 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,13 +38,6 @@ references:
         username: hdmf
         password: $DOCKERHUB_PASSWORD
 
-  py36: &py36
-    docker:
-    - image: circleci/python:3.6.13-buster
-      auth:
-        username: hdmf
-        password: $DOCKERHUB_PASSWORD
-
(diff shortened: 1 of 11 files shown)


rly:
The pynwb-dev-python39 test fails because Python 3.9 is not yet supported by PyNWB (but would be after this PR and a corresponding one in PyNWB).

rly:
I could not find an early release of ruamel.yaml in the 0.15.* range that worked for Python 3.7 and all OS's, for use as a minimum version requirement for ruamel.yaml, so I bumped up the minimum requirement to 0.16. If there is a need, we can explore what versions in the 0.15.* range would work.

ajtritt (approved):

PR #636, opened 2021-06-28
Retrieving Correct Fields for ExternalResources/Adjusting methods for Fields

mavaylon1:
## Motivation

We updated the structure of ExternalResources by creating a new column in the object table called "relative_path". This is the path to the attribute that is using an external reference. Field is still present, but the primary use is to add clarity to compound data types, where it denotes the field of that is using an external reference. With these changes are updates to tests and the demo.

## How to test the behavior?
1. Use the unit tests for resources.py

## Checklist


src/hdmf/common/Fields.ipynb
```diff
@@ -0,0 +1,618 @@
+{
```
rly: Was this intended to be included?

rly:
Please remove `test_externalresources.sqlite` and add that file to the `.gitignore`

src/hdmf/common/resources.py
```diff
@@ -293,41 +325,74 @@ def get_resource(self, **kwargs):
...
+             'doc': 'the attribute of the container for the external reference', 'default': None},
             {'name': 'key', 'type': (str, Key), 'default': None,
              'doc': 'the name of the key or the Row object from the KeyTable for the key to add a resource for'},
             {'name': 'resources_idx', 'type': Resource, 'doc': 'the resourcetable id', 'default': None},
             {'name': 'resource_name', 'type': str, 'doc': 'the name of the resource to be created', 'default': None},
             {'name': 'resource_uri', 'type': str, 'doc': 'the uri of the resource to be created', 'default': None},
             {'name': 'entity_id', 'type': str, 'doc': 'the identifier for the entity at the resource', 'default': None},
-            {'name': 'entity_uri', 'type': str, 'doc': 'the URI for the identifier at the resource', 'default': None})
+            {'name': 'entity_uri', 'type': str, 'doc': 'the URI for the identifier at the resource', 'default': None},
+            {'name': 'field', 'type': str, 'default': None,
```
rly: For the sake of clean code organization, please move the 'field' definition to under the 'attribute' definition. It should not make a functional difference.

src/hdmf/common/resources.py
```diff
@@ -293,41 +325,74 @@ def get_resource(self, **kwargs):
...
+                    parent_cls = self.type_map.get_dt_container_cls(data_type=parent_spec.data_type, autogen=False)
+                    if isinstance(container, parent_cls):
+                        parent_id = container.object_id
+                        # We need to get the path of the spec for relative_path
+                        absolute_path = spec.path
+                        relative_path = re.sub("^.+?(?="+container.data_type+")", "", absolute_path)
+                        object_field = self._check_object_field(parent_id, relative_path, field)
+                    else:
+                        msg = 'Container not the nearest data_type'
+                        raise ValueError(msg)
```
rly, line 383: Please add a test for this situation

rly:
I added a [commit](https://github.com/hdmf-dev/hdmf-common-schema/pull/62) to hdmf-common-schema to update the version number. Please update the subrepo in this PR

src/hdmf/common/resources.py
```diff
@@ -145,7 +151,9 @@ class ExternalResources(Container):
             {'name': 'objects', 'type': ObjectTable, 'default': None,
              'doc': 'the table storing object information'},
             {'name': 'object_keys', 'type': ObjectKeyTable, 'default': None,
-             'doc': 'the table storing object-resource relationships'})
+             'doc': 'the table storing object-resource relationships'},
+            {'name': 'type_map', 'type': TypeMap, 'default': None,
+             'doc': 'the type_map'})
```
```suggestion
            {'name': 'type_map', 'type': TypeMap, 'default': None,
             'doc': 'The type map. If None is provided, the HDMF-common type map will be used.'})
```

src/hdmf/common/resources.py
```diff
@@ -199,75 +210,96 @@ def _add_resource(self, **kwargs):
...
             objecttable_idx = list(set(objecttable_idx) & set(field_idx))
 
         if len(objecttable_idx) == 1:
             return self.objects.row[objecttable_idx[0]]
         elif len(objecttable_idx) == 0:
-            return self._add_object(container, field)
+            return self._add_object(container, relative_path, field)
         else:
-            raise ValueError("Found multiple instances of the same object_id and field in object table")
+            raise ValueError("Found multiple instances of the same object_id and relative_path in object table")
```
rly: Please add a test for this situation

src/hdmf/common/resources.py
```diff
@@ -293,41 +325,74 @@ def get_resource(self, **kwargs):
...
+                parent_spec = spec.parent  # return the parent spec of the attribute
+                if parent_spec.data_type is None:
+                    while parent_spec.data_type is None:
+                        parent_spec = parent_spec.parent  # find the closest parent with a data_type
+                    parent_cls = self.type_map.get_dt_container_cls(data_type=parent_spec.data_type, autogen=False)
+                    if isinstance(container, parent_cls):
+                        parent_id = container.object_id
+                        # We need to get the path of the spec for relative_path
+                        absolute_path = spec.path
+                        relative_path = re.sub("^.+?(?="+container.data_type+")", "", absolute_path)
```
rly, line 379: What does this relative expression do? Please add a comment. Also note that `pathlib` is a handy standard library for path manipulation that may be useful here.

src/hdmf/common/resources.py
```diff
@@ -277,7 +306,7 @@ def get_key(self, **kwargs):
             else:
                 return self.keys.row[key_idx_matches[0]]
 
-    @docval({'name': 'resource_name', 'type': str, 'default': None})
+    @docval({'name': 'resource_name', 'type': str, 'doc': 'The name of the resource.'})
```
rly, line 313: I removed `'default': None`. Why would someone call `get_resource` without a name? That would just raise an error.

src/hdmf/common/resources.py
```diff
@@ -186,88 +197,103 @@ def _add_entity(self, **kwargs):
...
-        Specify that an object (i.e. container and field) uses a key to reference
-        an external resource
+        Specify that an object (i.e. container and relative_path) uses a key to reference
+        an external resource.
         """
         obj, key = popargs('obj', 'key', kwargs)
         return ObjectKey(obj, key, table=self.object_keys)
 
-    def _check_object_field(self, container, field):
+    @docval({'name': 'container', 'type': (str, AbstractContainer), 'default': None,
```
rly: When would `container` ever be None here? Also you had the default for `relative_path` and `field` as None and then if they were None, you changed them to `''`. I simplified this by setting the default to `''`.

src/hdmf/common/resources.py
```diff
@@ -186,88 +197,103 @@ def _add_entity(self, **kwargs):
...
         """
         if isinstance(container, str):
             objecttable_idx = self.objects.which(object_id=container)
         else:
             objecttable_idx = self.objects.which(object_id=container.object_id)
 
         if len(objecttable_idx) > 0:
-            field_idx = self.objects.which(field=field)
+            field_idx = self.objects.which(relative_path=relative_path)
             objecttable_idx = list(set(objecttable_idx) & set(field_idx))
```
rly: Does this need to be updated to handle both `relative_path` and `field`?

docs/gallery/plot_external_resources.py
```diff
@@ -58,7 +58,7 @@
 #    They are disambiguated by the :py:class:`~hdmf.common.resources.Object` associated
 #    with each.
```
oruebel:
```suggestion
#    They are disambiguated by the :py:class:`~hdmf.common.resources.Object` associated
#    with each. I.e.,  we may have keys with the same name in different objects, but for a particular object
#    all keys must be unique. This means the :py:class:`~hdmf.common.resources.KeyTable` may contain
#    duplicate entries, but the :py:class:`~hdmf.common.resources.ObjectKeyTable` then must not assign 
#    duplicate keys to the same object.
```

docs/gallery/plot_external_resources.py
```diff
@@ -73,6 +73,16 @@
 #    This may or may not include a string representing the resource and a colon.
 #    Use the format provided by the resource. For example, Identifiers.org uses the ID ``ncbigene:22353``
 #    but the NCBI Gene uses the ID ``22353`` for the same term.
+# 8. In a majority of cases, :py:class:`~hdmf.common.resources.Object` objects will have an empty string
+#    for 'field'. The :py:class:`~hdmf.common.resources.ExternalResources` class supports compound data_types.
+#    In that case, 'field' would be the field of the compound data_type that has an external reference.
+# 9. The :py:class:`~hdmf.common.resources.Object` object we add to the
+#    :py:class:`~hdmf.common.resources.ObjectTable` has a data_type.
```
oruebel:
```suggestion
# 9. The :py:class:`~hdmf.common.resources.Object.object_id` object we add to the
#    :py:class:`~hdmf.common.resources.ObjectTable` has a data_type.
```

Not sure if this is correct edit. Can you clarify what you want to express in rule 9. ``hdmf.common.resources.Object`` has ``object_id``, ``relative_path``, and ``field``. The ``object_id`` refers to an object with a ``data_type`` but the ``relative_path`` will point to an object without a ``data_type``.

docs/gallery/plot_external_resources.py
```diff
@@ -73,6 +73,16 @@
...
+# 10. In some cases, the attribute that needs an external reference is not a object with a 'data_type'.
+#     The user must then use the nearest object that has a data type to be used as the container. When
+#     adding an external resource for an object with a data type, users should not provide an attribute.
+#     When adding an external resource for an attribute of an object, users need to provide
+#     the name of the attribute.
```
oruebel:
```suggestion
# 10. In some cases, the attribute that needs an external reference is not a object with a 'data_type'.
#     The user must then use the nearest object that has a data type to be used as the parent object. When
#     adding an external resource for an object with a data type, users should not provide an attribute.
#     When adding an external resource for an attribute of an object, users need to provide
#     the name of the attribute.
```

docs/gallery/plot_external_resources.py
```diff
@@ -196,20 +202,21 @@
...
-# In the above example, the ``field`` keyword argument was empty because the data
-# of the :py:class:`~hdmf.container.Data` object passed in for the ``container``
-# argument was being associated with a resource. However, you may want to associate
-# an attribute of a :py:class:`~hdmf.container.Data` object with a resource, or
-# a dataset or attribute of a :py:class:`~hdmf.container.Container` object with
-# a resource. To disambiguate between these different fields, you can set the
-# 'field' keyword.
+# It is important to keep in mind that the :py:class:`~hdmf.common.resources.Object` object
+# we add to the :py:class:`~hdmf.common.resources.ObjectTable` is the object
+# holding the attribute or the closest parent object. Recall that the objects in the
```
oruebel:
```suggestion
# It is important to keep in mind that the :py:class:`~hdmf.common.resources.Object.object_id` object
# we add to the :py:class:`~hdmf.common.resources.ObjectTable` is the object
# holding the attribute or the closest parent object. Recall that the objects in the
```

Doesn't the `hdmf.common.resources.Object` refer to the full object (i.e, including object_id, relative_path, and field), not just the closest parent? The closest parent is `hdmf.common.resources.Object.object_id`

mavaylon1: hdmf.common.resources.Object does refer to the full Object. I was saying how the Object needs to be the closest/holding the attribute. hdmf.common.resources.Object.object_id is the id of that object.

docs/gallery/plot_external_resources.py
```diff
@@ -196,20 +202,21 @@
...
+# :py:class:`~hdmf.common.resources.ObjectTable` must have a 'data_type'.
+# A common example would be with the :py:class:`~hdmf.common.table.DynamicTable` class, which holds
+# :py:class:`~hdmf.common.table.VectorData` objects as columns. If we wanted to add an external
+# reference on a column from a :py:class:`~hdmf.common.table.DynamicTable`, then we would use the
+# column as the object and not the :py:class:`~hdmf.common.table.DynamicTable` (Refer to rule 10).
```
oruebel: This text implies that we should use``er.add_ref(container=genotypes.genotype_name, attribute=None)`` in the example below. I think it's fine to leave the example as is (since this is probably what most users will intuitively do), but we should mention here that ``add_ref`` internally resolves the object to the closest parent, so that ``er.add_ref(container=genotypes, attribute='genotype_name')`` and ``er.add_ref(container=genotypes.genotype_name, attribute=None)`` will ultimatly both use the object_id of the ``genotypes.genotype_name`` VectorData column and not the object_id of the ``genotypes`` table.

src/hdmf/common/resources.py
```diff
@@ -186,88 +196,105 @@ def _add_entity(self, **kwargs):
...
         uri = kwargs['uri']
         resource = Resource(resource_name, uri, table=self.resources)
         return resource
 
     @docval({'name': 'container', 'type': (str, AbstractContainer),
-             'doc': 'the Container/Data object to add or the object_id for the Container/Data object to add'},
-            {'name': 'field', 'type': str, 'doc': 'the field on the Container to add'})
+             'doc': 'The Container/Data object to add or the object id of the Container/Data object to add.'},
+            {'name': 'relative_path', 'type': str, 'doc': 'The relative path on the Container to add.'},
+            {'name': 'field', 'type': str, 'default': '',
```
```suggestion
            {'name': 'relative_path', 'type': str, 'doc': 'The relative path on the object to add.'},
            {'name': 'field', 'type': str, 'default': '',
```

We need to be careful here with language. The relative_path is relative with respect to the file hierarchy described by the schema. The Container is an API class, and as such does not have relative paths.

src/hdmf/common/resources.py
```diff
@@ -186,88 +196,105 @@ def _add_entity(self, **kwargs):
...
+        container, relative_path, field = popargs('container', 'relative_path', 'field', kwargs)
         if isinstance(container, AbstractContainer):
             container = container.object_id
-        obj = Object(container, field, table=self.objects)
+        obj = Object(container, relative_path, field, table=self.objects)
         return obj
 
-    @docval({'name': 'obj', 'type': (int, Object), 'doc': 'the Object to that uses the Key'},
-            {'name': 'key', 'type': (int, Key), 'doc': 'the Key that the Object uses'})
+    @docval({'name': 'obj', 'type': (int, Object), 'doc': 'The Object to that uses the Key.'},
```
```suggestion
    @docval({'name': 'obj', 'type': (int, Object), 'doc': 'The Object that uses the Key.'},
```

src/hdmf/common/resources.py
```diff
@@ -186,88 +196,105 @@ def _add_entity(self, **kwargs):
...
         """
         obj, key = popargs('obj', 'key', kwargs)
         return ObjectKey(obj, key, table=self.object_keys)
 
-    def _check_object_field(self, container, field):
+    @docval({'name': 'container', 'type': (str, AbstractContainer),
+             'doc': ('The Container/Data object that uses the key or '
+                     'the object id for the Container/Data object that uses the key.')},
+            {'name': 'relative_path', 'type': str,
+             'doc': 'The relative path of the Container that uses the key.', 'default': ''},
```
```suggestion
            {'name': 'relative_path', 'type': str,
             'doc': 'The relative path to the object that uses the key.', 'default': ''},
```

src/hdmf/common/resources.py
```diff
@@ -363,34 +425,34 @@ def add_ref(self, **kwargs):
...
 
     @docval({'name': 'container', 'type': (str, AbstractContainer),
-             'doc': 'the Container/data object that is linked to resources/entities',
-             'default': None},
-            {'name': 'field', 'type': str,
-             'doc': 'the field of the Container',
-             'default': None})
+             'doc': 'The Container/data object that is linked to resources/entities.'},
+            {'name': 'relative_path', 'type': str, 'doc': 'The relative path of the Container.',
+             'default': ''},
```
```suggestion
            {'name': 'relative_path', 'type': str, 'doc': 'The relative path to the relevant data object within the parent data object given by the container parameter.',
             'default': ''},
```

docs/gallery/plot_external_resources.py
oruebel: In that case we should maybe say:

It is important to keep in mind that when adding and  :py:class:`~hdmf.common.resources.Object`` to the  :py:class:`~hdmf.common.resources.ObjectTable`, the parent object identified by :py:class:`~hdmf.common.resources.Object.object_id` must be the closest parent to the target object (i.e., :py:class:`~hdmf.common.resources.Object.relative_path` must be the shortest possible path and as such cannot contain any objects with a `data_type` and associated `object_id`)

oruebel (approved):

PR #645, opened 2021-07-16
Add functionality for linked DynamicTables

### LINKED ISSUE #646: DyanmicTableRegion not working correctly with AlignedDynamicTable
## Description

When a ``DynamicTableRegion`` points to an ``AlignedDynamicTable`` loading data via the slicing into the  ``DynamicTableRegion`` currently does not retrieve the data from all columns of the ``AlignedDynamicTable`` but it only loads the data from the columns in the primary table, ignoring all data from the category subtables in ``AlignedDynamicTable``. 

**Cause:** It appears that the cause for this is issue is that ``AlignedDynamicTable`` only overwrites the ``__getitem__`` function but not the ``get`` function. ``DynamicTableRegion`` uses the later, and since ``AlignedDynamicTable.get`` is inherited from ``DynamicTable`` the category subtables are being ignored.

**Solution:** Add ``AlignedDynamicTable.get`` based on the existing ``AlignedDynamicTable.__get__``

## Checklist

- [X] Have you ensured the bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?

### LINKED ISSUE #651: AlignedDyanmicTable does not support [int, str] type slicing
## Description

DynamicTable`` uses the syntax ``dt[int, str]`` (i.e., ``dt[row_index, column_name]``) to select a single cell from the table. ``AlignedDynamicTable`` however currently only supports slicing with tuples of strings to select whole tables or columns, e.g. ``at[category_name, columns_name]``. To retrieve a single cell, ``AlignedDynamicTable`` currently requires that the row index be listed last, e.g., ``at[category_name, columns_name, 0]`` For consistency with ``DynamicTable``, ``AlignedDynamicTable`` should be updated to allow for selections of the form:
* ``[int, str]`` (i.e., ``[row_index, category_name]``) to select a single row of a category table
* ``[int, (str, str)]`` (i.e., [row_index, category_name, column_name]``) to select a single cell of the table. 
   * ``[int, str, str]`` should work the same as ``[int, (str, str)]``

## Checklist

- [X] Have you ensured the bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?

oruebel:
## Motivation

For the icephys work with NWB we would like to simplify interacting with DynamicTables that reference other tables via DynamicTableRegion, creating a collection of linked tables. In the ICEphys case this is a "simple" linear hierarchy of tables, but in principle a table may contain any number of DynamicTableRegion columns. This PR adds several functions to simplify introspection of linked DynamicTables and conversion to pandas DataFrames. 

## TODO

- [X] Fix #646 by adding ``AlignedDynamicTable.get``
- [X] Fix #651 by updating ``AlignedDynamicTable.get`` to support slicing with  ``[int, (str, str)]``, ``[int, str, str]``, and ``[int, str]`` to select a single cell or row of a category table, repectively
- [X] Add ``AlignedDynamicTable.get_colnames(...)`` functions to allow us to keep compliance of the ``colnames`` property with ``DynamicTable`` while providing an easy way to get the full list of column names.
- [X] Set name of DataFrame in ``DynamicTable.to_dataframe()`` and ``DynamicTable.get`` 
- [X] Add helper functions to ``DynamicTable`` to deal with foreign columns:
    - [X] ``DynamicTable.get_foreign_columns`` to identify if the table contains ``DynamicTableRegion`` columns
    - [X] ``DynamicTable.has_foreign_columns`` to  identify which columns are``DynamicTableRegion`` columns 
    - [X] ``DynamicTable.get_linked_tables`` to retrieve all tables linked to either directly or indirectly from
      the current table via ``DynamicTableRegion``
- [x] Implement the same helper functions also for ``AlignedDynamicTable``
    - [x] ``DynamicTable.get_foreign_columns`` to identify if the table contains ``DynamicTableRegion`` columns
    - [X] ``DynamicTable.has_foreign_columns`` to  identify which columns are``DynamicTableRegion`` columns 
    - [x] ``DynamicTable.get_linked_tables`` to retrieve all tables linked to either directly or indirectly from
      the current table via ``DynamicTableRegion``
- [X] Add new module ``hdmf.common.hierarchicaltable`` with helper functions to facilitate conversion of linked tables to a single Pandas dataframe. 
     - [X] ``to_hierarchical_dataframe`` to merge linked tables into a single consolidated pandas DataFrame.
     - [X] ``drop_id_columns`` to remove "id" columns from a DataFrame.
     - [X] ``flatten_column_index`` to replace a ``pandas.MultiIndex`` with a regular ``pandas.Index``
- [x]  Add test for DyanmicTableRegion pointing to AlignedDynamicTable to check that the all columns are used
- [x] Add tests for hierarchicaltable.py for 
    - [X] to_hierarchical_dataframe
    - [x] drop_id_columns
    - [x]  flatten_column_index functions
- [X] File issue tickets for open TODO items for future PRs 
    - [X] ``to_hierarchical_dataframe`` should be updated to support resolution of more than one DynamicTableRegion column. See https://github.com/hdmf-dev/hdmf/issues/649
    - [x] Add tutorial for DynamicTableRegion and how to use for linking to tables and for creating linked tables. See https://github.com/hdmf-dev/hdmf/issues/648

## Checklist


src/hdmf/common/alignedtable.py
```diff
@@ -259,3 +262,60 @@ def __getitem__(self, item):
...
+    @docval({'name': 'ignore_category_tables', 'type': bool,
+             'doc': "Ignore the category tables and only check in the main table columns", 'default': False},
+            allow_extra=False)
+    def get_foreign_columns(self, **kwargs):
+        """
+        Determine the names of all columns that link to another DynamicTable, i.e.,
+        find all DynamicTableRegion type columns. Similar to a foreign key in a
+        database, a DynamicTableRegion column references elements in another table.
+
+        :returns: List of tuples of strings with the column names
```
rly:
```suggestion
        :returns: List of tuples (str, str) where the first string is the name of the
                  category table (or None if the column is in the main table) and the
                  second string is the column name.
```

src/hdmf/common/table.py
```diff
@@ -47,9 +47,22 @@ def add_row(self, **kwargs):
...
+        Retrieve elements from this VectorData
+
+        :param key: Selection of the elements
+        :param **kwargs: Ignored
+        """
         return super().get(key)
 
     def extend(self, ar, **kwargs):
+        """
+        The extend_data method adds all the elements of the iterable arg to the end of this VectorData.
```
```suggestion
        """Add all elements of the iterable arg to the end of this VectorData.
```

src/hdmf/data_utils.py
```diff
@@ -27,6 +27,12 @@ def append_data(data, arg):
 
 
 def extend_data(data, arg):
+    """
+    The extend_data method adds all the elements of the iterable arg to the end of  data.
```
rly:
```suggestion
    """Add all the elements of the iterable arg to the end of data.
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+        rtype='pandas.DataFrame',
+        is_method=False)
+def to_hierarchical_dataframe(dynamic_table):
+    """
+    Create a hierarchical pandas.DataFrame that represents all data from a collection of linked DynamicTable.
+
+    **LIMITATIONS:** Currently this function only supports DynamicTables with a single DynamicTableRegion column.
+    If a table has more than one foreign DynamicTableRegion column then the function will expand only the
+    first DynamicTableRegio column found for each table. Any additional DynamicTableRegion columns will remain
+    nested.
```
rly:
```suggestion
    Create a hierarchical pandas.DataFrame that represents all data from a collection of linked DynamicTables.

    **LIMITATIONS:** Currently this function only supports DynamicTables with a single DynamicTableRegion column.
    If a table has more than one DynamicTableRegion column then the function will expand only the
    first DynamicTableRegion column found for each table. Any additional DynamicTableRegion columns will remain
    nested.
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+    data = []
+    columns = None
+    index_names = None
+
+    # If we have indexed columns (other than our hierarchical column) then our index data for our
+    # MultiIndex will contain lists as elements (which are not hashable) and as such create an error.
+    # As such we need to check if we have any affected columns so we can  fix our data
+    # indexed_column_indicies = np.where([isinstance(dynamic_table[colname], VectorIndex)
+    #                                    for colname in dynamic_table.colnames if colname != hcol_name])[0]
+    # indexed_column_indicies += 1  # Need to increment by 1 since we add the row id in our iteration below
```
rly: Does this still need to be here? If so, please add a comment as to why it is commented out and how that changes/limits the behavior of the function if at all.

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Case 1:  Our DynamicTableRegion column points to a regular DynamicTable
+    #          If this is the case than we need to de-normalize the data and flatten the hierarchy
+    if not hcol_target.has_foreign_columns():
+        # 1) Iterate over all rows in our hierarchical columns (i.e,. the DynamicTableRegion column)
+        #    First we here get a list of DataFrames, one for each row. If our hcol is a VectorIndex
+        #    (i.e., out column is a ragged array), then simply loading the data from the VectorIndex
```
rly:
```suggestion
        #    (i.e., our column is a ragged array), then simply loading the data from the VectorIndex
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+        #    will do the trick. If we have a regular VectorData column, then we need to load the
+        #    elements ourselves (using slice syntax to make sure we get DataFrames) one-row-at-a-time
+        if isinstance(hcol, VectorIndex):
+            rows = hcol.get(slice(None), index=False, df=True)
+        else:
+            rows = [hcol[i:(i+1)] for i in range(len(hcol))]
+        # Now we iterate over the rows, where each row is described by a DataFrame with one-or-more rows
+        for row_index, row_df in enumerate(rows):
+            # 1.1): Since hcol is a DynamicTableRegion, each row returns another pandas.DataFrame so we
+            #       next need to iterate over all rows in that table to denormalize our data
```
rly: There is no 1.2 and no guarantee that hcol is a DynamicTableRegion. Can this comment be removed and the lower numbers updated from "1.1.1" to "1.1"?
```suggestion
            # Iterate over all rows in that table to denormalize our data
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+                #        table, ii) all columns (except the hierarchical column we are flattening), and
+                #        iii) the index (i.e., id) from our target row
+                index_data = ([dynamic_table.id[row_index], ] +
+                              [dynamic_table[row_index, colname]
+                               for colname in dynamic_table.colnames if colname != hcol_name])
+                # for i in indexed_column_indicies:  # Fix data from indexed columns
+                #    index_data[i] = tuple(index_data[i])  # Convert from list to tuple (which is hashable)
+                index.append(tuple(index_data))
+                # Determine the names for our index and columns of our output table if this is the first row.
+                # These are constant for all rows so we only need to do this onle once for the first row.
```
rly:
```suggestion
                # These are constant for all rows so we only need to do this only once for the first row.
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+                if row_index == 0:
+                    index_names = ([(dynamic_table.name, 'id')] +
+                                   [(dynamic_table.name, colname)
+                                    for colname in dynamic_table.colnames if colname != hcol_name])
+                    columns = pd.MultiIndex.from_tuples([(hcol_target.name, 'id'), ] +
+                                                        [(hcol_target.name, c) for c in row_df.columns],
+                                                        names=('source_table', 'label'))
```
rly: When would these lines result in a different `columns` than the `columns` generated by lines 98-100? i.e. can this be removed and the if condition on line 94 removed so that lines 95-100 are always run?

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+            columns = pd.MultiIndex.from_tuples([(hcol_target.name, 'id'), ] +
+                                                [(hcol_target.name, c) for c in hcol_target.colnames],
+                                                names=('source_table', 'label'))
+
+    # Case 2:  Our DynamicTableRegion columns points to another table with a DynamicTableRegion
+    else:
+        # 1) First we need to recursively flatten the hierarchy by calling 'to_hierarchical_dataframe()'
+        #    (i.e., this function) on the target of our hierarchical column
+        #    First we here get a list of DataFrames, one for each row. If our hcol is a VectorIndex
+        #    (i.e., out column is a ragged array), then simply loading the data from the VectorIndex
```
rly:
```suggestion
        #    (i.e., our column is a ragged array), then simply loading the data from the VectorIndex
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+        #    will do the trick. If we have a regular VectorData column, then we need to load the
+        #    elements ourselves (using slice syntax to make sure we get DataFrames) one-row-at-a-time
+        hcol_hdf = to_hierarchical_dataframe(hcol_target)
+        if isinstance(hcol, VectorIndex):
+            rows = hcol.get(slice(None), index=False, df=True)
+        else:
+            rows = [hcol[i:(i+1)] for i in range(len(hcol))]
```
rly: Can we move these lines and associated comment to outside of the containing if statement? This seems to be the same as lines 63-66.

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+        # 2) Iterate over all rows, where each row is described by a DataFrame with one-or-more rows
+        for row_index, row_df_level1 in enumerate(rows):   # need hcol[:] here  in case this is an h5py.Dataset
```
rly: What does the comment on this line refer to?

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,258 @@
...
+            # 1.1): Since hcol is a DynamicTableRegion, each row returns another DynamicTable so we
+            #       next need to iterate over all rows in that table to denormalize our data
```
rly: See comment above regarding 1.1

src/hdmf/common/hierarchicaltable.py
oruebel: I think this can go. As I was going through testing I realized that this was actually not used. And I honestly couldn't come up with a case where this would be still needed.

src/hdmf/common/hierarchicaltable.py
oruebel: Ok, I'll update the comments. The original logic assumed that hcol was a VectorIndex (which is true for the ICEPhys tables) but depending on the structure of the tables hcol can either be a DynamicTableRegion or a VectorIndex that points to a DynamicTableRegion. This is being dealt with before the loop now when constructing the rows array.

src/hdmf/common/hierarchicaltable.py
oruebel: I'll check, but I think yes.

src/hdmf/common/hierarchicaltable.py
oruebel:
```suggestion
        for row_index, row_df_level1 in enumerate(rows):
```

src/hdmf/common/hierarchicaltable.py
oruebel: Sounds good.

src/hdmf/common/hierarchicaltable.py
oruebel: I remove the commented code.

src/hdmf/common/hierarchicaltable.py
oruebel: I updated the comments throughout the to_hierarchical_dataframe function accordingly.

src/hdmf/common/hierarchicaltable.py
oruebel: In the "Case 2" part (i.e., in the else statement) the logic inside and outside of the loop is identical, so I was indeed able to removed the logic in the loop and updated the code per your suggestion.

In the "Case 1" part of the code, the key difference is that within the loop we can use ``row_df.columns`` to determine the columns and outside we have to look at ``hcol_target.colnames``. I need to take a closer look at why this leads to a difference (I honestly don't remember, as I wrote this part of the code nearly a year ago) but tests are failing without the logic, so I left this as is for now until I had a chance to investigate further.

src/hdmf/common/hierarchicaltable.py
oruebel: Yes we can. Done.

src/hdmf/common/hierarchicaltable.py
oruebel: Done.

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,241 @@
...
+        index_names = ([(dynamic_table.name, "id")] +
+                       [(dynamic_table.name, colname)
+                        for colname in dynamic_table.colnames if colname != hcol_name] +
+                       hcol_hdf.index.names)
+        columns = hcol_hdf.columns
+
+    # Construct the pandas dataframe with the hierarchical multi-index
+    multi_index = pd.MultiIndex.from_tuples(index, names=index_names)
+    out_df = pd.DataFrame(data=data, index=multi_index, columns=columns)
+    # Update the dataframe to remove id columns if requested
```
rly:
```suggestion
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,241 @@
...
+                for row_tuple_level3 in hcol_hdf.loc[[row_df_level2[0]]].itertuples(index=True):
+                    # Determine the column data for our row.
+                    data.append(row_tuple_level3[1:])
+                    # Determine the multi-index tuple for our row,
+                    index_data = ([dynamic_table.id[row_index], ] +
+                                  [dynamic_table[row_index, colname]
+                                   for colname in dynamic_table.colnames if colname != hcol_name] +
+                                  list(row_tuple_level3[0]))
+                    index.append(tuple(index_data))
+        # Determine the names for our index and columns of our output table if this is the first row
```
rly:
```suggestion
        # Determine the names for our index and columns of our output table
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,241 @@
...
+    :raises ValueError: In case the num_levels is not >0
+    :raises TypeError: In case that dataframe parameter is not a pandas.Dataframe.
+    """
+    dataframe, max_levels, inplace = getargs('dataframe', 'max_levels', 'inplace', kwargs)
+    if max_levels is not None and max_levels <= 0:
+        raise ValueError('max_levels must be greater than 0')
+    # Compute the new column names
+    col_names = [__flatten_column_name(col) for col in dataframe.columns.values]
+    # Apply the max_levels filter. Make sure to do this only for columns that are actually tuples
+    # in order not ot accidentally shorten the actual string name of columns
```
rly:
```suggestion
    # in order not to accidentally shorten the actual string name of columns
```

src/hdmf/common/hierarchicaltable.py
```diff
@@ -0,0 +1,241 @@
...
+    select_levels = slice(None) if max_levels is None else slice(-max_levels, None) if max_levels > 1 else -1
```
rly:
```suggestion
    if max_levels is None:
        select_levels = slice(None)
    elif max_levels == 1:
        select_levels = -1
    else:  # max_levels > 1
        select_levels = slice(-max_levels, None)
```
Easier to read code

src/hdmf/common/hierarchicaltable.py
oruebel: Ok, the culprit is ``AlignedDynamicTable``. In the case of a regular ``DynamicTable`` ``hcol_target.colnames`` will give us the full list of column names but for ``AlignedDynamicTable`` this is only the columns of the main table without the subtables. I.e., we should be able to eliminate the difference in logic here, but need to deal with this difference in how we determine the columns.

rly:
I made a few more comments. Otherwise, looks good to me

src/hdmf/common/hierarchicaltable.py
oruebel: I added a ``AlignedDynamicTable.get_colnames`` to allow us to deal with this issues. The problem is that the property ``AlignedDynamicTable.colnames`` can't really change to include all columns as the logic of ``DynamicTable`` depends on this, so we need a different way to get all the columns. With this we can eliminate the duplication in code you mentioned. 

This same issue also affected both the logic for determining ``index_names`` and ``columns`` for both Case 1 and 2 here. I'll try to add some more test to see that I can hit all the different cases in the logic here.

oruebel:
@rly thanks for the helpful code review! Your comments have helped clean up ``to_hierarchical_dataframe`` and revealed some additional issues with the handling of ``AlignedDynamicTable``. I appreciate the feedback.

oruebel:
@rly I have now resolved all comments. Can you please have another look and approve the PR if it looks good. 
Should this be merged with dev or another release branch?

oruebel:
@rly while adding another test to ensure ``to_hierarchical_dataframe`` is working correctly with ``AlignedDynamicTable`` higher up in the table hierarchy (not just at the bottom) I discovered two issues:
1. issue #651 and 
2. there were also a few more places in ``to_hierarchical_dataframe``  where ``dataframe.colnames`` was used instead of ``dataframe.get_colnames``. 

Those issues are now fixed as well. I think these should be last changes on this PR 🤞

src/hdmf/common/alignedtable.py
```diff
@@ -20,6 +20,10 @@ class AlignedDynamicTable(DynamicTable):
     defines a 2-level table in which the main data is stored in the main table implemented by this type
     and additional columns of the table are grouped into categories, with each category being'
     represented by a separate DynamicTable stored within the group.
+
+    NOTE: To remain compatavibility with DynamicTable, the attribute colnames represents only the
+          columns of the main table (not including the category tables). To get the full list of
+          columns names uses the get_colnames() function instead.
```
rly:
```suggestion
    NOTE: To remain compatible with DynamicTable, the attribute colnames represents only the
          columns of the main table (not including the category tables). To get the full list of
          column names, use the get_colnames() function instead.
```

src/hdmf/common/alignedtable.py
```diff
@@ -209,6 +213,23 @@ def add_row(self, **kwargs):
         for category, values in category_data.items():
             self.category_tables[category].add_row(**values)
 
+    @docval({'name': 'include_category_tables', 'type': bool,
+             'doc': "Ignore sub-category tables and just look at the main table", 'default': False},
+            {'name': 'ignore_category_ids', 'type': bool,
+             'doc': "Ignore id columns of sub-category tables", 'default': False})
+    def get_colnames(self, **kwargs):
+        """Get the full list of names of columns for this table"""
```
rly:
```suggestion
        """Get the full list of names of columns for this table
        
        :returns: List of tuples (str, str) where the first string is the name of the DynamicTable
                  that contains the column and the second string is the name of the column. If
                  include_category_tables is False, then a list of column names is returned.
        """
```

src/hdmf/common/alignedtable.py
```diff
@@ -257,9 +278,29 @@ def get(self, item, **kwargs):
                 return self.get_category(item).to_dataframe()
         elif isinstance(item, tuple):
             if len(item) == 2:
-                return self.get_category(item[0])[item[1]]
+                if isinstance(item[0], (int, np.integer)):
```
rly, line 329: Please add a description for this case in the function docstring, and optionally the error message on line 305.

oruebel: Ok, I have overhauled the docstring for ``AlignedDynamicTable.get`` and add further documentation to the body and error message.

oruebel:
@rly thanks for the additional comments. I have updated the source accordingly.

src/hdmf/common/alignedtable.py
```diff
@@ -245,21 +250,59 @@ def to_dataframe(self, **kwargs):
...
+
+        * **int, list, array, slice** : Return one or multiple row of the table as a pandas.DataFrame. For example:
+              * ``self[0]`` : Select the first row of the table
+              * ``self[[0,3]]`` : Select the first and fourth row of the table
+              * ``self[1:4]`` : Select the rows with index 1,2,3 from the table
+
+        * **string** : Return a column from the main table or a category table. For example:
+              * ``self['column']`` : Return the column from the main table.
+              * ``self['my_category']`` : Returns a DataFrame of the ``my_category`` category table.
+                This is a shorthand for ``self.get_category.to_dataframe()``.
```
rly:
```suggestion
                This is a shorthand for ``self.get_category('my_category').to_dataframe()``.
```

src/hdmf/common/alignedtable.py
```diff
@@ -245,21 +250,59 @@ def to_dataframe(self, **kwargs):
...
+
+               The tuple itself then may take the following forms:
+
+               * Select a single column from a table via:
+                   * ``self[category, column]``
+               * Select a single full row of a given category table via:
+                   * ``self[row, category]`` (recommended, for consistency with DynamicTable)
+                   * ``self[category, row]``
+               * Select a single cell via:
+                   * ``self[row, (category, column)]`` (recommended, for consistency with DynamicTable))
```
rly:
```suggestion
                   * ``self[row, (category, column)]`` (recommended, for consistency with DynamicTable)
```

rly (dismissed):

rly (approved):

PR #672, opened 2021-10-04
GenericDataChunkIterator

CodyCBakerPhD:
## Motivation

Requested PR for a more general type of `DataChunkIterator` that is also less abstract in functionality than the `AbstractDataChunkIterator`; this is termed the `GenericDataChunkIterator`. This class is still partially abstract in that it is built only to handle chunking (in writing to a HDF5 file) and buffering (as a general python iterator) of arbitrary dimensional data. The user must still specify how to go about accessing a particular type of data through the `_get_data` and `_get_dtype` methods. This is because of two reasons: (a) different data structure can sometimes be sensitive to how direct access can occur through slicing, e.g. `my_np_memmap[slice_1, slice_2]` may not access data same way as `my_np_array[slice_1, slice_2]`, and (b) we want this iterator to be agnostic to data access API's, which may not always return the exact some data object (`array`, `memmap`, `list`, `generator`, etc) but are instead called via the same retrieval function that handles the actual data objects separately.

The class is also designed to allow users to fully specify and control chunk and buffer shapes if desired, though offers robust default setting according to hdf5 cache suggestions.

In this respect, a basic implementation of this class found [here](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/utils/recordingextractordatachunkiterator.py) showed something like a 20% improved speed performance from the original `DataChunkIterator`. This is mainly due to two reasons:

1. More custom buffer shapes than provided by the `DataChunkIterator`, which automatically grabs full extents from axes other than the `iter_axis`
2. Equivocated chunk shapes between the iterator access and a `H5DataIO` wrapper, which would have potentially pulled and subsetted across chunks from the iterator without a manual override, leading to excess I/O operations. Example: `DataChunkIterator` with `iter_axis=0` pulls as much row data as fits in the `buffer_size`, but a default `H5DataIO` would set chunk shapes in the actual dataset as spanning potentially unequal rows.

## How to test the behavior?

A full suite has been added to `test_core_GenericDataChunkIterator.py`

## Checklist

Still some todo, in addition to below...

- [x] Add tests for new Iterator
- [x] Add documentation for new Iterator
- [x] See if old DataChunkIterator can be considered a sub case and still be back-compatible
- [x] Discuss the best way to incorporate parallelization through an MPI passed through the `DataIO`
- [x] Swap to using hdmf docval instead of annotation typing
- [x] Update tutorial

Required:

- [X] Is your contribution compliant with our coding style? This can be checked running `flake8` from the source directory. (Code is in `black` format, a subset of PEP8)


CodyCBakerPhD:
@oruebel This PR is now ready for full review! We've simplified and optimized the code a lot through multiple iterations and uses on `nwb-conversion-tools`, from what we added in the first commit. Should be a lot easier to read now, and added a ton of documentation. Let me know if you spot gaps or need clarification on anything.

Now, on the topic of using this to overhaul the existing `DataChunkIterator` class, I really don't think it would be possible to fully replace the functionality across arbitrary data objects by using the `GenericDataChunkIterator` as an intermediary, largely because the `DataChunkIterator` has so many special cases based on the inferred class of the passed data object type (`h5.Dataset`, `np.array`, etc). And it would actually be nice I think to always have a fully instantiated class like the `DataChunkIterator` that always works on any arbitrary object, even if the performance isn't by default as good as the more finely tuned one (which can also depend on how the user sets up the `_get_data` method for the `GenericDataChunkIterator`, another reason we leave that to them to specify; an example of that is, for certain `np.memmap`'s, it's much faster to slice consecutively along certain axes as opposed to others, and we don't really want to assume we know that by default).

docs/gallery/plot_generic_data_chunk_tutorial
```diff
@@ -0,0 +1,118 @@
+"""
+
+.. _dynamictable-tutorial:
+
+GenericDataChunkIterator Tutorial
+=====================
```
oruebel:
```suggestion

GenericDataChunkIterator Tutorial
==================================
```

docs/gallery/plot_generic_data_chunk_tutorial
```diff
@@ -0,0 +1,118 @@
...
+
+This is a tutorial for interacting with ``GenericDataChunkIterator`` objects. This tutorial
+is written for beginners and does not describe the full capabilities and nuances
+of the functionality. This tutorial is designed to give
+you basic familiarity with how ``GenericDataChunkIterator`` works and help you get started
```
oruebel:
```suggestion
you basic familiarity with how :py:class:`~hdmf.data_utils.GenericDataChunkIterator` works and help you get started
```

docs/gallery/plot_generic_data_chunk_tutorial
```diff
@@ -0,0 +1,118 @@
...
+with creating a specific instance for your data format or API access pattern.
+
+"""
+
+###############################################################################
+# Introduction
+# ------------
+# The :py:class:`~hdmf.data_utils.GenericDataChunkIterator` class represents a semi-abstract
+# version of a DataChunkIterator that automatically handles the selection of buffer regions
```
oruebel:
```suggestion
# version of a :py:class:`~hdmf.data_utils.AbstractDataChunkIterator` that automatically handles the selection of buffer regions
```

tests/unit/utils_test/test_core_GenericDataChunkIterator.py
```diff
@@ -0,0 +1,175 @@
...
+        self.assertEqual(
+            str(error.exception),
+            f"Some dimensions of chunk_shape ({chunk_shape}) do not evenly divide the buffer shape ({buffer_shape})!"
+        )
+
+    def test_buffer_option_assertions(self):
+        buffer_gb = -1
+        with self.assertRaises(AssertionError) as error:
+            self.TestNumpyArrayDataChunkIterator(array=self.test_array, buffer_gb=buffer_gb)
+        self.assertEqual(str(error.exception), f"buffer_gb ({buffer_gb}) must be greater than zero!")
```
bendichter:
```suggestion
        with self.assertRaises(AssertionError, msg=f"buffer_gb ({buffer_gb}) must be greater than zero!"):
            self.TestNumpyArrayDataChunkIterator(array=self.test_array, buffer_gb=buffer_gb)
```

oruebel (dismissed):
This PR looks good to. Very nice work.

A couple of additional suggestions below. However, I believe these are probably best addressed as separate issues/PRs. As a first step, I would suggest that you create issue tickets for these items as you deem appropriate, just so we don't forget to chat about these topics.

- HDMF currently misses a tutorial for iterative write. I think it would make sense to update the tutorial to discuss iterative write a bit more by adding 1) a short introduction on what iterative write is, 2) a brief section on ``AbstractDataChunkIterator``, and 3) at the end add a short section on ``DataChunkIterator``
- It would be useful to update the iterative data write tutorial in the PyNWB docs [here](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/iterative_write.html) to use this new iterator where appropriate and/or add a short section about ``GeneraticDataChunkIterator`` there as well and point to this tutorial
- Eventually, I think it may make sense to update ``DataChunkIterator`` to inherit from ``GenericDataChunkIterator`` to reduce redundancy

src/hdmf/data_utils.py
```diff
@@ -2,9 +2,12 @@
 from abc import ABCMeta, abstractmethod
 from collections.abc import Iterable
 from warnings import warn
+from typing import Tuple
+from itertools import product, chain
 
 import h5py
 import numpy as np
+import psutil
```
rly: if `psutil` is used only in `GenericDataChunkIterator`, can we make it an optional import? Since HDMF is a library, it would be better to keep it relatively lightweight in terms of dependencies. 

(To make it an optional import, import it when first used and raise an informative error if the package is not found. Also describe the optional import in the installation requirements and readme.)

requirements-dev.txt
```diff
@@ -8,3 +8,4 @@ flake8-print==4.0.0
 importlib-metadata==4.6.1
 python-dateutil==2.8.2
 tox==3.24.0
+psutil==5.8.0
```
rly:
```suggestion
```
`requirements-dev.txt` lists additional dev requirements. If `psutil` is listed in `requirements.txt`, it is not needed here

requirements-min.txt
```diff
@@ -6,3 +6,4 @@ pandas==1.0.5
 ruamel.yaml==0.16
 jsonschema==2.6.0
 setuptools
+psutil==5.8.0
```
rly:
```suggestion
psutil==2.0.0
```
This file lists minimum requirements and should match the `setup,py`

tests/unit/utils_test/test_core_GenericDataChunkIterator.py
rly: `assertRaises(error, msg=msg)` doesn't check that the error message equals `msg`. Instead, it outputs `msg` on error. https://stackoverflow.com/questions/58687475/self-assertraises-as-a-context-manager-with-msg-argument-not-working-as-expected

Use [`self.assertRaisesWith`](https://hdmf.readthedocs.io/en/stable/hdmf.testing.testcase.html#hdmf.testing.testcase.TestCase.assertRaisesWith) for an exact match, or `self.assertRaisesRegex` for a regex match instead.

rly: To clean up the test code, I recommend changing all of the message tests here to use `self.assertRaisesWith`.

CodyCBakerPhD: Thanks @rly, didn't know about that function but it's really nice! Swapped all assertion tests to this format.

CodyCBakerPhD:
On the `psutil` stuff, I'll just swap to using `os` directly in this case; there's a way to do it it's just a bit messier. `psutil` has a much cleaner function call but between adding another external dependency or handling lazy imports, I think it's probably better to just use a builtin library for it.

CodyCBakerPhD:
@rly After some thought, I decided it's better to just remove the `psutil` feature (only 1 line of code) from the base class and allow downstream users to decide if they want to have something like that for their own uses (like we do anyway on `nwb-conversion-tools`).

I looked into using `os` but apparently it only works linux systems and gave unreliable values as well.

I also added some new tests to ensure the `H5DataIO` wrapper works correctly around this new class (mostly that chunk shapes are communicated correctly), similar to the `DataChunkIterator`.

src/hdmf/data_utils.py
```diff
@@ -125,6 +127,225 @@ def maxshape(self):
...
+
+        k = np.floor(
+            (buffer_gb * 1e9 / (np.prod(self.chunk_shape) * self.dtype.itemsize)) ** (1 / len(self.chunk_shape))
+        )
+        return tuple([
+            min(max(int(x), self.chunk_shape[j]), self.maxshape[j])
+            for j, x in enumerate(k * np.array(self.chunk_shape))
+        ])
+
+    __docval_init = (
```
bendichter, line 133: why define this outside the constructor?

CodyCBakerPhD: (a) Makes the constructor cleaner and easier to read,
(b) Leaves the door open for if an advanced user wants to override it to give better performance on their specific data. I designed the one-shot calculation around 'typical' data shapes we usually see in `ecephys`, but there may be edge cases where it isn't as performant and the user might for example want to apply an exponentialized scaling preference for one axis over another. Or if they wanted really tight bounds with the size arguments, they could add iteration (which we avoid here because for `typical` data shapes we work with the one-shot calculation gets it pretty close). This is also why I moved many of the constructor shaping assertions to affect both user-specified shapes and those generated by these functions.

CodyCBakerPhD:
Ben also requested partial buffer shaping (i.e., `buffer_shape=(1,None, 2, ..)` with `buffer_gb=10` to indicate to load up to 10 GB along the unconstrained axes of the shape).

Putting this back to draft mode until then as it will take a bit of work.

CodyCBakerPhD:
@oruebel @rly Ready for final review; pushing off the partial shaping feature to a future PR

docs/gallery/plot_generic_data_chunk_tutorial.py
```diff
@@ -0,0 +1,118 @@
...
+#  ``hdmf.backends.hdf5.h5tools.HDF5IO`` context with a corresponding ``hdmf.container.Container``, these details
+#  will be automatically parsed.
+
+###############################################################################
+# .. note::
+#   There may be conflicts of nomenclature found between HDMF and HDF5. 'Chunks' in the HDF5 dataset are the tiny
+#   pieces of data that get compressed or cached together, and they should always be less than 1 megabyte (MB) in
+#   size for optimal performance, in constrast to the 'buffer' of the iterator whose size which should be as much
+#   free RAM as can be safely used. This buffer utilizes 'DataChunks' in the HDMF sense, which is any piece of
+#   data smaller than the full extent of the maximum shape.
```
oruebel:
```suggestion
###############################################################################
# .. note::
#   There is some overlap here in nomenclature between HDMF and HDF5. The term *chunk* in both
#   HDMF and HDF5 refer to a subset of dataset, however, in HDF5 a chunk is a piece of dataset on disk,
#   whereas in the context of the data chunk iteration a chunk is a block data in memory. As such, the 
#   requirements on the shape and size of chunks are different. In HDF5  chunks are typically small pieces
#   of a dataset that get compressed and cached together, and they should usually be small in size for 
#   optimal performance  (typically 1 MB or less). In contrast, a :py:class:`~hdmf.data_utils.DataChunk` in 
#   HDMF acts as a block of data for writing data to dataset. To improve performance and  avoid repeat 
#   updates to the same `Chunk` in the HDF5 file, :py:class:`~hdmf.data_utils.DataChunk` objects for write 
#   should align with `Chunks` in the HDF5 file, i.e., the :py:class:`~hdmf.data_utils.DataChunk.selection`   
#   should fully cover one or more `Chunks`  in the HDF5 file to avoid repeat updates to the same
#   `Chunks` in the HDF5 file. The `buffer` of the iterator then collects multiple 
#   :py:class:`~hdmf.data_utils.DataChunk` objects to help reduce the number of small I/O operations 
#   and help improve performance. The `buffer` should usually be large, i.e, be able to use
#   as much free RAM as can be safely used. 
```

How does this sound?

oruebel (dismissed):
Looks good to me. I added on minor comment, to rewrite a note in the tutorial, but that is more a difference in style and is not critical.

docs/gallery/plot_generic_data_chunk_tutorial.py
bendichter: we have changed from "chunk" to "buffer"

bendichter: I don't think the buffer collections multiple :py:class:`~hdmf.data_utils.DataChunk` objects, I think it's just one. We are using the term "buffer" to try to make this distinction more clear.

docs/gallery/plot_generic_data_chunk_tutorial.py
```diff
@@ -0,0 +1,127 @@
...
+#   of a dataset that get compressed and cached together, and they should usually be small in size for 
+#   optimal performance  (typically 1 MB or less). In contrast, a :py:class:`~hdmf.data_utils.DataChunk` in 
+#   HDMF acts as a block of data for writing data to dataset. To improve performance and  avoid repeat 
+#   updates to the same `Chunk` in the HDF5 file, :py:class:`~hdmf.data_utils.DataChunk` objects for write 
+#   should align with `Chunks` in the HDF5 file, i.e., the :py:class:`~hdmf.data_utils.DataChunk.selection`   
+#   should fully cover one or more `Chunks`  in the HDF5 file to avoid repeat updates to the same
+#   `Chunks` in the HDF5 file. The `buffer` of the iterator then collects multiple 
+#   :py:class:`~hdmf.data_utils.DataChunk` objects to help reduce the number of small I/O operations 
+#   and help improve performance. The `buffer` should usually be large, i.e, be able to use
+#   as much free RAM as can be safely used.
```
CodyCBakerPhD:
```suggestion
###############################################################################
# .. note::
#   There is some overlap here in nomenclature between HDMF and HDF5. The term *chunk* in both
#   HDMF and HDF5 refer to a subset of dataset, however, in HDF5 a chunk is a piece of dataset on disk,
#   whereas in the context of the  :py:class:`~hdmf.data_utils.DataChunk` iteration is a block of data in memory. As such, the 
#   requirements on the shape and size of chunks are different. In HDF5 these chunks are pieces
#   of a dataset that get compressed and cached together, and they should usually be small in size for 
#   optimal performance  (typically 1 MB or less). In contrast, a :py:class:`~hdmf.data_utils.DataChunk` in 
#   HDMF acts as a block of data for writing data to dataset, and spans multiple HDF5 chunks to improve performance. This is achieved by avoiding repeat 
#   updates to the same `Chunk` in the HDF5 file, :py:class:`~hdmf.data_utils.DataChunk` objects for write 
#   should align with `Chunks` in the HDF5 file, i.e., the :py:class:`~hdmf.data_utils.DataChunk.selection`   
#   should fully cover one or more `Chunks`  in the HDF5 file to avoid repeat updates to the same
#   `Chunks` in the HDF5 file. This is what the `buffer` of the :py:class`~hdmf.data_utils.GenericDataChunkIterator` does, which upon each iteration returns a single
#   :py:class:`~hdmf.data_utils.DataChunk` object (by default > 1 GB) that perfectly spans many HDF5 chunks (by default < 1 MB) to help reduce the number of small I/O operations 
#   and help improve performance. In practice, the `buffer` should usually be even larger than the default, i.e, 
#   as much free RAM as can be safely used. 
```

CodyCBakerPhD: Thanks for helping elaborate that; as Ben mentioned, each iteration of the iterators returns a `DataChunk` that spans many HDF5 chunks, which leaves the small I/O operations purely to `h5py` to handle. We tried it the other way a while back (each `DataChunk` perfectly matching the chunk shape of the HDF5 dataset) but found that this way was about 5-10% faster (and vastly simplified the code).

oruebel: Sounds good.

src/hdmf/data_utils.py
```diff
@@ -125,6 +127,272 @@ def maxshape(self):
...
+                product(
+                    *[
+                        chain(range(buffer_shape_axis, max_shape_axis, buffer_shape_axis), [max_shape_axis])
+                        for max_shape_axis, buffer_shape_axis in zip(self.maxshape, self.buffer_shape)
+                    ]
+                ),
+            )
+        )
+
+        if self.display_progress:
```
rly, line 254: The progress bar option is nice but should be tested at least to make sure the code runs without error. This would also help us ensure high test code coverage.

I recommend adding tqdm to `requirements-dev.txt` and adding a test that sets `display_progress=True`. Also add a test on whether "total" is present in `progress_bar_options`.

src/hdmf/data_utils.py
```diff
@@ -125,6 +127,272 @@ def maxshape(self):
...
+                self.progress_bar_options = dict()
+
+            try:
+                from tqdm import tqdm
+
+                if "total" in self.progress_bar_options:
+                    warn("Option 'total' in 'progress_bar_options' is not allowed to be over-written! Ignoring.")
+                    self.progress_bar_options.pop("total")
+                self.progress_bar = tqdm(total=self.num_buffers, **self.progress_bar_options)
+            except ImportError:
```
rly, line 265: Since it is not so easy to test both when tqdm is importable and when it is not, I recommend adding this comment which removes this section from consideration when computing code coverage.
```suggestion
            except ImportError:  # pragma: no cover
```

src/hdmf/data_utils.py
```diff
@@ -125,6 +127,272 @@ def maxshape(self):
...
+        return self
+
+    def __next__(self):
+        """
+        Retrieve the next DataChunk object from the buffer, refilling the buffer if necessary.
+
+        :returns: DataChunk object with the data and selection of the current buffer.
+        :rtype: DataChunk
+        """
+        if self.display_progress:
```
rly, line 347:
```suggestion
        if self.display_progress:  # pragma: no branch
```

src/hdmf/data_utils.py
```diff
@@ -125,6 +127,272 @@ def maxshape(self):
...
+            self.progress_bar.update(n=1)
+        try:
+            buffer_selection = next(self.buffer_selection_generator)
+            return DataChunk(data=self._get_data(selection=buffer_selection), selection=buffer_selection)
+        except StopIteration:
+            if self.display_progress:
```
rly, line 353:
```suggestion
            if self.display_progress:  # pragma: no branch
```
This will make the code coverage tool ignore the fact that the else condition is never encountered.

src/hdmf/data_utils.py
rly: I cannot seem to figure out how to run the GitHub Action that runs the coverage tool on this PR from your fork. From what I can tell, it should have been run, or present me with the option to approve the workflow.

rly (dismissed):

src/hdmf/data_utils.py
rly: I think I got the GH Action to run such that the coverage report from your fork is uploaded to the HDMF codecov page. You should be able to view the coverage report here: https://app.codecov.io/gh/hdmf-dev/hdmf/commit/e229b771bec984fda7affab75090f5c0d82dae43
and browse which lines of `data_utils.py` have changed in code coverage.

You can also run
```
python -m coverage run test.py -u
coverage report -m  # shell output
coverage html -d tests/coverage/htmlcov  # html output
```
locally and view a local version of this report.

CodyCBakerPhD: Thanks for the info!

src/hdmf/data_utils.py
CodyCBakerPhD: No need according to my local coverage reports; this gets hit by the new tests.

src/hdmf/data_utils.py
CodyCBakerPhD: No need according to my local coverage reports; this gets hit by the new tests.

CodyCBakerPhD:
@rly Thanks for the suggestions; I added full tests for the progress bar using a local temp file (you can pass `file=open_io` as an option to `tqdm`, which makes it easy). Also added new tests to get 100% coverage, according to the local runs. I don't see the uploads on codecov (https://app.codecov.io/gh/hdmf-dev/hdmf/branch/generic_data_chunk) to confirm this, though.

Only one being skipped by the `pragma` trick is the lazy import. For that kind of thing, on `nwb-conversion-tools`, we run two batteries of tests; partial, which uses minimal install and (will one day) check loading assertions on classes/functions that do lazy imports like that, and the full tests that install all the additional packages and actually run every single test without skips. For coverage, I believe there are ways to merge reports so the plan is to take the coverage from each battery and put them together. Haven't figure that out yet though, if you're interested I'll let you know in the future.

rly (approved):

rly:
@CodyCBakerPhD thanks for the changes! I went ahead and updated the PR to:
1) update tests to work with and without `tqdm` and removed the `pragma` trick
1) run CI tests with and without `tqdm`
1) run the coverage GH Actions workflow with and without `tqdm`
1) describe the new `requirements-opt.txt` file in the documentation
1) fix minor flake8 and documentation warnings/errors

I know there is a way to merge coverage reports locally, but here I am running them as separate jobs and uploading each coverage report to codecov.io which then merges the reports. That is sufficient for our purposes. (Also, there could (but is not currently) code that runs only on windows or only on linux where the only way to merge reports is via an external service like codecov.io.)

This PR looks good to me!

rly:
I'll leave this open for a day in case anyone else wants to comment. Then I'll squash and merge it.

CodyCBakerPhD:
> and uploading each coverage report to codecov.io which then merges the reports.

I did not know it did that! That's super useful, thanks for letting me know. We do have some code that only runs on certain OS so having cross-platform coverage uploaded will be great.

Issue #674, opened 2021-10-27
jsonschema is limited to <4 which seems unnecessary.


TheChymera:
## Description

The test suite passes (`1210 passed, 9 skipped, 1 xfailed, 41 warnings in 41.19s`) with the exact same numbers whether I have jsonschema `3.2.0` or the more recent `4.1.0`.

Is this safe to remove?

## Steps to Reproduce

```
pytest -vv
```

## Environment

    Python Executable: Python
    Python Version: 3.9
    Operating System: Linux
    HDMF Version: 3.1.1

oruebel:
I think upgrading the jsonschema version should probably be save. Feel free to submit a PR for this.

PR #675, opened 2021-10-29
Works with newer jsonschema

TheChymera:
As discussed in https://github.com/hdmf-dev/hdmf/issues/674

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ reqs = [
     'scipy>=1.1,<2',
     'pandas>=1.0.5,<2',
     'ruamel.yaml>=0.16,<1',
-    'jsonschema>=2.6.0,<4',
+    'jsonschema>=2.6.0',
     'setuptools',
 ]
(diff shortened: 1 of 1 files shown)


rly:
Since new versions of json-schema, the schema, may introduce [breaking changes](https://json-schema.org/draft/2019-09/release-notes.html#incompatible-changes), and jsonschema, the package, may also introduce breaking changes in the API or its support of a newer incompatible schema, I would prefer to set a maximum jsonschema package version. This is a conservative approach so that we do not get bug reports saying hdmf doesn't work out of the box because of a new incompatible dependency. The flip side to this approach is that we need to regularly monitor (or get feedback from users) when a new major version of a dependency is OK to use and we should relax the requirements.

rly:
I've increased the max major version of jsonschema to 4 in #681 so we can close this.

rly:
Thank you for the contribution!

PR #686, opened 2021-12-03
Er ontology

mavaylon1:
## Motivation

New Ontology class to create and the facilitate the use of ontologies. 

## How to test the behavior?
```
```
Unit tests

## Checklist


docs/gallery/ontology.py
```diff
@@ -0,0 +1,130 @@
...
+###############################################################################
+# Creating a LocalOntology
+# ------------------------------------------------------
+# Users are able to create their own ontologies that will be stored locally on
+# their machine with :py:class:`~hdmf.ontology.LocalOntology`. Users will define
+# the entities of the ontology within a python dictionary. To add new entries,
+# it is required that the entries have both the entity id and the corresponding uri.
+# Once added, users can add or remove entries with built-in methods.
+# :py:class:`~hdmf.ontology.LocalOntology` also supports the ability to write
+# and read the ontology to and from a YAML file.
```
bendichter: Can you make this in second person? Instead of "Users are able to create their own ontologies that will be stored locally...", I would prefer "Create your own ontologies that will be stored locally..."

docs/gallery/ontology.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Let's create an instance of :py:class:`~hdmf.ontology.LocalOntology`.
+ontology = LocalOntology(version='1.0', ontology_name='ontology_name', ontology_uri='ontology_uri')
```
bendichter: Can you replace these values with real values instead of placeholders?

bendichter: You need to import this.

src/hdmf/container.py
```diff
@@ -187,13 +188,17 @@ def __new__(cls, *args, **kwargs):
...
+            {'name': 'ontology', 'type': Ontology,
+             'doc': 'Optional ontology solely for thr purpose of controlling data values',
+             'default': None})
     def __init__(self, **kwargs):
         name = getargs('name', kwargs)
         if '/' in name:
             raise ValueError("name '" + name + "' cannot contain '/'")
         self.__name = name
         self.__field_values = dict()
+        self.ontology = kwargs['ontology']
```
rly: It does not make sense to have an ontology on a container. Which attribute or dataset of the container is being associated with this ontology? I think it does make sense to have it as a property of Data though. So I would recommend moving this there.

rly:
I recommend merging dev into this branch so that recent changes can be incorporated since it has been a while.

PR #688, opened 2022-01-04
Adding MeasurementData Class

CodyCBakerPhD:
## Motivation

This is the first attempt at incorporating conversion and offset values at a more fundamental level in `hdmf` to allow the `data` of a `VectorData` field to be stored in some arbitrary `dtype`, but with intention to reflect true values in some scientific unit, with the mapping between the two given by `data_in_units = data_in_vector * conversion + offset` (still discussing if offset should be applied first or not, so might change before we're done here).

This sort of thing has been very commonly used for heavy data downstream in `pynwb` for particular objects, but we figured this is general enough to be of use at a higher level.

Example: I store my electrical (Volts) data in a dataset as `int16`, and attach a conversion factor of `0.195` and offset of `-6389.76`. To coerce to scientific units, simply apply the equation given above to any selection of data and voila; we can store the data in a fundamental base type with an itemsize of 2 bytes per value instead of saving the entire array as `floats` (itemsize of 8 bytes per value).

## How to test the behavior?

'Tests' are added, but the class and therefore tests are pretty basic so not sure what else to include. By intention, the data is not auto-scaled on retrieval from the object in the `hdf5` file, it's just the intention that these attributes always be stored directly adjacent to the `data` field for easy access. 

## Checklist


CodyCBakerPhD:
@rly @oruebel I do get errors (and I presume they will show on CI) regarding the class not being present in the `hdmf-common-schema`, which makes sense since I didn't change that at all here to reflect the draft PR over there (https://github.com/hdmf-dev/hdmf-common-schema/pull/63) - how do I adjust this draft PR here to properly point to the draft schema over there?

PR #719, opened 2022-04-12
Added functionality to load plugins for HDMF 

oruebel:
## Motivation

I am trying to create a new data backend for HDMF and would like to define it as a separate library outside of HDMF. However, since the backend uses a different type of array data type that is not part of the default ``hdmf.utils.__macros`` definition, I need to modify the definition to add the custom array type. The challenge is that once HDMF is loaded, updating the ``hdmf.utils.__macros``does not propagate such that, e.g., classes like ``DataIO`` or ``DatasetBuilder`` will still not accept the additional array types.  

One solution to this problem is to make the updates to ``hdmf.utils.__macros`` before HDMF is fully loaded. This is the approach that this PR implements by defining the concept of plugins for HDMF. To do this, this PR adds:

* ``hdmf.utils.get_available_plugins`` to find all available plugins
* ``hdmf.utils.configure_macro_plugins`` which is called during the import of `hdmf.utils``. The function iterates over all plugins to collect and implement all modifications to ``hdmf.utils.__macros``

The plugin mechanism is based on a simple naming convention, i.e., plugins for HDMF must start with ``hdmf_`` and end with ``_plugin``. To define changes to the ``hdmf.utils.__macros`` the plugin must the implement the ``configure_hdmf_docval_macros`` function which simply returns a list of modifications to be made to ``hdmf.utils.__macros``. 

The main question is: a) is this the best approach to do this and b) is there an alternate approach that would allow modifications of ``hdmf.utils.__macros``  and have HDMF still honor those changes in docval?

## Checklist

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -60,10 +60,13 @@ def configure_macro_plugins():
         if hasattr(plugin, 'configure_hdmf_docval_macros'):
             update_macros = getattr(plugin, 'configure_hdmf_docval_macros')()
             if update_macros is not None:
+                # update_macros is a list of tuples with the changes
                 for macro in update_macros:
+                    # create a new macro
                     if macro[0] not in __macros:
                         __macros[macro[0]] = list()
                         __macros[macro[0]].append(macro[1])
+                    # update an existing macro by adding a new allowable type
                     else:
                         if macro[1] not in __macros[macro[0]]:
                             __macros[macro[0]].append(macro[1])
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -8,12 +8,68 @@ from enum import Enum
 import h5py
 import numpy as np
 
-
 __macros = {
     'array_data': [np.ndarray, list, tuple, h5py.Dataset],
     'scalar_data': [str, int, float, bytes, bool],
     'data': []
 }
+
+
+def get_available_plugins():
+    """
+    Discover available plugin modules.
+
+    HDMF plugins must use the following naming convention. The package name must start with "hdmf_"
+    and end with "_plugin"
+
+    :returns: Dictionary where the key is the name of the discovered plugin module and the value is
+              the module itself
+
+    """
+    import importlib
+    import pkgutil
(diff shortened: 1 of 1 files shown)


rly:
This approach seems reasonable to me. We might be able to defer the resolution of the docval macro until each time the decorated function is called and then update the function double-underscore variables accordingly, but that might be tedious and inefficient. See https://github.com/hdmf-dev/hdmf/pull/720 for an attempt at it.

I did not realize there was such a strong dependence on order of imports in making sure the docval macros are up-to-date. We should add a "NOTE" comment about that in the `docval_macro` function saying that this decorator must be used in classes that are loaded before any classes that might use the macro.

oruebel:
> This approach seems reasonable to me.

Thanks for taking a look! I'll clean this up and add tests and let you know once its ready. 

> I did not realize there was such a strong dependence on order of imports

Inside HDMF we are using the ``@docval_macro`` decorator to register classes, e.g., DataChunkIterator with the macros. Within HDMF this works I think because those are processed during the import of HDMF. However, when we are outside of HDMF, the module gets imported and then changing the macro doesn't propagate to the modules/classes that Python has already imported. I.e., as long as things are within HDMF itself, the current approach works. That is why I didn't have any issues when we created the backend as part of HDMF itself. However, now that I'm trying to pull the hdmf-zarr backend out of hdmf it triggers this tricky problem. 

> We might be able to defer the resolution of the docval macro ...

I think that should be doable but I agree that would probably make things less efficient. To be honest, I think this is a very particular issue that in practice will be rare so I would error on the side of keeping changes to HDMF as minimal as possible.

oruebel:
After some discussion we decided to add zarr.Array to the docval macro directly in HDMF if Zarr is installed. This has the advantage that: 1) it will allow users that have data in Zarr to convert there data more easily with HDMF (independent of the HDMF Zarr backend), 2) will require only minimal changes in HDMF, 3) introduces a purely optional dependency only, 4) keeps the implementation of the hdmf-zarr backend simple. 

Closing this PR and will create a new PR with the new approach.

PR #730, opened 2022-05-10
Warn when using positional args in data type class init

rly:
## Motivation

First step toward resolving https://github.com/NeurodataWithoutBorders/pynwb/issues/1189 . See that issue ticket for more motivation and background.

This PR makes `__init__` of all HDMF-common data types raise a warning when a user passes positional arguments instead of keyword arguments. This includes auto-generated classes. In the process, I replaced `call_docval_func(super().__init__, kwargs)` with simply `super().__init__(**kwargs)`. The difference is that the former allows extra arguments in `kwargs` which are filtered out by `call_docval_func` and the latter requires that you pop out any extra arguments first. This is IMO better because it is more explicit and is one less iteration through and parsing of `kwargs`.

The `__init__` of base types `AbstractContainer`, `Container`, `Data`, and `Table` are unaffected. We can revisit adding a warning to those.

We have also discussed adding this warning to `DynamicTable.add_row`. That could be done in a separate PR.

In a future major release, we can change the warning to an error. 

## Checklist


oruebel (approved):
Looks like PyNWB test will need to be updated as some of the tests that check for warning fail since this PR adds new warnings

rly:
Yes, PyNWB - both source code and tests - will have to be updated. There are lots of tests using positional arguments...

rly:
@bendichter @ajtritt I would appreciate your feedback on this change! This does not change any functionality but adds warnings and starts us on a path toward changed functionality.

ajtritt (approved):

PR #743, opened 2022-06-22
Improve error message when using GDCI and chunk shape > dataset size

rly:
## Motivation

When using ``GenericDataChunkIterator`` and passing a chunk shape that happens to be larger than the dataset size, e.g., the chunk shape = (2000, 10) and the dataset size, or max shape, is (1000, 10), then an error is raised with the message "Some dimensions of chunk_shape ((2000, 10)) exceed the manual buffer shape ((1000, 10))". This is confusing because I did not set the buffer shape. 

This PR adds a check for chunk shape > dataset size and provides a error message that more clearly indicates the problem.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 3.3.2 (June 23, 2022)
+
+### Bug fixes
+- Fix error message when using ``GenericDataChunkIterator`` and chunk shape is larger than dataset size. @rly ()
+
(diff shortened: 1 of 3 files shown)


bendichter (approved):

CodyCBakerPhD:
Ah, thanks for catching that. I'd also be open to just setting `chunk_shape` and `buffer_shape` to the `maxshape` whenever `maxsize < chunk_size`, too. But maybe that can be done in the future.

PR #746, opened 2022-07-13
S3fs

bendichter:
## Motivation

Add suport for s3fs

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # HDMF Changelog
 
 ## HDMF 3.3.3 (Upcoming)
-- Relax input validation of `HDF5IO` to allow for s3fs support. The `mode` arg was given a default value of "r" and the `file` arg can now accept an `S3File` type argument. @bendichter ([#746](https://github.com/hdmf-dev/hdmf/pull/746))
+- Relax input validation of `HDF5IO` to allow for s3fs support. Existing arguments of `HDF5IO` are modified as follows: i) `mode` was given a default value of "r", ii) `path` was given a default value of `None`, and iii) `file` can now accept an `S3File` type argument. @bendichter ([#746](https://github.com/hdmf-dev/hdmf/pull/746))
 
 ## HDMF 3.3.2 (June 27, 2022)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # HDMF Changelog
 
+## HDMF 3.3.3 (Upcoming)
+- Relax input validation of `HDF5IO` to allow for s3fs support. The `mode` arg was given a default value of "r" and the `file` arg can now accept an `S3File` type argument. @bendichter ([#746](https://github.com/hdmf-dev/hdmf/pull/746))
+
 ## HDMF 3.3.2 (June 27, 2022)
 
 ### Bug fixes
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -36,13 +36,14 @@ class HDF5IO(HDMFIO):
 
     __ns_spec_path = 'namespace'  # path to the namespace dataset within a namespace group
 
-    @docval({'name': 'path', 'type': (str, Path), 'doc': 'the path to the HDF5 file'},
+    @docval({'name': 'path', 'type': (str, Path), 'doc': 'the path to the HDF5 file', 'default': None},
             {'name': 'manager', 'type': (TypeMap, BuildManager),
              'doc': 'the BuildManager or a TypeMap to construct a BuildManager to use for I/O', 'default': None},
             {'name': 'mode', 'type': str,
              'doc': ('the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-", "x"). '
                      'See `h5py.File <http://docs.h5py.org/en/latest/high/file.html#opening-creating-files>`_ for '
-                     'more details.')},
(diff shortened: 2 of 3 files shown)


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -36,16 +36,17 @@ class HDF5IO(HDMFIO):
...
                      'See `h5py.File <http://docs.h5py.org/en/latest/high/file.html#opening-creating-files>`_ for '
-                     'more details.')},
+                     'more details.'),
+             'default': 'r'},
+            {'name': 'manager', 'type': (TypeMap, BuildManager),
+             'doc': 'the BuildManager or a TypeMap to construct a BuildManager to use for I/O', 'default': None},
             {'name': 'comm', 'type': 'Intracomm',
              'doc': 'the MPI communicator to use for parallel I/O', 'default': None},
-            {'name': 'file', 'type': File, 'doc': 'a pre-existing h5py.File object', 'default': None},
+            {'name': 'file', 'type': [File, "S3File"], 'doc': 'a pre-existing h5py.File object', 'default': None},
```
oruebel, line 49: I didn't know that passing just the name of the class as part of the type  argument (here ``"S3File"``) was allowed. I always thought we always had to directly provided all types. But looking at https://github.com/hdmf-dev/hdmf/blob/0e50852145bbf9a4676b6edc936317f882f9a47e/src/hdmf/utils.py#L99 in the type checking function that ``docval`` uses, this seems to be indeed valid. The type checking is probably not quite as strict (since it is purely based on name, not actual type) but seems (for all intended purposes) sufficient for this purpose and avoids creating a dependency on S3fs here. Thanks for teaching me something new @bendichter

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -54,10 +55,13 @@ def __init__(self, **kwargs):
         path, manager, mode, comm, file_obj, driver = popargs('path', 'manager', 'mode', 'comm', 'file', 'driver',
                                                               kwargs)
 
+        if path is None and file is None:
+            raise ValueError("You must supply either a path or a file.")
```
oruebel, line 59: Can you please add a unit test for this additional error case. Otherwise this PR looks good to me.

src/hdmf/backends/hdf5/h5tools.py
bendichter: I think if a user creates a class called "S3File" to get around input validation they have voided their warranty 😛

oruebel: Agreed

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # HDMF Changelog
 
+## HDMF 3.3.3 (Upcoming)
+- Relax input validation of `HDF5IO` to allow for s3fs support. The `mode` arg was given a default value of "r" and the `file` arg can now accept an `S3File` type argument. @bendichter ([#746](https://github.com/hdmf-dev/hdmf/pull/746))
```
```suggestion
- Relax input validation of `HDF5IO` to allow for s3fs support. Existing arguments of `HDF5IO` are modified as follows: i) `mode` was given a default value of "r", ii) `path` was given a default value of `None`, and iii) `file` can now accept an `S3File` type argument. @bendichter ([#746](https://github.com/hdmf-dev/hdmf/pull/746))
```

oruebel (dismissed):
Looks good to me

bendichter:
@oruebel I need an official approval to merge it

oruebel (approved):

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -36,16 +36,17 @@ class HDF5IO(HDMFIO):
...
-             'doc': 'the BuildManager or a TypeMap to construct a BuildManager to use for I/O', 'default': None},
+    @docval({'name': 'path', 'type': (str, Path), 'doc': 'the path to the HDF5 file', 'default': None},
             {'name': 'mode', 'type': str,
              'doc': ('the mode to open the HDF5 file with, one of ("w", "r", "r+", "a", "w-", "x"). '
                      'See `h5py.File <http://docs.h5py.org/en/latest/high/file.html#opening-creating-files>`_ for '
-                     'more details.')},
+                     'more details.'),
+             'default': 'r'},
+            {'name': 'manager', 'type': (TypeMap, BuildManager),
+             'doc': 'the BuildManager or a TypeMap to construct a BuildManager to use for I/O', 'default': None},
```
rly, line 46: Moving this argument in the argument list for `HDF5IO` is technically a breaking change, though a very minor one. I agree with making the change because `HDF5IO(path, mode, manager, ...)` follows convention better than `HDF5IO(path, manager, mode)`. But we should probably bump HDMF to 3.4.0 in this case.

rly: Actually never mind. I see what happened here. "mode" did not have a default value before, so it would be parsed as `HDF5IO(path, mode, manager, ...)`. Ignore these comments.

bendichter: Actually, I don't think it is a breaking change. At least the intention is that it isn't. This is a quark of docval- args with no default go first in positional arguments even if they aren't listed first in docval. Once I added default values for everything, I needed to move "manager" later in the list so that it *didn't* cause a breaking change

bendichter: I'm glad you brought this up! I should have mentioned this in the PR comments

PR #747, opened 2022-07-19
Setup HDF5 dataset without having data ahead of time

ajtritt:
## Motivation

Add the ability to create a dataset on disk and get access to the handle to the dataset _without_ having to close and repoen a file. 

## How to test the behavior?
```python
dataio = H5DataIO(shape=(5,), dtype=int)
foo = Foo('foo1', dataio, "I am foo1", 17, 3.14)
bucket = FooBucket('bucket1', [foo])
foofile = FooFile(buckets=[bucket])

io = HDF5IO(self.path, manager=self.manager, mode='w')
io.write(foofile)

foo.my_data.dataset[:] = [0, 1, 2, 3, 4]
io.close()
```

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -478,11 +478,11 @@ class H5DataIO(DataIO):
              'default': False},
             {'name': 'shape',
              'type': tuple,
-             'doc': 'the shape of the new dataset',
+             'doc': 'the shape of the new dataset, used only if data is None',
              'default': None},
             {'name': 'dtype',
              'type': (str, type, np.dtype),
-             'doc': 'the data type of the new dataset',
+             'doc': 'the data type of the new dataset, used only if data is None',
              'default': None}
             )
     def __init__(self, **kwargs):
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -475,11 +475,20 @@ class H5DataIO(DataIO):
             {'name': 'allow_plugin_filters',
              'type': bool,
              'doc': 'Enable passing dynamically loaded filters as compression parameter',
-             'default': False}
+             'default': False},
+            {'name': 'shape',
+             'type': tuple,
+             'doc': 'the shape of the new dataset',
+             'default': None},
+            {'name': 'dtype',
+             'type': (str, type, np.dtype),
+             'doc': 'the data type of the new dataset',
+             'default': None}
             )
     def __init__(self, **kwargs):
         # Get the list of I/O options that user has passed in
-        ioarg_names = [name for name in kwargs.keys() if name not in ['data', 'link_data', 'allow_plugin_filters']]
+        ioarg_names = [name for name in kwargs.keys() if name not in ['data', 'link_data', 'allow_plugin_filters',
+                                                                      'dtype', 'shape']]
         # Remove the ioargs from kwargs
         ioarg_values = [popargs(argname, kwargs) for argname in ioarg_names]
         # Consume link_data parameter
@@ -494,6 +503,9 @@ class H5DataIO(DataIO):
         call_docval_func(super().__init__, kwargs)
         # Construct the dict with the io args, ignoring all options that were set to None
         self.__iosettings = {k: v for k, v in zip(ioarg_names, ioarg_values) if v is not None}
+        if self.data is None:
+            self.__iosettings['dtype'] = self.dtype
+            self.__iosettings['shape'] = self.shape
         # Set io_properties for DataChunkIterators
         if isinstance(self.data, AbstractDataChunkIterator):
             # Define the chunking options if the user has not set them explicitly.
@@ -525,6 +537,18 @@ class H5DataIO(DataIO):
         if isinstance(self.data, Dataset):
             for k in self.__iosettings.keys():
                 warnings.warn("%s in H5DataIO will be ignored with H5DataIO.data being an HDF5 dataset" % k)
+
+        self.__dataset = None
+
+    @property
+    def dataset(self):
+        return self.__dataset
+
+    @dataset.setter
+    def dataset(self, val):
+        if self.__dataset is not None:
+            raise ValueError("Cannot overwrite H5DataIO.dataset")
+        self.__dataset = val
 
     def get_io_params(self):
         """
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -1040,9 +1040,11 @@ class HDF5IO(HDMFIO):
             return None
         name = builder.name
         data = builder.data
+        dataio = None
         options = dict()   # dict with additional
         if isinstance(data, H5DataIO):
             options['io_settings'] = data.io_settings
+            dataio = data
             link_data = data.link_data
             data = data.data
         else:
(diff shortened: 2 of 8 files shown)


src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -475,11 +475,20 @@ class H5DataIO(DataIO):
             {'name': 'allow_plugin_filters',
              'type': bool,
              'doc': 'Enable passing dynamically loaded filters as compression parameter',
-             'default': False}
+             'default': False},
+            {'name': 'shape',
+             'type': tuple,
+             'doc': 'the shape of the new dataset',
```
rly:
```suggestion
             'doc': 'the shape of the new dataset, used only if data is None',
```

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -475,11 +475,20 @@ class H5DataIO(DataIO):
...
+             'default': None},
+            {'name': 'dtype',
+             'type': (str, type, np.dtype),
+             'doc': 'the data type of the new dataset',
```
rly:
```suggestion
             'doc': 'the data type of the new dataset, used only if data is None',
```

src/hdmf/data_utils.py
```diff
@@ -930,6 +934,16 @@ def data(self, val):
...
+    @property
+    def dtype(self):
+        """Get the wrapped data object"""
+        return self.__dtype
+
+    @property
+    def shape(self):
+        """Get the wrapped data object"""
+        return self.__shape
+
```
rly: This will break backward compatibility because currently `dtype` and `shape` are deferred to `self.data.dtype` and `self.data.shape`. I suggest one of these:
1. create a subclass of `DataIO` for empty datasets where `dtype` and `shape` are defined this way and use that
2. add an if statement in these: `if self.__dtype is not None, return self.__dtype, else, return self.__getattr__("dtype")`
3. change the name of these properties to something like `dtype_to_set` and `shape_to_set`

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -1577,7 +1577,7 @@ def test_link(self):
 
             self.assertTrue(self.foo2.my_data.valid)  # test valid
             self.assertEqual(len(self.foo2.my_data), 5)  # test len
-            self.assertEqual(self.foo2.my_data.shape, (5,))  # test getattr with shape
+            self.assertEqual(self.foo2.my_data.size, (5,))  # test getattr with size
```
rly: I am surprised this works. When I put a breakpoint here, I can see that `self.foo2.my_data.size` is 5 ~~, and self.assertEqual(self.foo2.my_data.size, (5,)) raises an AssertionError~~. So why does this pass???

rly:
```
            self.assertEqual(self.foo2.my_data.size, (5,))  # test getattr with size
            self.assertEqual(self.foo2.my_data.size, 5)  # test getattr with size
>           self.assertEqual(5, (5,))  # test getattr with size
E           AssertionError: 5 != (5,)
```
That's weird.

rly:
```python
type(self.foo2.my_data.size)  # np.int64
self.assertEqual(np.int64(5), (5,))  # True
```

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -494,6 +503,9 @@ def __init__(self, **kwargs):
         call_docval_func(super().__init__, kwargs)
         # Construct the dict with the io args, ignoring all options that were set to None
         self.__iosettings = {k: v for k, v in zip(ioarg_names, ioarg_values) if v is not None}
+        if self.data is None:
+            self.__iosettings['dtype'] = self.dtype
+            self.__iosettings['shape'] = self.shape
```
oruebel, line 508: Since ``dtype`` and ``shape`` are only allowed if ``data`` is ``None`` I would suggest adding a check here to raise a ``ValueError`` if the user tries to provide ``data`` and change the ``dtype`` or ``shape``.  We should then also add a unit test to check for those errors. 

```suggestion
        if self.data is None:
            self.__iosettings['dtype'] = self.dtype
            self.__iosettings['shape'] = self.shape
        else:
            if getargs('dtype', kwargs) is not None:
                 raise ValueError("Setting the dtype when data is not None is not supported")
            if  getargs('shape', kwargs) is not None:
                 raise ValueError("Setting the shape when data is not None is not supported")
```

oruebel (commented):
Overall this looks good to me.

src/hdmf/backends/hdf5/h5_utils.py
ajtritt: @oruebel I moved this to DataIO. What do you think?

oruebel: Sounds reasonable. In that case I would update the doc for those parameters in `DataIO` as well to mention this behavior.

src/hdmf/data_utils.py
```diff
@@ -907,10 +907,19 @@ class DataIO:
     used to pass dataset-specific I/O parameters to the particular HDMFIO backend.
     """
 
-    @docval({'name': 'data', 'type': 'array_data', 'doc': 'the data to be written', 'default': None})
+    @docval({'name': 'data', 'type': 'array_data', 'doc': 'the data to be written', 'default': None},
+            {'name': 'dtype', 'type': (type, np.dtype), 'doc': 'the data type of the dataset', 'default': None},
+            {'name': 'shape', 'type': tuple, 'doc': 'the shape of the dataset', 'default': None})
```
oruebel:
```suggestion
    @docval({'name': 'data', 
             'type': 'array_data', 
             'doc': 'the data to be written', 
             'default': None},
            {'name': 'dtype', 
             'type': (type, np.dtype), 
             'doc': 'the data type of the dataset. Not used if data is specified.', 
             'default': None},
            {'name': 'shape', 
             'type': tuple, 
             'doc': 'the shape of the dataset. Not used if data is specified.', 
             'default': None})
```

oruebel (dismissed):

oruebel (dismissed):

tests/unit/test_io_hdf5_h5tools.py
rly: This is probably why. A little gotcha when comparing numpy ints with other types
```python
>>> 5 == (5, )
False
>>> np.int64(5) == (5, )
array([ True])
>>> bool(np.int64(5) == (5,))
True
```

```suggestion
            self.assertEqual(self.foo2.my_data.shape, (5,))  # test getattr with shape
```
Can we revert this change?

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -1595,8 +1595,8 @@ def test_link(self):
         with self.assertRaisesWith(InvalidDataIOError, "Cannot get length of data. Data is not valid."):
             len(self.foo2.my_data)
 
-        with self.assertRaisesWith(InvalidDataIOError, "Cannot get attribute 'shape' of data. Data is not valid."):
-            self.foo2.my_data.shape
+        with self.assertRaisesWith(InvalidDataIOError, "Cannot get attribute 'size' of data. Data is not valid."):
+            self.foo2.my_data.size
```
```suggestion
        with self.assertRaisesWith(InvalidDataIOError, "Cannot get attribute 'shape' of data. Data is not valid."):
            self.foo2.my_data.shape
```
And this change.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1319,6 +1325,35 @@ def __setup_chunked_dset__(cls, parent, name, data, options=None):
...
+            raise ValueError(f"Cannot setup empty dataset {pp(parent.name, name)} without shape")
+        if 'dtype' not in io_settings:
+            raise ValueError(f"Cannot setup empty dataset {pp(parent.name, name)} without dtype")
+        if isinstance(io_settings['dtype'], str):
+            # map to real dtype if we were given a string
+            io_settings['dtype'] = cls.__dtypes.get(io_settings['dtype'])
+        try:
+            dset = parent.create_dataset(name, **io_settings)
+        except Exception as exc:
+            raise Exception("Could not create dataset %s in %s" % (name, parent.name)) from exc
```
rly, line 1354: Can you add a test for this exception, e.g., you try to create an empty dataset with a name, and the dataset already exists or dtype=object.

rly:
Please fix flake8 changes and add a note to the changelog

rly (approved):

rly:
Thanks. LGTM!

PR #752, opened 2022-08-04
Mark `call_docval_func` and `fmt_docval_args` with PendingDeprecationWarning

### LINKED ISSUE #748: [Bug]: Change behavior of call_docval_func
### What happened?

Any code that uses `call_docval_func` on a Container's `__init__` class often results in warnings that using positional arguments will be deprecated soon. This is beavuse `call_docval_func` takes in a function and keyword arguments to pass to the function, splits the arguments into positional arguments and keyword arguments based on whether the docval on the function says that the arguments is required or has a default value, and then calls the function, passing positional and keyword arguments. 

I see no reason to take keyword arguments being passed to a function and remove the keyword before calling the function. 

### Steps to Reproduce

```shell
n/a
```

### Traceback

```shell
n/a
```

### Operating System

Windows

### Python Executable

Python

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

rly:
## Motivation

Fix #748.

I also removed all calls to `call_docval_func` and `fmt_docval_args` which were unnecessary in all cases.

`call_docval_func` and `fmt_docval_args` are unnecessary and add complexity in stacktraces and debugging. Their primary purpose is to 1) remove arguments that are not listed in a function's docval before calling the function and 2) parse kwargs into args and kwargs before calling the function (which then parses those back into kwargs -- unnecessary, see #748).

I believe point 1. is also unnecessary and hides potential errors. For example, if function `func` accepts args "a", "b", "c" and calls `super().func` passing all args, and `super().func` accepts only "a", then it may be convenient to strip "b" and "c" transparently, but this is not typical Python behavior and a naive user may think "b" and "c" are used by `super().func` since they are being passed to it. I think it is better for code readability for the user to have to pop "b" and "c" from kwargs explicitly before calling `super().func` on the remaining kwargs (just "a"). Either that or `call_docval_func` should be renamed to something like `call_docval_func_without_extra_args` or something to be clearer.

@oruebel @ajtritt what do you think?

## Checklist

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -429,6 +429,7 @@ def fmt_docval_args(func, kwargs):
             ret_kwargs.update(kwargs_copy)
     else:
         raise ValueError('no docval found on %s' % str(func))
+    warnings.warn("hello")
     return ret_args, ret_kwargs
 
 
@@ -464,7 +465,13 @@ def call_docval_func(func, kwargs):
     warnings.warn("call_docval_func will be deprecated in a future version of HDMF. Please note that call_docval_func "
                   "transparently removes all arguments not accepted by the function's docval.",
                   PendingDeprecationWarning)
-    fargs, fkwargs = fmt_docval_args(func, kwargs)
+    with warnings.catch_warnings(record=True):
+        # catch and ignore only PendingDeprecationWarnings from fmt_docval_args so that two
+        # PendingDeprecationWarnings saying the same thing are not raised
+        warnings.simplefilter("ignore", UserWarning)
+        warnings.simplefilter("always", PendingDeprecationWarning)
+        fargs, fkwargs = fmt_docval_args(func, kwargs)
+
     return func(*fargs, **fkwargs)
 
 
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -410,8 +410,11 @@ def fmt_docval_args(func, kwargs):
 
     Useful for methods that wrap other methods
     '''
-    warnings.warn("fmt_docval_args will be deprecated in a future version of HDMF. Please note that fmt_docval_args "
-                  "transparently removes all arguments not accepted by the function's docval.",
+    warnings.warn("fmt_docval_args will be deprecated in a future version of HDMF. Instead of using fmt_docval_args, "
+                  "call the function directly with the kwargs. Please note that fmt_docval_args "
+                  "removes all arguments not accepted by the function's docval, so if you are passing kwargs that "
+                  "includes extra arguments and the function's docval does not allow extra arguments (allow_extra=True "
+                  "is set), then you will need to pop the extra arguments out of kwargs before calling the function.",
                   PendingDeprecationWarning)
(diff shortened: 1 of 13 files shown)


src/hdmf/utils.py
```diff
@@ -429,7 +432,38 @@ def fmt_docval_args(func, kwargs):
...
+
 def call_docval_func(func, kwargs):
+    """Call the function with only the keyword arguments that are accepted by the function's docval.
+
+    Extra keyword arguments are not passed to the function unless the function's docval has allow_extra=True.
+    """
+    warnings.warn("call_docval_func will be deprecated in a future version of HDMF. Please note that call_docval_func "
+                  "transparently removes all arguments not accepted by the function's docval.",
+                  PendingDeprecationWarning)
     fargs, fkwargs = fmt_docval_args(func, kwargs)
```
oruebel: Calling ``fmt_docval_args(func, kwargs`` inside of ``call_docval_func`` will raise two warnings. We should here catch the warning for ``fmt_docval_args`` and just raise the warning for ``call_docval_func``.

src/hdmf/utils.py
```diff
@@ -429,7 +432,38 @@ def fmt_docval_args(func, kwargs):
...
+#         raise ValueError('No docval found on %s' % str(func))
+
+
 def call_docval_func(func, kwargs):
+    """Call the function with only the keyword arguments that are accepted by the function's docval.
+
+    Extra keyword arguments are not passed to the function unless the function's docval has allow_extra=True.
+    """
+    warnings.warn("call_docval_func will be deprecated in a future version of HDMF. Please note that call_docval_func "
+                  "transparently removes all arguments not accepted by the function's docval.",
```
oruebel: I think many users/developers will be confused by the note about  "transparently removes arguments". Instead, it would be useful to describe how to properly replace ``call_docval_func`` in code so folks know what to do.

src/hdmf/utils.py
```diff
@@ -410,6 +410,9 @@ def fmt_docval_args(func, kwargs):
 
     Useful for methods that wrap other methods
     '''
+    warnings.warn("fmt_docval_args will be deprecated in a future version of HDMF. Please note that fmt_docval_args "
+                  "transparently removes all arguments not accepted by the function's docval.",
+                  PendingDeprecationWarning)
```
oruebel: Same comment as for below for ``call_docval_func``. I think many users/developers will be confused by the note about  . Instead, it would be useful to describe how to properly replace ``fmt_docval_args`` in code so folks know what to do.

oruebel:
> call_docval_func and fmt_docval_args are unnecessary and add complexity 

In general these changes seem fine. However, this will probably require a large number of changes in PyNWB to avoid generating warnings from NWB Container classes.

rly:
> > call_docval_func and fmt_docval_args are unnecessary and add complexity
> 
> In general these changes seem fine. However, this will probably require a large number of changes in PyNWB to avoid generating warnings from NWB Container classes.

I have already removed all of these references in the last release.

oruebel (approved):

PR #771, opened 2022-09-22
Use shape for DataIO length

ajtritt:
## Motivation

When shape is specified in DataIO, length should be obtainable. However, because `data` was not being set, DataIO was declaring itself not valid and raising an error.

This also adds the requirement in H5DataIO that `dtype` and `shape` are specified together. [h5py's `create_dataset`](https://docs.h5py.org/en/stable/high/group.html#h5py.Group.create_dataset) sets `dtype` to 32-bit float by default. I opted to not rely on this default behavior out of concern of creating unintended behavior in hdmf. 

## How to test the behavior?
```python
dataio = H5DataIO(dtype=int, shape=(10, 10))
assert len(dataio) == 10
```

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -489,6 +489,10 @@ class H5DataIO(DataIO):
         # Get the list of I/O options that user has passed in
         ioarg_names = [name for name in kwargs.keys() if name not in ['data', 'link_data', 'allow_plugin_filters',
                                                                       'dtype', 'shape']]
+
+        if kwargs['data'] is None:
+            if kwargs['shape'] is None or kwargs['dtype'] is None:
+                raise ValueError("Must specify 'dtype' and 'shape' if not specifying 'data'")
         # Remove the ioargs from kwargs
         ioarg_values = [popargs(argname, kwargs) for argname in ioarg_names]
         # Consume link_data parameter
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -637,9 +637,3 @@ class H5DataIO(DataIO):
         if isinstance(self.data, Dataset) and not self.data.id.valid:
             return False
         return super().valid
-
-    def __len__(self):
-        maxshape = self.__iosettings.get('maxshape', None)
-        if maxshape is not None and maxshape[0] is not None:
-            return maxshape[0]
-        return super().__len__()
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -633,3 +633,9 @@ class H5DataIO(DataIO):
         if isinstance(self.data, Dataset) and not self.data.id.valid:
             return False
         return super().valid
+
+    def __len__(self):
+        maxshape = self.__iosettings.get('maxshape', None)
+        if maxshape is not None and maxshape[0] is not None:
+            return maxshape[0]
+        return super().__len__()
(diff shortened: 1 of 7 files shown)


src/hdmf/data_utils.py
```diff
@@ -995,6 +995,8 @@ def __deepcopy__(self, memo):
 
     def __len__(self):
         """Number of values in self.data"""
+        if self.__shape is not None:
+            return self.__shape[0]
```
rly, line 1004: Is it possible for `self.__shape` to become out of sync with `self.data`? For example, if the user creates a DataIO without data and then sets data later but sets it to a shape not equal to self.__shape (or a dtype not equal to self.__dtype)?

src/hdmf/common/io/table.py
```diff
@@ -82,6 +82,8 @@ def process_field_spec(cls, classdict, docval_args, parent_cls, attr_name, not_i
             # the spec does not know which table this DTR points to
             # the user must specify the table attribute on the DTR after it is generated
             column_conf['table'] = True
+        else:
+            column_conf['class'] = dtype
```
rly, line 86: I *think* this flow would work for subclasses of `VectorData` but it would be good to test that.

rly:
Besides the comment above, this looks good to me.

src/hdmf/data_utils.py
ajtritt: If I understand L950 correctly, no that's not possible.

ajtritt: And I just added a test to make sure that doesn't happen.

rly: Ah, I see. Looks good. Thanks!

src/hdmf/common/io/table.py
ajtritt: Do you mean for stuff like `EnumData`, `DynamicTableRegion`, etc?

rly: yeah, well not `DynamicTableRegion`, because that's handled above, and I am not sure `EnumData` is fully handled by `__columns__` but I think it is. I was more thinking of custom extensions to VectorData. Making such a test might be a pain though. I think it would be fine but feel free to add a test.

rly (dismissed):

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -489,6 +489,10 @@ def __init__(self, **kwargs):
         # Get the list of I/O options that user has passed in
         ioarg_names = [name for name in kwargs.keys() if name not in ['data', 'link_data', 'allow_plugin_filters',
                                                                       'dtype', 'shape']]
+
+        if kwargs['data'] is None:
+            if kwargs['shape'] is None or kwargs['dtype'] is None:
+                raise ValueError("Must specify 'dtype' and 'shape' if not specifying 'data'")
```
oruebel: Should this check be part of ``DataIO.__init__`` instead? I can't think of a use case where we want to allow this in ``DataIO`` (or any subclass thereof) but not in ``HDF5IO``.

ajtritt: Yes, it has been moved to DataIO

oruebel (dismissed):

src/hdmf/data_utils.py
```diff
@@ -922,7 +922,7 @@ class DataIO:
     def __init__(self, **kwargs):
         data, dtype, shape = popargs('data', 'dtype', 'shape', kwargs)
         if data is None:
-            if dtype is None or shape is None:
+            if (dtype is None) ^ (shape is None):
```
oruebel, line 925: Why use XOR instead of OR here? Should this error not also be raised if both dtype and shape are None?

ajtritt: See the commit message. It's done for backward compatibility i.e. set_dataio

rly (approved):

oruebel (approved):

PR #780, opened 2022-10-28
Added error catcher for potentially unsafe memory operation (GenericDataChunkIterator)

CodyCBakerPhD:
Related to a memory leak detected in https://github.com/catalystneuro/neuroconv/pull/194

Basic explanation is, when the `GenericDataChunkIterator` was initially developed and tested in real-world situations, the data types and scales at that time weren't large enough to cause this problem. What happens is `np.prod(some_iterable_of_ints)` uses `int32` in its product operation (max value of only 2,147,483,648 on the positive end), whereas Python uses `int64` for `*` operations (max value of  9,223,372,036,854,775,808) .

So if we had recursively used `*` like

```python
val=1
for x in some_iterable_of_ints:
    val *= x
```
the data type would have been large enough to avoid the problem.

Anyhow, this was detected initially via the warnings numpy throws when it first happens in the operation, but given that this value determines how much data to load into RAM it's a bit more important of a problem than a warning (it actually killed my entire system, lol...) so I'm catching it here and upgrading to an informative warning that will also let us know in the future if we ever need to push it further than this.

Message appears like

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # HDMF Changelog
 
+## Upcoming
+
+### Bug fixes
+- Fixed an issue with the `data_utils.GenericDataChunkIterator` where if the underlying dataset was such that the `numpy.product` of the `maxshape` exceeded the range of the default `int32`, buffer overflow would occur and cause the true buffer shape to exceed available memory. This has been resolved by upcasting all shape attributes and operations to use the `uint64` data type. @codycbakerphd ([#780](https://github.com/hdmf-dev/hdmf/pull/780))
+
+
 ## HDMF 3.4.7 (October 26, 2022)
 
 ### Bug fixes
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -153,7 +153,7 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             doc=(
                 "If chunk_shape is not specified, it will be inferred as the smallest chunk "
                 "below the chunk_mb threshold.",
-                "Defaults to 1MB."
+                "Defaults to 1MB.",
             ),
             default=None,
         ),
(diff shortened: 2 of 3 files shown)


CodyCBakerPhD:
Need to...

- [x] Add a test for the error capture with example values of what it would take to get there
- [x] update changelog

bendichter:
> np.prod(some_iterable_of_ints) uses int32 in its product operation

Is that true even if the inputs are int64?

CodyCBakerPhD:
> Is that true even if the inputs are int64?

If the input `some_iterable_of_ints` is an actual numpy array set to dtype `np.int64` then `np.prod` will use the `dtype` of that array. These arguments are usually specified by users as `list`'s or `tuple`'s of native python `int`'s, which by default are 32-bit and only return a 64-bit (or higher I presume) after using the built-in multiplication operator.

An alternative approach here could just be a pre-step to upcast all relevant iterables of integers to numpy arrays of a much higher itemsize and all the `np.prod` calls would then use the `dtype` of that array.

oruebel:
> An alternative approach here could just be a pre-step to upcast all relevant iterables of integers to numpy arrays of a much higher itemsize and all the `np.prod` calls would then use the `dtype` of that array.

Unless this causes large cost for compute or memory (which I presume it does not), this seems like a good option as it addresses the issue early on, can help prevent future downstream problems, and ensures that those variables have a consistent definition as numpy arrays

CodyCBakerPhD:
@bendichter @oruebel What would you think if I added a lazily-imported `psutil` check that there's actually enough space in memory to load the estimated buffers as a general method for avoiding this kind of thing? Basically if someone has `psutil` installed a similar warning to what I originally posted will prevent the process from getting far enough to do any damage (note we always install and use `psutil` for similar things on NeuroConv)

oruebel:
> What would you think if I added a lazily-imported `psutil` check that there's actually enough space in memory to load the estimated buffers as a general method for avoiding this kind of thing

Warning early is a good idea. I would make it an optional, lazy import and optional dependency for `psutil`. For the warning we may want issue a warning for any settings that will results in requests larger than some large fraction (let's say 50%) of the memory to indicate that this may lead to slow-down.

CodyCBakerPhD:
Good for review now - `psutil` feature will be in subsequent PR

oruebel (approved):
Looks good to me

PR #781, opened 2022-11-02
Ensure slices of GenericDataChunkIterator have consistent data type

CodyCBakerPhD:
A follow-up to #780 lead to some issues in our dev branch testing suite on NeuroConv (https://github.com/catalystneuro/neuroconv/actions/runs/3379295582/jobs/5610660624); as explained in the `CHANGELOG.md`, the issue occurs because the `np.uint64` upcasting from #780 was indeed propagating down to the level of the `selection` ranges (tuple of slices of integers) of each `DataChunk` on the call to `next(some_iterator)`. However, not all ranges had consistent data types which lead to a strange interaction when you tried to do otherwise self-consistent operations with those slice values, such as `selection.stop - selection.start` returning as a `float` because the `stop` was a `uint64` and the `start` was a `int32`. While this isn't a really big problem since the actual value of the operation is unaffected, some downstream code relied on the output of such things having a consistent data type so I think we might as well ensure that as upstream as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,11 @@
 # HDMF Changelog
 
-## Upcoming
-
-### Bug fixes
-- Fixed an issue with the `data_utils.GenericDataChunkIterator` where if the underlying dataset was such that the `numpy.product` of the `maxshape` exceeded the range of the default `int32`, buffer overflow would occur and cause the true buffer shape to exceed available memory. This has been resolved by dropping all `numpy` operations (which forced casting within the passed data type) in favor of the unlimited precision of Python builtin integer types @codycbakerphd ([#780](https://github.com/hdmf-dev/hdmf/pull/780)) ([#781](https://github.com/hdmf-dev/hdmf/pull/781))
-
-
-## HDMF 3.4.7 (October 26, 2022)
+## HDMF 3.4.7 (November 9, 2022)
 
 ### Bug fixes
 - Fix an issue where not providing an optional argument to `__init__` of an auto-generated `MultiContainerInterface`
   class raised an error. @rly ([#779](https://github.com/hdmf-dev/hdmf/pull/779))
+- Fixed an issue with the `data_utils.GenericDataChunkIterator` where if the underlying dataset was such that the `numpy.product` of the `maxshape` exceeded the range of the default `int32`, buffer overflow would occur and cause the true buffer shape to exceed available memory. This has been resolved by dropping all `numpy` operations (which forced casting within the passed data type) in favor of the unlimited precision of Python builtin integer types @codycbakerphd ([#780](https://github.com/hdmf-dev/hdmf/pull/780)) ([#781](https://github.com/hdmf-dev/hdmf/pull/781))
 
 ## HDMF 3.4.6 (October 4, 2022)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,7 @@
 ## Upcoming
 
 ### Bug fixes
-- Fixed an issue with the `data_utils.GenericDataChunkIterator` where if the underlying dataset was such that the `numpy.product` of the `maxshape` exceeded the range of the default `int32`, buffer overflow would occur and cause the true buffer shape to exceed available memory. This has been resolved by upcasting all shape attributes and operations to use the `uint64` data type. @codycbakerphd ([#780](https://github.com/hdmf-dev/hdmf/pull/780))
-- Fixed issue following #780 where not all of the data types of the `selection` ranges for each DataChunk were consistent, leading to unexpected casting issues when attempting operations between endpoints (such as `selection.stop - selection.start` returning `float64` because one was a `uint64` and the other was a `int32`). @codycbakerphd ([#781](https://github.com/hdmf-dev/hdmf/pull/781))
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
OK well as much as I wanted to just have each selection from the iterator utilize the largest possible integer dtype, this apparently caused many more issues across different format-specific I/O packages for NeuroConv, so I guess we'll just enforce `int32` values here.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Bug fixes
 - Fixed an issue with the `data_utils.GenericDataChunkIterator` where if the underlying dataset was such that the `numpy.product` of the `maxshape` exceeded the range of the default `int32`, buffer overflow would occur and cause the true buffer shape to exceed available memory. This has been resolved by upcasting all shape attributes and operations to use the `uint64` data type. @codycbakerphd ([#780](https://github.com/hdmf-dev/hdmf/pull/780))
+- Fixed issue following #780 where not all of the data types of the `selection` ranges for each DataChunk were consistent, leading to unexpected casting issues when attempting operations between endpoints (such as `selection.stop - selection.start` returning `float64` because one was a `uint64` and the other was a `int32`). @codycbakerphd ([#781](https://github.com/hdmf-dev/hdmf/pull/781))
```
oruebel: Since #780 has been released yet and this PR just fixes issues with that PR, I think it is fine in the Changelog to just update and add a link to this PR with the previous entry in the changelog

CodyCBakerPhD: Simplified now

CodyCBakerPhD:
OK so after digging into this a bit deeper I've decided to just remove all the `numpy` array usages from the GenericDataChunkIterator calculations. They were only being used for conveniences like fancy indexing to make certain statements require less typing.

The main reasons for this are:

(i) while native Python `int` types are, as previously stated, by default `int32` they automatically upscale any time an operation is performed that would require it. What I did not realize is they actually upscale to an arbitrary resolution, even those beyond what `numpy` itself is capable of (which on the upper end would be `np.uint64`); the only caveat for product operations (as I showed previous) is to use the `math.prod` instead of `np.prod` (which casts back to fixed numpy dtypes) or iterative calls to the `*` operator

(ii) `numpy` operations appear to try to keep the outputs within the space of the dtype of the constituent objects (which was the source of all of this problem to begin with) and require the  user/dev to instead decide when to upcast things. This puts a lot of responsibility in our hands and also restricts us to certain value ranges (see i) that we may, at some point in the future, surpass

(iii) Most importantly, the values contained in the slices of our `DataChunk.selection`, the main output from the use of the iterator, are expected to be native Python ints for downstream use - it's technically a break in back-compatibility if any of them end up getting casted to `numpy` types, which is what triggered the failures detected by NeuroConv. While hard-casting the final values in the selection as `ints` before return (my original attempt at this PR) will 'kind of' work, the actual calculations prior to that are still restricted to the range provided by the `numpy` space, which is not future-proof

The tests here will require some fixing, I'll try to get to that later today, but the main code should be updated to this new standard.

CodyCBakerPhD:
I've also updated all the annotation return types (not actually used for anything, but readable by developers) to indicate this more rigorous enforcement.

oruebel:
> OK so after digging into this a bit deeper I've decided to just remove all the `numpy` array usages from the

Thanks for digging into this and the detailed reasoning. This makes sense to me.

CodyCBakerPhD:
OK, changes should be read to review now

src/hdmf/data_utils.py
```diff
@@ -127,6 +128,19 @@ def maxshape(self):
...
+def _temporary_prod(iterable):
+    """
+    We want to use math.prod but it is only available in Python >= 3.8 versions.
+
+    Once 3.7 support is dropped, this function can be removed
+    """
+    result = 1
+    for x in iterable:
+        result *= x
+    return result
```
oruebel: Should this function be be removed?

CodyCBakerPhD: Only once Python 3.7 is no longer supported by HDMF (still the lower bound of the CI right now)

src/hdmf/data_utils.py
```diff
@@ -127,6 +128,19 @@ def maxshape(self):
         raise NotImplementedError("maxshape not implemented for derived class")
 
 
+# TODO: remove this function
```
CodyCBakerPhD:
```suggestion
# TODO: remove this function once support for Python 3.7 is dropped
```

src/hdmf/data_utils.py
oruebel: Since this function is only use by ``GenericDataChunkIterator`` would it make sense to make this a private classmethod of ``GenericDataChunkIterator`` rather than exposing this on the module level?

End of life for Python 3.7 seems to be in ~7 months (27 Jun 2023). @CodyCBakerPhD can you create an issue ticket to make sure we have record/reminder that this function should be removed in the future.

oruebel: A one-line alternative for `` math.prod `` would be to use the following  ``functools.reduce(operator.mul, iterable, 1)``  (where ``iterable`` is your data, like in the current function)

oruebel: @rly what are your thoughts regarding the ``_temporary_prod``.

CodyCBakerPhD: I'll just switch to `functools.reduce(operator.mul, iterable, 1)` for now and leave notes and tickets that it can be simplified later on

CodyCBakerPhD: Reminder filed: https://github.com/hdmf-dev/hdmf/issues/785

oruebel: Perfect. Thanks!

oruebel (dismissed):

rly (approved):

rly:
Thanks for the update and PR!

PR #787, opened 2022-11-10
better shape error message

### LINKED ISSUE #1419: better shape check error in validator
see NeurodataWithoutBorders/helpdesk#16

bendichter:
## Motivation

fix https://github.com/NeurodataWithoutBorders/pynwb/issues/1419

## How to test the behavior?
```
```

## Checklist


rly:
I prefer "Any" instead of ":" to denote that any length is allowed.

rly:
I opted for `"*"` to represent a wildcard/any length to be consistent with what we are doing in LinkML and numpydantic. I also think `"*"` is pretty clear as a wildcard, whereas `":"` means select all elements

Finally getting this across the finish line. @stephprince Could you please review since I made a bunch of changes? @bendichter LMK if this is all right. Thanks for the PR!

rly:
PyNWB dev tests fail because it is searching for an exact match on the error message. I'll update PyNWB as well.

oruebel (dismissed):
Looks good to me

stephprince:
Looks good to me for addressing when any length is allowed for a dimension

However going back to the original issue, I find the message still confusing when multiple shapes are allowed. For example, in `TestDocValidator.test_multi_shape` the new warning is: `"incorrect shape for arg1: got (1, 3), and expected ((*,), (*, 2))"`. 

Do you think this would be potentially clarified by joining all expected shapes in the `allowable_shapes` list  with an "or"?

rly:
> Do you think this would be potentially clarified by joining all expected shapes in the `allowable_shapes` list with an "or"?

Great idea! It's so easy to be blind to what is not intuitive after seeing the same thing for years. I'll make the change.

rly:
@stephprince Updated!

rly (approved):

PR #799, opened 2022-12-05
Updated ExternalResources

mavaylon1:
## Motivation

The motivation is to bring pieces from other ER branches in development into one place ready for merge into dev to have an updated, clean starting point for the upcoming development of a separate External Resources File.

## Checklist


mavaylon1:
I need to change the csv to tsv, but so far what do you think? @rly 
I also did the fix for the workflows. Should that be in a separate branch?

Besides the tsv change it is good to go. I decided to keep the denormalized method here and will create a new branch and request for the normalized method.

rly:
@mavaylon1 Can you make a separate PR to fix the workflows? Thanks.

rly:
> @mavaylon1 Can you make a separate PR to fix the workflows? Thanks.

Actually never mind. I have a different fix so that we can use the latest version of `tox` instead of pinning to the older version. I'm testing it out here https://github.com/NeurodataWithoutBorders/pynwb/pull/1608 and will create an analogous PR in HDMF.

mavaylon1:
@rly I'll remove my workflow fix after your PR has been merged.

rly:
@mavaylon1 yes. I'll merge the other one first

src/hdmf/common/resources.py
```diff
@@ -472,7 +525,7 @@ def get_keys(self, **kwargs):
         Return a DataFrame with information about keys used to make references to external resources.
         The DataFrame will contain the following columns:
             - *key_name*:              the key that will be used for referencing an external resource
-            - *resources_idx*:         the index for the resourcetable
+            - *resources_idx*: cs        the index for the resourcetable
```
rly: Is the `cs` a typo?

mavaylon1: Weird. Yeah it is.

src/hdmf/common/resources.py
```diff
@@ -654,3 +707,136 @@ def export_to_sqlite(self, db_file):
             self.entities[:])
         connection.commit()
         connection.close()
+
+    @docval({'name': 'path', 'type': str, 'doc': 'path of the tsv file to write'}, allow_extra=False)
+    def to_tsv(self, **kwargs):
+        """
+        Write ExternalResources as a single, flat table to TSV
+        Internally, the function uses :py:meth:`pandas.DataFrame.to_tsv`. Pandas can
```
```suggestion
        Internally, the function uses :py:meth:`pandas.DataFrame.to_csv`. Pandas can
```

src/hdmf/common/resources.py
```diff
@@ -654,3 +708,139 @@ def export_to_sqlite(self, db_file):
...
+        df.to_csv(path, sep='\t')
+
+    @classmethod
+    @docval({'name': 'path', 'type': str, 'doc': 'path of the tsv file to read'},
+            allow_extra=False,
+            returns="ExternalResources loaded from TSV")
+    def from_tsv(cls, **kwargs):
+        """
+        Read ExternalResources from a flat tsv file
+        Fromatting of the TSV files are assumed to consistent with the format
```
```suggestion
        Fromatting of the TSV files are assumed to be consistent with the format
```

src/hdmf/common/resources.py
```diff
@@ -165,6 +166,59 @@ def __init__(self, **kwargs):
...
+        :raises AssertionError: Raised if any differences are found. The function collects
+            all differences into a single error so that the assertion will indicate
+            all found differences.
+        """
+        errors = []
+        try:
+            pd.testing.assert_frame_equal(left.keys.to_dataframe(),
+                                          right.keys.to_dataframe(),
+                                          check_dtype=check_dtype)
+        except AssertionError as e:
```
rly, line 190: It may seem silly, but can you test each of these error cases? In general, there may be an error in the exception handling (probably not here though). We should strive for 100% code coverage on new changes to HDMF, given that our goal is to get to 100% code coverage for the whole repo.

mavaylon1: Sure

src/hdmf/common/resources.py
```diff
@@ -654,3 +708,136 @@ def export_to_sqlite(self, db_file):
...
+        df = self.to_dataframe(use_categories=True)
+        df.to_csv(path, sep='\t')
+
+    @classmethod
+    @docval({'name': 'path', 'type': str, 'doc': 'path of the tsv file to read'},
+            returns="ExternalResources loaded from TSV", rtype="ExternalResources")
+    def from_tsv(cls, **kwargs):
+        """
+        Read ExternalResources from a flat tsv file
+        Fromatting of the TSV files are assumed to be consistent with the format
```
```suggestion
        Formatting of the TSV file is assumed to be consistent with the format
```

src/hdmf/common/resources.py
```diff
@@ -654,3 +708,136 @@ def export_to_sqlite(self, db_file):
...
+        generated by :py:meth:`~hdmf.common.resources.ExternalResources.to_tsv`.
+        The function attempts to validate that the data in the TSV is consistent.
```
```suggestion
        The function attempts to validate that the data in the TSV is consistent
        and parses the data from the denormalized table in the TSV to the
        normalized linked table structure used by ExternalResources. 
```

src/hdmf/common/resources.py
```diff
@@ -654,3 +708,136 @@ def export_to_sqlite(self, db_file):
...
+        Currently the checks focus on ensuring that row id links between tables are valid.
```
oruebel, line 750: Validating consistency of all values during read may be too expensive, but I think it would be useful to add ``validate_tsv`` function to allows us to validate tsv files. In this way, if a user does indeed edit an ExternalResources table in TSV they can validate it. Also this would make it easier to add validation of ExternalResources the validator. This does not need to be part of this PR, but I just wanted to mention that this would be useful to have. Maybe we can create a separate issue for this.

mavaylon1: Yes the validation is the next step after this is merged. This is mostly to get ER updated on dev.

src/hdmf/common/resources.py
```diff
@@ -165,6 +166,59 @@ def __init__(self, **kwargs):
...
+                                          right.entities.to_dataframe(),
+                                          check_dtype=check_dtype)
+        except AssertionError as e:
+            errors.append(e)
+        try:
+            pd.testing.assert_frame_equal(left.object_keys.to_dataframe(),
+                                          right.object_keys.to_dataframe(),
+                                          check_dtype=check_dtype)
+        except AssertionError as e:
+            errors.append(e)
```
rly, line 215: It looks like this assertion isn't being triggered. Could you check the test for this? @mavaylon1

mavaylon1: The tests pass for each case. I'm not sure what you mean.

rly: The tests are passing but that line is not being run by the tests. You can check the code coverage here: https://app.codecov.io/gh/hdmf-dev/hdmf/pull/799 or by running:
```
pytest --cov
coverage html && open ./tests/coverage/htmlcov/index.html
```

oruebel: the corresponding test for this should be ``test_invalid_object_keys_assert_external_resources_equal``. Probably a bug in how the test is setting up the data.

src/hdmf/common/resources.py
```diff
@@ -165,6 +166,57 @@ def __init__(self, **kwargs):
...
+        try:
+            pd.testing.assert_frame_equal(left.object_keys.to_dataframe(),
+                                          right.object_keys.to_dataframe(),
+                                          check_dtype=check_dtype)
+        finally:
+            if len(errors) > 0:
+                msg = ''.join(str(e)+"\n\n" for e in errors)
+                raise AssertionError(msg)
```
rly: Could you please explain this change? Why remove the `except` statement? Does this `finally` run only on the last `try` block?

rly: @mavaylon1 I'm still confused about this block of code and the change. Could you please explain it? It might be worth leaving a comment for future us there

mavaylon1: @rly The line removed was an except that was redundant. The prior except/try blocks accounted for all the tables. The finally just says "run this if statement always" and it will run after all the prior try blocks.

oruebel: The part that is confusing to me is that this try block does not have an except at all, i.e., the error that may be raised by this assert is not being recorded. Also, it is not clear to me why this code is part of the finally statement, rather than just code after the except. I.e., I would have expected something like:

```
try:
    pd.testing.assert_frame_equal(left.object_keys.to_dataframe(),
                                    right.object_keys.to_dataframe(),
                                    check_dtype=check_dtype)
except AssertionError as e:
    errors.append(e)
   
if len(errors) > 0:
    msg = ''.join(str(e)+"\n\n" for e in errors)
    raise AssertionError(msg)
```

Having the code in the finally statement is in principle fine, but I don't think it is necessary. However, without the except part, the error message from the last assert won't be recorded in the AssertionError that is being raised by the function

mavaylon1: Ahh I see what you're saying. I'll fix it now.

src/hdmf/common/resources.py
```diff
@@ -165,6 +166,59 @@ def __init__(self, **kwargs):
...
+                                          check_dtype=check_dtype)
+        except AssertionError as e:
+            errors.append(e)
+        try:
+            pd.testing.assert_frame_equal(left.object_keys.to_dataframe(),
+                                          right.object_keys.to_dataframe(),
+                                          check_dtype=check_dtype)
+        except AssertionError as e:
+            errors.append(e)
+        finally:
```
rly: Can you remove the `finally` block? I do not think it does anything anymore and it raises the question of what it does. 

(technically it will get called if an error that is not an AssertionError is raised within the L210-L215 try/except block, which is OK but unlikely. but if we want that behavior, then there should be a `finally` block after every try/except block. i think this is unnecessary)

mavaylon1: @rly sure.

rly: Thanks!

rly:
Looks good to me! Thanks for your work on this.

rly (approved):

Issue #785, opened 2022-11-09
[Feature]: When Python 3.7 support is dropped, use math.prod instead of functools


CodyCBakerPhD:
### What would you like to see added to HDMF?

Leaving this as a reminder...

In the `GenericDataChunkIterator` from PR #781, we used a `functools` and `operator` workaround to `math.prod` for Python 3.7 support.

When Python 3.7 support is dropped, we can swap to fully using `math.prod`.

### What solution would you like?

Replace all the lines commented with `# TODO: replace when Python 3.7 support is dropped` in the `GenericDataChunkIterator`

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/hdmf-dev/hdmf/issues)?

PR #803, opened 2022-12-17
Add Python 3.11 support

rly:
## Motivation

- Add Python 3.11 support
~- Drop Python 3.7 support. Major scientific Python packages, such as numpy, pandas, and scipy, have already dropped Python 3.7 support in their latest releases~
- Do not set upper bounds on package installation dependencies. Having no upper bounds is generally recommended for libraries. See https://iscinumpy.dev/post/bound-version-constraints/
- Also test Pandas 2.0.0

~NOTE: h5py has not yet released wheels for Python 3.11, so all tests will currently fail.~
## Checklist


rly:
Now that h5py has released wheels for Python 3.11, I would like to make a minor release of hdmf in the next week or two to add support for Python 3.11.

TODO:
- [ ] Update pynwb to use python 3.11 in a branch and make sure tests pass
- [ ] Update branch protection

yarikoptic:
> * Drop Python 3.7 support. Major scientific Python packages, such as numpy, pandas, and scipy, have already dropped Python 3.7 support in their latest releases

yikes!  numpy even dropped 3.8 already in its ~~`master`~~ `main` as of a month back: https://github.com/numpy/numpy/pull/22902 to be released in 1.25.0 in July 2023  but 3.8 is EOLing only in over 1.5 years (14 Oct 2024) !!  I wonder why they are rushing ?

yarikoptic:
I guess might stem from https://numpy.org/neps/nep-0029-deprecation_policy.html where 3.9 becomes minimal supported some time this Apr 2023. yet to comprehend it fully but at least it answers the question "why".  It also doesn't imply that downstream projects (such as hdmf etc) *have* to drop versions no longer supported by numpy etc, since they could still be used with earlier versions of python and numpy which might still be supported or simply be supported/present in some given distribution.

rly:
Yes, we don't have to drop Python 3.7 support for newer versions of HDMF (EOL is 27 Jun 2023). The main benefits are that we can use Python 3.8+ only features (not critical but nice to have) and we need not test on Python 3.7 in CI. Dropping Python 3.6 was more useful since Python 3.7 introduced some nice-to-have features. Python 3.8 is less important. I'll update the PR, and we can drop Python 3.7 if we come across an incompatibility.

rly:
Note that https://github.com/hdmf-dev/hdmf/issues/785 highlights an inconvenience with Python 3.7. We have worked around it though.

rly:
PyNWB tests pass locally on my Mac M1 with Python 3.11. We can check for issues with other OSs  in the PyNWB repo.

I also made a change to the `DynamicTable.to_dataframe()` to ensure that the index column of the dataframe has type int64, for consistency across OS (windows uses int32 by default, but mac and linux use int64).

oruebel (approved):

PR #812, opened 2023-01-07
Allow custom message as part of TestCase.assert... functions

oruebel:
## Motivation

Similar to standard assert functions of ``unittest.TestCase`` it will be useful to allow specification of a custom message to be added for asserts defined by HDMFs``hdmf.testing.testcase.TestCase`` class. This is useful, e.g., when tests are parametrized and we want to include test parameters as part of the error message from asserts when a test is failing. 

I came across this issue in hdmf_zarr where I need to run the same test with different backend parameters and it would be useful to include additional information in error messages when tests are failing.

## Checklist

### DIFF
--- a/src/hdmf/testing/testcase.py
+++ b/src/hdmf/testing/testcase.py
@@ -34,29 +34,39 @@ class TestCase(unittest.TestCase):
 
         return self.assertWarnsRegex(warn_type, '^%s$' % re.escape(exc_msg), *args, **kwargs)
 
-    def assertContainerEqual(self, container1, container2,
-                             ignore_name=False, ignore_hdmf_attrs=False, ignore_string_to_byte=False):
+    def assertContainerEqual(self,
+                             container1,
+                             container2,
+                             ignore_name=False,
+                             ignore_hdmf_attrs=False,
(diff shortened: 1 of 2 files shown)


src/hdmf/testing/testcase.py
```diff
@@ -66,47 +76,103 @@ def assertContainerEqual(self, container1, container2,
...
+                                      ignore_string_to_byte=ignore_string_to_byte,
+                                      message=message)
         elif isinstance(f1, Data):
             self._assert_data_equal(f1, f2,
                                     ignore_hdmf_attrs=ignore_hdmf_attrs,
-                                    ignore_string_to_byte=ignore_string_to_byte)
+                                    ignore_string_to_byte=ignore_string_to_byte,
+                                    message=message)
         elif isinstance(f1, (float, np.floating)):
             np.testing.assert_allclose(f1, f2)
```
rly: Can you add `message` to the `np.testing` calls too? There are a few/

oruebel: Good catch. Done.

mavaylon1 (approved):

rly (approved):

Issue #181, opened 2019-10-18
Document AbstractContainer.__fields__ in HDMF and PyNWB


oruebel:
The ``Container`` class supports the specification of ``__fields__`` in PyNWB these are described as a list of strings https://pynwb.readthedocs.io/en/stable/building_api.html#nwbfields and they seem to be omitted in the HDMF docs https://hdmf-dev.github.io/hdmfdocs It will be useful to update/add documentation for __fields__. In particular __fields__ allows for lists for dicts with the keys ``{'name', 'child', 'required_name', 'doc', 'settable'}`` and it would be useful to document the specific function of those keys both in HDMF and PyNWB

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you included a minimal code snippet that reproduces the issue you are encountering?
- [X] Have you checked our [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?

mavaylon1:
#827

PR #827, opened 2023-03-02
Update documentation for fields

mavaylon1:
## Motivation

Add further documentation for __fields__ in AbstractContainer
## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -233,7 +233,7 @@ class AbstractContainer(metaclass=ExtenderMeta):
         `fields` allows for lists and for dicts with the keys {'name', 'child', 'required_name', 'doc', 'settable'}.
         1. name: The name of the field property
         2. child: A boolean value to set the parent/child relationship between the field property and the container.
-        3. required_name: TBD
+        3. required_name: The name the field property must have such that `name` matches `required_name`.
         4. doc: Documentation of the field property
         5. settable: TBD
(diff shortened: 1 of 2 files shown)


mavaylon1:
Fix #181

mavaylon1:
@rly Could you explain to me the difference between name and required_name? From my understanding (from the setter), when required_name is set, it just means the name must match the required_name. 
I couldn't get much info from the code about settable so some insight would be great about that key.

src/hdmf/container.py
```diff
@@ -228,6 +228,15 @@ def get_ancestor(self, **kwargs):
...
     def fields(self):
+        '''
+        Subclasses use this class attribute to add properties to autogenerate.
+        `fields` allows for lists and for dicts with the keys {'name', 'child', 'required_name', 'doc', 'settable'}.
+        1. name: The name of the field property
+        2. child: A boolean value to set the parent/child relationship between the field property and the container.
+        3. required_name: The name the field property must have such that `name` matches `required_name`.
+        4. doc: Documentation of the field property
+        5. settable: TBD
+        '''
```
oruebel: ``settable`` refers to whether a setter function will be added for the field. If ``true`` a setter function is created so that the field can be changed after creation. If false, then the field is private and cannot be set after init.

oruebel (approved):

Issue #835, opened 2023-03-13
[Bug]: Deploy Release Authentication


mavaylon1:
### What happened?

Upon release 3.5.2, the Deploy Release Action returned
```
25hWARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/          
         Invalid or non-existent authentication information. See                
         https://pypi.org/help/#invalid-auth for more information.              
Error: Process completed with exit code 1.
```

### Steps to Reproduce

```shell
Action for release
```

### Traceback

```shell
25hWARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/          
         Invalid or non-existent authentication information. See                
         https://pypi.org/help/#invalid-auth for more information.              
Error: Process completed with exit code 1.
```

### Operating System

Linux

### Python Executable

Conda

### Python Version

3.9

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this bug was not already [reported](https://github.com/hdmf-dev/hdmf/issues)?

mavaylon1:
@rly @oruebel Some kind of authentication issue for a new release

mavaylon1:
@rly I saw that you changed/updated the password on the doc. I updated the password here and now get a new error before the publish phase.
In the upload phase
```
twine upload: error: the following arguments are required: dist

```

PR #837, opened 2023-03-14
Update deploy_release.yml

mavaylon1:
## Motivation

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/.github/workflows/deploy_release.yml
+++ b/.github/workflows/deploy_release.yml
@@ -43,8 +43,8 @@ jobs:
         run: |
           python -m pip install twine
           ls -1 dist
-          twine upload --repository-url https://test.pypi.org/legacy/ -u ${{ secrets.BOT_PYPI_USER }} -p ${{ secrets.BOT_PYPI_PASSWORD }} --skip-existing dist/*
-          # twine upload -u ${{ secrets.BOT_PYPI_USER }} -p ${{ secrets.BOT_PYPI_PASSWORD }} --skip-existing dist/*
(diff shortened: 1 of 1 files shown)


mavaylon1:
@rly I manually did a release by changing the code to what you see on the file. Though I am not sure if this was a one time fix or an actual fix. I know that "--skip-existing dist/* " is used to upload new versions to pypi so I am not sure if I should add it back in. The error (as seen in #835)
```
twine upload: error: the following arguments are required: dist
```
made me just think to add dist to the upload and remove the skip. The release is out and so this is now a future question vs a question for this release.

rly:
`twine upload [arguments] dist/*` is used to upload new versions. Usually the arguments like `-u username`, `-p password`, and `--skip-existing` go first. Did you get the `twine upload: error: the following arguments are required: dist` error locally? If so, did you have a distribution in the `dist` folder? The complaint might actually have been that there was nothing to upload. bash will resolve `dist/*` to be a list of all the items in the `dist/` directory but if there is nothing there (you need to generate wheels in the previous CI steps), then it would complain that the expected value for `dist` is missing.

Now that the username and password are fixed (sorry, my bad), I think the original command should work. Let's reset it (i.e., change it back to the way it was before the debug change) and let's try this again in the next release.

mavaylon1:
I'll change it back, but I got the dist issue on github and locally. I agree that we will see for sure on the next release. I'll change it back.

rly (approved):

PR #841, opened 2023-03-27
ER updates and Term set integration

mavaylon1:
## Motivation

This will host the next stage of tools for ExternalResources, including the following:

- [x] Add ``object_type`` column to the ``ObjectTable`` to save the ``data_type`` string of the parent object {5/17}
- [x] Add method to query based on ``object_type`` (including ``attribute`` and ``field``) {5/17}
- [ ] Make the behavior whether to automatically add to the ExternalResources configurable {TBD}
- [x] Change ``NWBFile.external_resources`` to a method ``NWBFile.link_resources``, rename the attribute to ``NWBFile.__external_resources`` to make it private, and add a method ``NWBFile.get_linked_resources``  {5/17}
- [ ] We still need to decide on the which TSV format to use. Options are 1) flat TSV, 2) multiple TSV in a folder, 3) multiple TSV in a single zip file. {TBD on group feedback}
- [x] Update NWBHDF5IO to have ``external_resource`` as parameter that is a path to read/write external resources to/from. If we decide on a default path to store external resources at, then we could also allow a boolean``True/False`` option to use the default behavior. {5/18} (This is now supported for FLAT only)
- [ ] _After adding a Table to the NWBFile, make sure any additional data (either row or column) can auto-populate the linked ER._ {No longer planned due to open discussion on the issue}
- [ ] _Auto-add to ER for objects already existing in NWBFile._ {No longer planned due to open discussion on the issue}
- [ ] Update all tests for new ObjectTable column and File changing to Container only (no longer string support)
- [ ] Add tests for get_object_type
- [ ] import/export for ER to combine flat/norm {5/19}
- [ ] Move TermSet to HDMF as a branch {5/19}
- [ ] Fix to_dataframe for files_idx
## How to test the behavior?
```
```

## Checklist


src/hdmf/common/resources.py
```diff
@@ -13,7 +13,24 @@
...
+             'doc': 'The external resources to be used for the container.',
+             'default': None},)
+    def set_er(self, **kwargs):
+        """
+        Method to attach an instance of ExternalResources in order to auto-add terms/references to data.
+        """
+        external_resources = kwargs['external_resources']
+        self._external_resources = external_resources
+        if self._external_resources is not None:
+            self.add_child(self._external_resources)
```
oruebel: @mavaylon1  since you are using ``add_child`` here, does the class assume that the whichever class inherits from ExternalResourcesManager also inherits from AbstractContainer? If so, then we should mention this in the docstring and maybe call it ExternalResourcesManagerMIxin to clarify that it is supposed to be used in combination with other classes.

oruebel:
From the meeting with CatalystNeuro team:

- [ ] Add ``object_type`` column to the ``ObjectTable`` to save the ``data_type`` string of the parent object
- [ ] Add method to query based on ``object_type`` (including ``attribute`` and ``field``)
- [ ] Make the behavior whether to automatically add to the ExternalResources configurable 
- [ ] Change ``NWBFile.external_resources`` to a method ``NWBFile.link_resources``, rename the attribute to ``NWBFile.__external_resources`` to make it private, and add a method ``NWBFile.get_linked_resources`` 
- [ ] We still need to decide on the which TSV format to use. Options are 1) flat TSV, 2) multiple TSV in a folder, 3) multiple TSV in a single zip file.
- [ ] Update NWBHDF5IO to have ``external_resource`` as parameter that is a path to read/write external resources to/from. If we decide on a default path to store external resources at, then we could also allow a boolean``True/False`` option to use the default behavior. 

*These have been added to the TODO under the PR Motivation* ~ Matthew was here

mavaylon1:
@oruebel remind what " Make the behavior whether to automatically add to the ExternalResources configurable" means. Right now you can auto-add based on if the Object has been added to a NWBFile. I've tested this with DynamicTable and need to work out some kinks, but let me know if this is a different idea.

oruebel:
> what " Make the behavior whether to automatically add to the ExternalResources configurable" means

The scenario here was (if I remember correctly), a user links an ExternalResources object to an NWBFile and you have a TermSet associated with an object, e.g., a VectorData. In this case, should we automatically add to the ExternalResources as users add data or should the TermSet only be used for validation. Ultimately, this behavior should be configurable. 

In HDMF I guess this could be done either on the ExternalResourcesManager() level or in the mechanism that is being used to add a TermSet to an object.

mavaylon1:
@oruebel 
Case 1: A user has a NWBFile and adds an object with a Termset. The user then links an ExternalResources. On that link do we then retroactively search through the file to populate ExternalResources?

Case 2: Your case is user creates a NWBFile and links ER. User then adds a table with a VectorData that has a termset and so we want that auto-add to ER to not be automatic but be configurable. Would that be then a boolean parameter on Data?

oruebel:
> Case 1: A user has a NWBFile and adds an object with a Termset. The user then links an ExternalResources. On that link do we then retroactively search through the file to populate ExternalResources?

This case I think would be a parameter on ``ExternalResourceManager.link_resources``

> Case 2: Your case is user creates a NWBFile and links ER. User then adds a table with a VectorData that has a termset and so we want that auto-add to ER to not be automatic but be configurable. Would that be then a boolean parameter on Data?

I'm wondering whether that may be something for TermSet?

However, as we are thinking through these cases, it feels like this may create a lot of corner cases and "magic" that we need to do under the hood. Maybe the better approach could be to have a function ``ExternalResourcesManager.populate_linked_resources_from_term_sets``  that the user would call explicitly after they are done populating their data so that the user would decide when to iterate through all the objects to find objects with TermSets, and then add all the links. What do you think @mavaylon1 @rly ?

mavaylon1:
I think trying to navigate various parent/child hierarchies would be a lot to maintain on not only what is currently supported but any new features going forward. I agree that both cases have a lot of "magic" and I think with number of corner cases, only supporting populating ER "on add to the file" and if they don't have a linked ER first (which should be clearly stated as a best practice) then it is on them to add references. @oruebel

oruebel:
> only supporting populating ER "on add to the file"

Could you clarify this. Do you mean with "add to file" as adding a Container to a file? If ER is being populated when adding a Container then this would work if the Container is already populated with data before it is being added to the File but may not work if the user adds data to the Container after adding it to the file. 

> have a function `ExternalResourcesManager.populate_linked_resources_from_term_sets`

What I was referring to with with this is essentially a shorthand for calling ``add_ref_term_set`` for all objects of a file that have a TermSet associated with them, so that a user can populate ER with a single call. However, this would make it explicit when ER is being populate since the user would call the function explicitly. That is, in this case there would be no need to automatically add to ER at any time, but this just provides a convenient way to populate ER in one single call from a collection of objects. I.e,. the workflow in this case would be:

1. Create a NWBFile
2. Call ``ExternalResourceManager.link_resources`` for the NWBFile (this could happen here or really at any time after creating the file, independent of when Containers with data are being added to the file)
3. Create data Containers and set TermSets
4. Add the Containers to the NWBFile (a user may add data to the Containers before and/or after adding them to the file)
5. Write the NWBFile
6. Call ``ExternalResourcesManager.populate_linked_resources_from_term_sets`` on the NWBFile Doing this after writing the NWBFile is only necessary if a user uses DataChunkIterators for write where the data is not know before write. For all other cases where a users just adds data directly, this could also happen before writing the NWBFIle.
7. Write the ExternalResources

mavaylon1:
@oruebel That sounds ideal. What I am worried about is how easy/hard is it to search through the NWBFile for an object with a TermSet. Any where from a generic DynamicTable to something that has it's own section in the file, such as, electrode groups. The file is like a tree in a way where you have to go down different "branches" to find the "leaves" that has, in this case, a TermSet. Right now it's only Data and its subclasses, but we talked about TermSet being used elsewhere. 

I suppose its just starting at the file and iteratively checking the children for a TermSet, and then the child of the child and so on. But then what attributes that is not a child, such as experimenter, where it is not clear where the termset would exist. The NWBFile in this example? There's a lot of "what ifs" still in my view. 

As for going forward, I won't pursue the "enhancing" of auto-add to ER w/ TermSets until we decide.

oruebel:
> iteratively checking the children

There should already be functions on Container to get and/or iterate over all children in the full subtree

mavaylon1:
> > iteratively checking the children
> 
> There should already be functions on Container to get and/or iterate over all children in the full subtree

I'll play around with this and give an update Wednesday.

oruebel:
> I'll play around with this and give an update Wednesday.

In NWBFile there is the ``all_children`` method https://pynwb.readthedocs.io/en/stable/_modules/pynwb/file.html#NWBFile.all_children   It may make sense to just move that one to HDMF (maybe``AbstractContainer``) instead if we need it here.

PR #844, opened 2023-04-06
Use ruff, pyproject.toml, pre-commit; rm versioneer

rly:
## Motivation

This PR involves ~~a LOT of stylistic and~~ infrastructure changes that do not affect functionality but affect the developer experience.

The Python Packaging Authority is gradually phasing out `setup.py` in favor of `pyproject.toml`. We make the change here. 

This involves removing `versioneer` as a dependency due to challenges to get it to play nicely with `pyproject.toml`. Using `setuptools_scm` for setting the package version appears to be an adequate replacement. 

~~We will also now be using the popular `black` and `ruff` tools to impose a strict, mostly uncompromising style on the code base. `ruff` replaces `flake8` and `isort` and performs additional checks. Running these tools on the code base involves changing basically every python file...~~

We will now use `ruff` to replace `flake8`. It is significantly faster, sorts imports, and performs additional checks.

Finally, to help automate the usage of ~~`black`~~, `ruff`, `codespell`, I recommend that developers install and use `pre-commit`, which runs these tools as well as several other helpful utility functions, to clean up the code and identify issues prior to every commit.

I updated the developer installation instructions and the contributing guide accordingly. 

Minor: 
- `pytest` will now run `coverage` by default and create an html output in `./htmlcov` instead of `./tests/htmlcov`
- `MANIFEST.in` not excludes `__pycache__` and adds some missing files

~~The vast majority of the changes in this PR are the result of running `black` and `isort` on the codebase.~~

## Checklist


bendichter:
Could we also run precommit on the GitHub side? That way the user does not need to worry about precommit on their side. We have been using it at CN and it has been working great. @CodyCBakerPhD has experience setting this up.

CodyCBakerPhD:
Not that any experience is necessary: it's dirt simple to turn on at https://results.pre-commit.ci/, just needs to be done by someone with admin privelages

rly:
Thanks! I was going to ask yall how you did that on the GitHub side. Doesn't hurt to have both the GitHub side and the optional local installation.

rly:
@oruebel @mavaylon1 @bendichter @ajtritt 
This PR is now good to go. The code changes consist of stylistic changes from applying `black`, removing `versioneer`, updating `_due.py`, removing `test.py`, and moving `tests/unit/utils.py` to `tests/unit/helpers/utils.py`.

I have set up a pre-commit service to run `pre-commit` on the GitHub side on every commit. This might result in a new commit from the pre-commit service that refactors the code with `ruff` and `black`. If you are not running `pre-commit` locally, it is likely that the pre-commit service will make a commit in the branch after you push. Therefore, you may want to get in the habit of pulling the latest version of your branch, which would contain the changes from the pre-commit service, before you push new changes.

To avoid this inconvenience and also take full advantage of `pre-commit`, I recommend installing `pre-commit` locally.

.github/pull_request_template.md
mavaylon1, line 1: Why the switch?

docs/CONTRIBUTING.rst
```diff
@@ -79,27 +79,46 @@ From your local copy directory, use the following commands.
...
+Before you create a Pull Request, make sure you are following the Black_ preview style guide, which follows PEP8,
+with code E203 ignored. We also break from the Black format by configuring it with a max line length of 120.
+To check whether your code conforms to the Black_ preview style guide, simply run the ``black`` tool in the project's root
+directory with the ``--check`` argument. You can also run the ``black`` tool without the ``--check`` argument to have
+black automatically format the codebase to comply with the black preview style guide.
 
-Python Code Styleguide
-^^^^^^^^^^^^^^^^^^^^^^
+We also use the ruff_ tool to sort python imports automatically and double-check that the codebase
+conforms to PEP8 standards, and the codespell_ tool to check spelling.
```
mavaylon1: We also use the ruff_ tool to sort python imports automatically and double-check that the codebase
conforms to PEP8 standards, while using the codespell_ tool to check spelling.

src/hdmf/common/resources.py
```diff
@@ -82,21 +89,35 @@ class ObjectTable(Table):
...
-                 'an external resource reference key. Use an empty string if not applicable.')},
-        {'name': 'field', 'type': str,
-         'doc': ('The field of the compound data type using an external resource. '
-                 'Use an empty string if not applicable.')}
+        {
+            "name": "files_idx",
+            "type": int,
+            "doc": "The row idx for the file_object_id in FileTable containing the object.",
+        },
+        {"name": "object_id", "type": str, "doc": "The object ID for the Container/Data."},
```
mavaylon1: I think this should also follow the style changes of the others.

src/hdmf/common/resources.py
```diff
@@ -131,40 +158,56 @@ class ObjectKey(Row):
...
+            "default": None,
+            "doc": "The table storing user keys for referencing resources.",
+        },
+        {
+            "name": "files",
+            "type": FileTable,
+            "default": None,
+            "doc": "The table for storing file ids used in external resources.",
+        },
+        {"name": "entities", "type": EntityTable, "default": None, "doc": "The table storing entity information."},
```
mavaylon1: These should be changed to match the same style as the others

src/hdmf/common/resources.py
```diff
@@ -184,41 +227,53 @@ def assert_external_resources_equal(left, right, check_dtype=True):
         """
         errors = []
         try:
-            pd.testing.assert_frame_equal(left.keys.to_dataframe(),
-                                          right.keys.to_dataframe(),
-                                          check_dtype=check_dtype)
+            pd.testing.assert_frame_equal(
```
mavaylon1: I feel like the old way is easier to read.

mavaylon1: This comment applies through the file for the spacing changes.

rly:
`ruff` is 35x faster than `flake8` and more versatile. See https://github.com/charliermarsh/ruff

src/hdmf/common/resources.py
mavaylon1: This comment applies to the rest of the file to avoid repeating the same comment

src/hdmf/common/table.py
```diff
@@ -34,20 +34,30 @@ class VectorData(Data):
 
     __fields__ = ("description",)
 
-    @docval({'name': 'name', 'type': str, 'doc': 'the name of this VectorData'},
-            {'name': 'description', 'type': str, 'doc': 'a description for this column'},
-            {'name': 'data', 'type': ('array_data', 'data'),
-             'doc': 'a dataset where the first dimension is a concatenation of multiple vectors', 'default': list()},
-            allow_positional=AllowPositional.WARNING)
+    @docval(
+        {"name": "name", "type": str, "doc": "the name of this VectorData"},
```
mavaylon1: Same here with the style of the spacing should match the others

mavaylon1 (changes requested):
Made a few comments for style consistency and also a concern on whether the new style is better.

rly:
`black` is a strict code formatter. According to the `black` philosophy, by ceding control over the code style to the black style, we spend less mental effort worrying about whether our code conforms to the style of a given codebase and making minute adjustments to the code to comply to that style, e.g. where to put line breaks, whether to put a comma after the last list element, whether to use single or double quotes, etc. In some cases, this will make the code "uglier" or harder to read. We would have to adjust those cases manually.

We can debate whether these style changes are worth it. I think they are worth it because inconsistent styling across the codebase and PRs bothers me and I go and fix it or call it out in a PR, which feels like a waste of effort. I think that that is worth the cost of adjusting the hard to read code once. I get that that reasoning might just be a me problem though. Happy to discuss it!

docs/CONTRIBUTING.rst
rly: FYI you can use the "suggestion" option in a review comment to make suggestions like this. That is more readable, and I would just need to press accept to commit the change.

.github/pull_request_template.md
rly: See comment in the PR

docs/CONTRIBUTING.rst
rly:
```suggestion
We also use the ruff_ tool to sort python imports automatically and double-check that the codebase
conforms to PEP8 standards, while using the codespell_ tool to check spelling.
```

src/hdmf/common/resources.py
rly: I agree that this is ugly. I'll look into what I can change to make this consistent.

rly:
It looks like the rule for `black` is that this code:
```python
@docval({"name": "key_name", "type": str, "doc": "The name of the key to be added."})
```
is kept as is by `black`. If we want to change that to:
```python
@docval(
    {
        "name": "key_name",
        "type": str, 
        "doc": "The name of the key to be added.",
    }
)
```
Then we need to add a comma at the end of the last element in the dictionary. In general, I would say that the dicts within a single docval call should be consistent with each other. If all the dicts can fit on one line, then they should stay as one line. Otherwise, expand all the dicts to the second form above, which makes the multi-line dicts more readable, in my opinion.

rly:
Please hold off on reviews until I finish adjusting some code style.

rly:
I am breaking up this PR into multiple -- this one and one where we apply `black` -- in order to ease the review process and make the changes more visible. This is now ready for review again.

mavaylon1 (approved):

oruebel (approved):

PR #860, opened 2023-05-11
Use and apply black formatting

rly:
## Motivation

I added `black` to pre-commit so it will automatically modify the code to comply with black's uncompromising style guide. 

In this PR, I made some manual adjustments, particularly to docval statements so that the arguments they are consistent. Note that this code:
```python
@docval({"name": "key_name", "type": str, "doc": "The name of the key to be added."})
```
is kept as is by `black`. If we want to change that to:
```python
@docval(
    {
        "name": "key_name",
        "type": str, 
        "doc": "The name of the key to be added.",
    }
)
```
then we need to add a comma at the end of the last element in the dictionary. 

For us, I think we should keep the dict arguments within a single docval call consistent with each other. If all the dicts can fit on one line, then they should all be on one line. Otherwise, expand all the dicts to the second form above, which makes the arguments more readable in my opinon, because it is easier to see the parallelism of the arguments.

See also #844.

## How to test the behavior?
```
```

## Checklist


bendichter:
@rly I'm OK with either: basing the format of the docval calls on line length alone or forcing regularity. I think the way you have chosen is perfectly fine. I was not aware of the trailing comma feature of Black, but sure enough it's a feature of the package. [source](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma)

oruebel:
> or us, I think we should keep the dict arguments within a single docval call consistent with each other. If all the dicts can fit on one line, then they should all be on one line. Otherwise, expand all the dicts to the second form above

That sounds good to me too

bendichter (dismissed):

bendichter (approved):

bendichter:
@rly are we good to merge this?

PR #866, opened 2023-05-17
Convert object arrays to string dtype by default

oruebel:
## Motivation

This is related to https://github.com/hdmf-dev/hdmf-zarr/issues/79 when we want to convert a file from Zarr to HDF5 using the export function. This leads to errors for datasets of variable-length strings.

## Problem 

As far as I can tell the progression of the error is as follows:

* https://github.com/hdmf-dev/hdmf/pull/848 added a call to `clear_cache` before export. This means that existing builders are discarded and rebuild from containers
* Without `clear_cache` the dtype is determined from the Builder `builder.dtype` which is set correctly when reading from Zarr. However, with `clear_cache` the builders are being recreated, so that the `dtype` needs to be determined from the array directly
* For variable-length strings, hdmf_zarr uses the the convenience method to set  `dtype=str`, which is a short-hand for dtype=object, object_codec=numcodecs.VLenUTF8().  This means, variable length strings are stored as object and are being resolved via the object_coded
* Currently, the `ObjectMapper` does not consider this case, so that `numpy._object` is being returned as the dtype in this case
* This in turn leads to an error in `HDF5IO` because `h5py` does not know how to store generic Python objects. 

## Approach

This PR addresses this problem by setting the `dtype` for all arrays that have an object dtype to be a unicode string. In principle this should be safe (at least right now), because we should never have generic objects as data. However, it seems a more elegant solution would be if we could use the dtype from the original builder. 

@rly what do you think: 1) is the approach in this PR appropriate, 2) is there a better way to determine whether an object array stores strings, instead of just assuming objects are unicode, 3) is the `clear_cache` really necessary or is there an alternate approach.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## HMDF 3.6.1 (upcoming)
 
 ### Bug fixes
-- Fix compatability with hdmf_zarr for converting string arrays from Zarr to HDF5 by adding logic to determine the dtype for object arrays. @oruebel [#866](https://github.com/hdmf-dev/hdmf/pull/866)
+- Fix compatibility with hdmf_zarr for converting string arrays from Zarr to HDF5 by adding logic to determine the dtype for object arrays. @oruebel [#866](https://github.com/hdmf-dev/hdmf/pull/866)
 
 ## HDMF 3.6.0 (May 12, 2023)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HMDF 3.6.1 (upcoming)
+
+### Bug fixes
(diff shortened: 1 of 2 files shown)


rly:
I think this is fine to change. I cannot really think of a better way. We _could_ test whether an object is a zarr array and check the filters, but it is a bit messy...
```python
za = ...  # a zarr array
if hasattr(za, "filters"):
    # check whether one of the zarr array filters is of type <class 'numcodecs.vlen.VLenUTF8'> without
    # importing numcodecs...
    filter_types = [type(f).__module__ + "." + type(f).__qualname__ for f in za.filters]
    if "numcodecs.vlen.VLenUTF8" in filter_types:
        ret_dtype = 'utf8'
    else:
        raise ValueError("Unsupported Zarr dtype")
```

rly:
> 3) is the clear_cache really necessary or is there an alternate approach.
I will look into this

rly (approved):

PR #868, opened 2023-05-18
Allow reassigning and resetting AbstractContainer attributes/fields

### LINKED ISSUE #411: Allow resetting of unwritten attributes?
After working side by side with a few users converting their data to NWB using PyNWB, especially in Jupyter notebooks, a common frustration is that they cannot reset an attribute that already exists. They get the error: `"AttributeError: can't set attribute '...' -- already set"`. Why are users blocked from resetting attributes of unwritten containers? 

I can see that this would be problematic when resetting attributes on written containers, but what about unwritten containers (i.e., ones where container_source is None)?

I can also see that docval on `__init__` handles a lot of type checking and shape checking logic, but we can add such checks to every setter.

Example user workflow when writing a conversion script:
```python
nwbfile = NWBFile(...)  # create nwb file
nwbfile.add_acquisition(...)  # add a bunch of things to the NWB file
nwbfile.subject = Subject(...)
# user realizes they missed something in nwbfile.subject
# user re-runs the cell that sets nwbfile.subject and get the error "AttributeError: can't set attribute 'subject' -- already set"
# user doesn't really understand the problem or how to fix it...
# the only solution is to restart the notebook from the beginning when they created the nwbfile object
```

With the resetting of Subject, we would have to make sure that NWBFile updates its children, but that can be done.

Another example:
```
nwb.acquisition["EEG Data"].description = "new description"
```

It feels like an unnecessary hurdle that users (especially novice programmers) have to work around. 

Certain fields should not be settable though, e.g., `nwbfile.acquisition`

See also https://github.com/NeurodataWithoutBorders/pynwb/issues/148

## Checklist

### LINKED ISSUE #865: [Feature]: allow modification of fields of unbound containers

Add field to containers indicating if they are _bound_ to a file or _unbound_ (in memory only)

The  motivation is we would like to set chunking and compression after specifying the data

For unbound containers, you should be able to change existing fields

No.

### Code of Conduct

rly:
## Motivation

Fix #411, fix #865

Major change to allow setting attributes on Container objects to new values, including None. If the attributes were read from a file (i.e. `container_source is not None`), then a warning is raised.

Note that we currently do a lot of validation at the constructor level (type validation, shape validation) in docval. We do not do that for each setter. This can result in errors on write if the user is not careful. We could instead move validation of attributes to outside of docval and into the setter. That is significantly more development, but I think worthwhile in the long run. I will open a separate issue for this.

## Checklist


rly:
PyNWB tests need to be updated because container attributes that were previously set to None resulted in no entry in `self.fields`. This has the primary side effect that the __repr__ of an object now shows all the attributes that are set to None.  This affects the __repr__ of all objects with optional fields. For example:
```
  Fields:
    comments: no comments
-   continuity: None
-   control: None
-   control_description: None
    conversion: 1.0
    data: [1000 2000 3000]
    description: no description
    interval: 1
    offset: 0.0
-   rate: None
    resolution: -1.0
-   starting_time: None
    timestamps: [1. 2. 3.]
    timestamps_unit: seconds
    unit: unit
```
The attributes marked with a `dash` are new additions to the __repr__. I think it is good to be more transparent and complete. If you call `timeseries.rate`, when `rate` was not set, None will be returned. In the case of `TimeSeries` where (timestamps) or (starting_time and rate) should be set, this might be confusing to users.

I made this change to allow users to reset an attribute to None and make the behavior not dependent on the current state of the container. What do you think? Is this a worthwhile change? 

If not, I can add logic in the setters that says if the value is None AND the field is not set, do nothing and return.

bendichter:
Thanks for tackling this, @rly ! I'm testing incorporation with https://github.com/catalystneuro/neuroconv/pull/475 now..

bendichter:
While I like the ability to change values, I think too much freedom could be confusing for users. For instance, now a user is able to run:

```python
from pynwb.testing.mock.base import mock_TimeSeries

ts = mock_TimeSeries(timestamps=[1,2,3,4], rate=None)
ts.rate = 5
```

This will create a `TimeSeries` object that has both timestamps and rate, which is illegal, but this won't be flagged until the user goes to write.

Would it be possible to perform validation every time a `Container` class is modified? Or maybe we can make it so only values that have been set as non-`None` can be modified. Or maybe there might be some other way to do this?

rly:
@bendichter Good point. Yes, we can do that - we can call a generic validation function in the generic setter. Moving all validation in PyNWB classes to that function and making it work for auto-generated classes will take some more work though, but I think ultimately this would be a large improvement in usability.

For now, what if we constrain the changes in this PR to allow reassigning only `Data.data`? Type and shape validation will not occur (but it doesn't occur anyway if you pass in a DataIO object).

bendichter:
> For now, what if we constrain the changes in this PR to allow reassigning only Data.data? Type and shape validation will not occur (but it doesn't occur anyway if you pass in a DataIO object).

Sounds good to me!

bendichter:
wait, no, I need to overwrite `TimeSeries.data` and `TimeSeries.timestamps` as well. In fact, I'm not able to do that with the current branch:

```python
from pynwb.testing.mock.base import mock_TimeSeries

ts = mock_TimeSeries(timestamps=[1,2,3,4], rate=None)
ts.rate = 5  # this works
ts.data = [2,3,4,5]  # this does not work
```

```pytb
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [6], line 7
      5 ts = mock_TimeSeries(timestamps=[1,2,3,4], rate=None)
      6 ts.rate = 5
----> 7 ts.data = [2,3,4,5]

AttributeError: can't set attribute
```

rly:
Note: Now that the IO object is available on Container objects, for attributes, we may be able to easily modify any written attributes. We would need to store the path to the attribute and have a setter that detects if it was read from a file, the file is writeable, and then go to the object, and overwrite the value of the attribute.

PR #872, opened 2023-06-03
Addition of ER EntityKeyTable and bug fixes

mavaylon1:
## Motivation

1. Addition of EntityKeyTable and related methods/tests
2. Fixed bug where reusing a key object wouldn't be recorded in ObjectKeyTable

## How to test the behavior?
```
```

## Checklist


src/hdmf/common/resources.py
```diff
@@ -694,15 +754,27 @@ def to_norm_tsv(self, **kwargs):
...
+            df = self.children[i].to_dataframe()
+            df.to_csv(files[i], sep='\t', index=False)
+
+        with zipfile.ZipFile('er.zip', 'w') as zipF:
+          for file in files:
+              zipF.write(file)
+
+        # remove tsv files
+        for file in files:
+            remove_test_file('./'+file)
```
```suggestion
            if os.path.exists(path):
                os.remove('./'+file)
```

The `remove_test_file` function is a utility that is intended for use with test files. It has functionality that is slightly different than the above and it may be modified for the purpose of removing test files as the name suggests. The usage here in `to_norm_tsv` is different and so we should not use `remove_test_file`.

src/hdmf/common/resources.py
```diff
@@ -512,12 +553,45 @@ def add_ref(self, **kwargs):
...
+            # check if the key and object have been related in the ObjectKeyTable
 
+        entity = self.get_entity(entity_id=entity_id)
+        if isinstance(entity, bool):
+            entity = self._add_entity(entity_id, entity_uri)
+            self._add_entity_key(entity, key)
+        else:
+            # check for entity-key relationship in EntityKeyTable
+            key_idx = key.idx
+            entity_key_row_idx = self.entity_keys.which(keys_idx=key_idx)
```
rly, line 589: Can you add tests for this else condition?

src/hdmf/common/resources.py
```diff
@@ -512,12 +553,45 @@ def add_ref(self, **kwargs):
...
+            if len(object_key_row_idx)!=0:
+                obj_key_check = False
+                for row_idx in object_key_row_idx:
+                    obj_idx = self.object_keys['objects_idx', row_idx]
+                    if obj_idx == object_field.idx:
+                        obj_key_check = True
+                if not obj_key_check:
+                    self._add_object_key(object_field, key)
+            else:
+                self._add_object_key(object_field, key)
```
rly: Can you add a test? It will be immensely easier to add a test now than to do it in two years. We have a target of 95% code coverage as part of the U24 grant.

docs/gallery/plot_external_resources.py
```diff
@@ -306,17 +306,24 @@ def __init__(self, **kwargs):
...
 # ------------------------------------------------------
 # Users can read :py:class:`~hdmf.common.resources.ExternalResources` from the tsv format
-# by providing the path to the file.
-
-er_read = ExternalResources.from_flat_tsv(path='./er_example.tsv')
-remove_test_file('./er_example.tsv')
+# by providing the path to the directory.
+
+er_read = ExternalResources.from_norm_tsv(path='./')
+remove_test_file('./entities.tsv')
```
rly: Are all of these necessary? It seems like it is part of `to_norm_tsv` now.

src/hdmf/common/resources.py
```diff
@@ -243,19 +268,15 @@ def _add_file(self, **kwargs):
         file_object_id = kwargs['file_object_id']
         return File(file_object_id, table=self.files)
 
-    @docval({'name': 'key', 'type': (str, Key), 'doc': 'The key to associate the entity with.'},
-            {'name': 'entity_id', 'type': str, 'doc': 'The unique entity id.'},
+    @docval({'name': 'entity_id', 'type': str, 'doc': 'The unique entity id.'},
             {'name': 'entity_uri', 'type': str, 'doc': 'The URI for the entity.'})
     def _add_entity(self, **kwargs):
         """
         Add an entity that will be referenced to using the given key.
```
```suggestion
        Add an entity that will be referenced to using keys specified in ExternalResources.entity_keys.
```

src/hdmf/common/resources.py
```diff
@@ -424,6 +454,15 @@ def get_key(self, **kwargs):
...
                 return self.keys.row[key_idx_matches[0]]
 
+    @docval({'name': 'entity_id', 'type': str, 'doc': 'The ID for the identifier at the resource.'})
+    def get_entity(self, **kwargs):
+        entity_id = kwargs['entity_id']
+        entity = self.entities.which(entity_id=entity_id)
+        if len(entity)>0:
+            return self.entities.row[entity[0]]
+        else:
+            return False
```
```suggestion
            return None
```

A `get_x` method typically returns None if the object is not found or raises a ValueError.

src/hdmf/common/resources.py
```diff
@@ -512,12 +551,58 @@ def add_ref(self, **kwargs):
...
+                        obj_key_check = True
+                if not obj_key_check:
+                    self._add_object_key(object_field, key)
+            else:
+                msg = "Cannot find key object. Create new Key with string."
+                raise ValueError(msg)
+            # check if the key and object have been related in the ObjectKeyTable
 
+        entity = self.get_entity(entity_id=entity_id)
+        if isinstance(entity, bool):
```
```suggestion
        if entity is None:
```
Part of the suggested change above.

src/hdmf/common/resources.py
```diff
@@ -693,46 +764,68 @@ def to_norm_tsv(self, **kwargs):
...
                 continue
 
         # we need to check the idx columns in entities, objects, and object_keys
-        keys_idx = entities['keys_idx']
-        for idx in keys_idx:
-            if not int(idx) < keys.__len__():
-                msg = "Key Index out of range in EntityTable. Please check for alterations."
+        entity_idx = entity_keys['entities_idx']
+        for idx in entity_idx:
+            if not int(idx) < entities.__len__():
```
```suggestion
            if int(idx) >= len(entities):
```
It is much more common to call `len(x)` than `x.__len__()` (even though len(x) calls `x.__len__()`). Following conventions will help future maintainers understand the code. Deviating from convention suggests that there is some reason `len` is not being called.

Also, IMO `if x >= y` is easier to understand than `if not x < y`

src/hdmf/common/resources.py
```diff
@@ -753,130 +846,17 @@ def from_norm_tsv(cls, **kwargs):
                 msg = "Key Index out of range in ObjectKeyTable. Please check for alterations."
                 raise ValueError(msg)
 
+        keys_idx = entity_keys['keys_idx']
+        for idx in keys_idx:
+            if not int(idx) < keys.__len__():
```
```suggestion
            if int(idx) >= len(keys):
```

rly: same for this and the other uses of __len__ above (sorry I didn't catch these earlier)

rly: can you update the other three uses of `__len__` in this file?

mavaylon1: Done

rly (approved):

oruebel:
@mavaylon1 there are a few places in the `plot_external_resources.py` tutorial that should be updated to also list the new table: 

* https://github.com/hdmf-dev/hdmf/blob/07a42c295f7843832dd724aa9a6bcd5f8828e39a/docs/gallery/plot_external_resources.py#L38-L39  
* https://github.com/hdmf-dev/hdmf/blob/07a42c295f7843832dd724aa9a6bcd5f8828e39a/docs/gallery/plot_external_resources.py#L211C1-L211C1

PR #874, opened 2023-06-16
allow a `datetime.date` object to be used instead of `datetime.datetime` for isodatetime data

### LINKED ISSUE #321: Need field for specifying session start time accuracy
The `NWBFile` dataset `session_start_time` is designed to represent the real-world start time of the session. Based on discussions with @oruebel, I see the value in this field is that you can tell what time of day a dataset was collected, if it occurred at the same time as an earthquake, etc. 

But, most systems are not going to store times totally accurately with respect to the world clock at atomic resolution. How do we represent the inaccuracy of the session start time? Most datasets supply precision down to the second. Some datasets, for one reason or another, only know what day that the dataset was collected and thus enter 00:00:00 for the time, e.g. for this DataJoint to NWB converter:
https://github.com/vathes/DJ-NWB-Li-Daie-2015-2016/blob/ae69335a23e6a3c0c3809a54de5480baf8848712/pipeline/export/datajoint_to_nwb.py#L23

@oruebel and I think we should add a float attribute on `session_start_time` called `accuracy` that represents the +/- amount of time to which the `session_start_time` is estimated to be inaccurate. e.g., if the user thinks the value is accurate down to the second, then they should enter 0.5 for the accuracy. If they only know the day, then they should enter noon of that day as the start time and 43200 (12x60x60) for the accuracy. It would be optional, the default value would be np.nan, and we should encourage users to specify this value in tutorials and our best practices document.

This is a minor issue, but we should address it, or else, the `session_start_time` field will not have a consistent usage and meaning across files.

bendichter:
## Motivation

fix https://github.com/NeurodataWithoutBorders/nwb-schema/issues/321

thoughts, @rly ?

## Checklist


rly (dismissed):
Looks good to me! Thanks.

rly (approved):

PR #880, opened 2023-06-27
TermSet Integration with up-to-date HDMF

mavaylon1:
## Motivation

- [x] Add gallery doc on TermSet
- [x] Add data validation to Data
- [x] Add TermSet unit tests
- [x] Make sure everything works with ER and pynwb branch
- [x] Add tests for TermSet usage: data validation on Data
- [x] Add validation to DynamicTable
- [x] Add tests for TermSet usage: data validation on add_row
- [x] Add tests for TermSet usage: data validation on add_column
- [x] Add tests for TermSet usage: ExternalResources
## How to test the behavior?
```
```

## Checklist


mavaylon1:
@rly to address your concern of LinkML being not a lightweight dependency, in the future we will be using DynamicEnums and so might as well use LinkML parsing tools instead of making our own.

docs/gallery/example_term_set.yaml
```diff
@@ -0,0 +1,23 @@
+id: pynert/termset/species_example
+name: Species
```
oruebel, line 2: Can we also add a `version: 0.0.1` here to make sure all TermSets are versioned?

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,33 @@ def __init__(self, **kwargs):
...
+try:
+    import linkml_runtime  # noqa: F401
+    LINKML_INSTALLED = True
+except ImportError:
+    LINKML_INSTALLED = False
+
+if LINKML_INSTALLED:
+    from hdmf.term_set import TermSet
+
+    terms = TermSet(name='Species', term_schema_path='docs/gallery/example_term_set.yaml')
```
oruebel: Why is the `name` parameter needed? It looks like the name is already specified in the YAML. I think in most cases, we can assume that there is a single enum in the path file, so in that case, I think we should not need to the name.

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,147 @@
...
+# ----------------------------------------------------
+from hdmf.common import DynamicTable, VectorData
+try:
+    import linkml_runtime  # noqa: F401
+    LINKML_INSTALLED = True
+except ImportError:
+    LINKML_INSTALLED = False
+
+if LINKML_INSTALLED:
+    from hdmf.term_set import TermSet
```
oruebel: If LINKML is strictly optional, then maybe we can add the logic for checking whether TermSets are usable in the `hdmf.term_set` module, e.g., by having a flag `from hdmf.term_set import TERMSET_INSTALLED`.

oruebel: For the tutorials, I'm wondering whether we can just raise an error here if LINKML is not installed and say that this tutorial will not be rendered. Having the `if LINKML_INSTALLED` logic in every cell seems distracting.

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,147 @@
...
+
+######################################################
+# Viewing TermSet values
+# ----------------------------------------------------
+# :py:class:`~hdmf.TermSet` has methods to retrieve terms. The :py:func:`~hdmf.TermSet:view_set`
+# method will return a dictionary of all the terms and the corresponding information for each term.
+# Users can index specific terms from the :py:class:`~hdmf.TermSet`.
+if LINKML_INSTALLED:
+    terms = TermSet(name='Species', term_schema_path='docs/gallery/example_term_set.yaml')
+    terms.view_set
```
oruebel: I would print the output of `view_set` so a user can see the result in the tutorial.

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,147 @@
...
+    data = VectorData(
+        name='species',
+        description='...',
+        data=['Homo sapiens'],
+        term_set=terms)
+
+######################################################
+# Validate on append with TermSet
+# ----------------------------------------------------
+# As mentioned prior, when the term_set attribute is set all new data is validated. This true for both
```
oruebel:
```suggestion
# As mentioned prior, when the term_set attribute is set all new data is validated. This is true for both
```

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,33 @@ def __init__(self, **kwargs):
...
+        description='...',
+        data=['Homo sapiens', 'Ursus arctos horribilis'],
+        term_set=terms,
+    )
+
+    species = DynamicTable(name='species', description='My species', columns=[col1],)
+    er.add_ref_term_set(file=file,
+                        container=species,
+                        attribute='Species_Data',
+                       )
```
oruebel: Maybe this is part of PyNWB, but do we have a function to call `add_ref_term_set` for all objects that have term_set linked to it? E.g., could I call some version of `add_ref_term_set` with the `species` DynamicTable object and it would figure out all the columns in the table that have a term set and add them to the ExternalResources?

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,147 @@
+"""
+TermSet
+=======
+
+This is a user guide for interacting with the
+:py:class:`~hdmf.TermSet` class. The TermSet type
```
oruebel:
```suggestion
:py:class:`~hdmf.TermSet` class. The :py:class:`~hdmf.TermSet` type
```

src/hdmf/common/resources.py
```diff
@@ -2,9 +2,11 @@
...
 from . import get_type_map
-from ..container import Table, Row, Container, AbstractContainer, ExternalResourcesManager
+from ..container import Table, Row, Container, AbstractContainer, Data, ExternalResourcesManager
+from ..data_utils import DataIO
 from ..utils import docval, popargs, AllowPositional
 from ..build import TypeMap
+from ..term_set import TermSet
 from glob import glob
 import os
 import zipfile
```
oruebel, line 12: Not critical for this PR, but it would be nice to add `_repr_html_` function similar to #883 for TermSet as well to be able to render term sets nicely in the docs and Jupyter notebooks. Similarly, it would be nice to create a `__repr__` to make a nice plain-text rendering for TermSets.

src/hdmf/common/resources.py
```diff
@@ -405,6 +407,78 @@ def _get_file_from_container(self, **kwargs):
...
+            {'name': 'term_set', 'type': TermSet, 'default': None,
+             'doc': 'The TermSet to be used if the container/attribute does not have one.'}
+            )
+    def add_ref_term_set(self, **kwargs):
+        file = kwargs['file']
+        container = kwargs['container']
+        attribute = kwargs['attribute']
+        key = kwargs['key']
+        field = kwargs['field']
+        term_set = kwargs['term_set']
```
oruebel, line 430: Can you please add tests for TermSet. It looks like there are many uncovered lines.

src/hdmf/container.py
```diff
@@ -273,6 +274,20 @@ def generate_new_id(self, **kwargs):
             for c in self.children:
                 c.generate_new_id(**kwargs)
 
+    @docval({'name': 'term', 'type': str, 'doc': "term to be validated"},
+            {'name': 'term_set', 'type': TermSet, 'doc': 'the set of terms used to validate data on add'})
+    def validate_data(self, **kwargs):
+            """
```
oruebel: If a user must provide both the `term` and `term_set`, what is the utility of having this as part of the `AbstractContainer` class? Wouldn't it be simpler to have a `validate_term` method on the `TermSet`, i.e., call `term_set.validate(term)` instead.

src/hdmf/common/resources.py
mavaylon1: Some of the code cov lines don't make sense. Some ask for tests to cover the setting of the kwargs, while some are covered by tests but it doesn't pick up.

src/hdmf/container.py
```diff
@@ -535,11 +550,26 @@ class Data(AbstractContainer):
...
     def __init__(self, **kwargs):
         data = popargs('data', kwargs)
+        self.term_set = popargs('term_set', kwargs)
         super().__init__(**kwargs)
-        self.__data = data
+        if self.term_set is not None:
+            bad_data = []
+            for term in data:
+                if self.validate_data(term=term, term_set=self.term_set):
+                    continue
```
oruebel: Similar to my comment above, if `term_set` is already a member of `Data` then as a user I would expect to be able to call `validate_data` with just the term, i.e., `validate_data` should use `self.term_set` by default. Also, if `term=None`, I think `validate_data` could just do the iteration over all terms directly.

src/hdmf/container.py
```diff
@@ -535,11 +550,26 @@ class Data(AbstractContainer):
...
+            {'name': 'data', 'type': ('scalar_data', 'array_data', 'data'), 'doc': 'the source of the data'},
+            {'name': 'term_set', 'type': TermSet, 'doc': 'the set of terms used to validate data on add',
+             'default': None})
     def __init__(self, **kwargs):
         data = popargs('data', kwargs)
+        self.term_set = popargs('term_set', kwargs)
         super().__init__(**kwargs)
-        self.__data = data
+        if self.term_set is not None:
+            bad_data = []
```
```suggestion
            bad_data = [term for term in data if not  self.validate_data(term=term, term_set=self.term_set)]
```

I think this would be more efficient and readable in Python as a list comprehension

src/hdmf/common/resources.py
oruebel: Sorry, this comment applies to TermSet below, I'll close this thread and move it to the correct place.

src/hdmf/term_set.py
```diff
@@ -0,0 +1,91 @@
+from collections import namedtuple
+
+
```
oruebel, line 4: Not critical for this PR, but it would be nice to add `_repr_html_` function similar to #883 for TermSet as well to be able to render term sets nicely in the docs and Jupyter notebooks. Similarly, it would be nice to create a `__repr__` to make a nice plain-text rendering for TermSets.

src/hdmf/common/resources.py
mavaylon1: I believe this is due to the skip because it's highlighting the whole method as untested even though there's tests. @oruebel

src/hdmf/term_set.py
```diff
@@ -0,0 +1,91 @@
+from collections import namedtuple
```
oruebel, line 1: I'm wondering whether we should move `ExternalResources`, `ExternalResourceMangager`, `TermSet` etc. codes into a separate module (folder) for external_resources. I just think it would be easiest to keep those things in one place. However, that would break with the logic of having things defined in the common schema in the `hdmf.common` folder. Maybe have a module (folder) in `hdmf.common`, e.g,. `hdmf.common.external_resources`. @rly what do you think?

src/hdmf/common/resources.py
oruebel: Or are the tests not being run on GitHub because LinkML is not installed?

mavaylon1: Most likely. We left it as an optional dependency per Ryan's advice of linkml being a heavy dependency.

oruebel:
@mavaylon1 does is also work for arbitrary attributes of a Container, e.g., for the `species` dataset in NWB? It's fine if that is going to be part of a separate PR, I just wanted check.

mavaylon1:
> @mavaylon1 does is also work for arbitrary attributes of a Container, e.g., for the `species` dataset in NWB? It's fine if that is going to be part of a separate PR, I just wanted check. 

Validation only works for datasets as of now. We haven't figured out how to do so more broadly to include validating attributes. The issue being that the data has the termset parameter, and so each column will hold/have it's own unique set. However, I'm not sure how that would work for attributes, especially if an object has multiple attributes that require multiple term sets (unless we require attributes in an object to share). @oruebel

oruebel:
> We haven't figured out how to do so more broadly to include validating attributes

Thanks for the clarification. Makes sense. Let's plan to chat about possible designs for this next week. A couple of initial thoughts on how to possible do this: 
* One way could be to wrap the values of the attribute, similar to what we do with [`DataIO`](https://github.com/hdmf-dev/hdmf/blob/6043e776e2daf6fa03aa4063609fceb4689d83ab/src/hdmf/data_utils.py#L952C1-L952C14). In principle, I guess this could even be added to `DataIO` itself, but my initial guess is that (if we wrap values) that a separate wrapper class may be more appropriate. 
* Another way could be to have a `map` on the Container level to link attributes to term sets. I.e., in this case the Container would need to manage which attributes have term set. Validation may be tricky, but I think this could work as well, as long as we have setters for the attributes so we can do validation. 
* Alternatively, similar to the ExternalResourcesManager we could have a TermSetManager on the root Container to manage all linkages of fields to tem sets. Not sure if that's a good idea, but it's an option.

docs/gallery/example_term_set.yaml
mavaylon1: Sure. This isn't in the format LinkML specifies in the tutorials, but I think it is a good addition.

src/hdmf/container.py
mavaylon1: I think it would be the same level of complexity. Both are just term lookups. Having it on AbstractContainer could help us down the line expanding it to attributes, but I don't think a change is warranted.

src/hdmf/container.py
mavaylon1: Data is validated when a new instance of Data is created and is not supported if the data already exists. However, add_ref_termset will return "missing terms" from the data if the term is invalid to the termset. The user never really has to call "validate_data" directly. The method is called in the presence of a term_set for VectorData (if it is provided) and also with add_ref_term_set. Since the user doesn't call it directly, I don't think the term_set parameter being explicit matters.

docs/gallery/plot_term_set.py
mavaylon1: Wouldn't raising an error always lead to the gallery tests failing on git.

src/hdmf/container.py
mavaylon1: On second thought, if the user never calls it directly, then the method could be private and use self.term_set in the method rather than "termset=self.termset" and setting the kwarg later on.

src/hdmf/container.py
mavaylon1: Being on AbstractContainer allows, validate to be used outside of just Data, possibly with attributes. However, thus far Data is the only one using it and even though it stores the termset, I am not sure to remove the termset parameter. Are you suggesting moving it to the TermSet class and everything (both now and future) will call validate from there? I think I see the simplification that comes from that. I'll start implementing and see if there's any issues we didn't see.

mavaylon1:
@oruebel @rly 
I'm bringing all the comments here
1. Update validate data so that it exists on termset --> Done
2. Some tests don't run on git due to Linkml being optional and so coverage is not picking them up. Is this ok or do we want to add linkml to requirments.txt.
3. Linkml dependency: For the gallery it seems we need the conditional to make sure it doesn't fail on git unless I put the whole tutorial in a single if statement/try statement.
4. @oruebel mentioned the idea of moving ExternalResources, ExternalResourceMangager, TermSet etc. codes into a separate module (folder) for external_resources.

I'd really like to merge this week so I can then merge the pynwb.

oruebel:
> 2\. Some tests don't run on git due to Linkml being optional and so coverage is not picking them up. Is this ok or do we want to add linkml to requirments.txt.

I think adding LinkML as dependency for testing on GitHub is fine. @rly  won't we need LinkML as a dependency for Conda and Pip releases or do we (for now) expect that users install LinkML separately?

> 3\. Linkml dependency: For the gallery it seems we need the conditional to make sure it doesn't fail on git unless I put the whole tutorial in a single if statement/try statement.

We should add LinkML to the ReadTheDocs build to make sure the tutorial is being rendered there. For skipping the tutorial if LinkML is not installed, I think a cleaner approach may be to skip it in `conf.py` for the docs build, possibly as part of the [``exclude_patterns``](https://github.com/hdmf-dev/hdmf/blob/5269c1f632abab2df074ccda60c346735c2c141c/docs/source/conf.py#L124C1-L125C41). I think something along the following lines should work: 

```
# List of patterns, relative to source directory, that match files and
# directories to ignore when looking for source files.
exclude_patterns = ["_build", "test.py"]

try:
    import linkml_runtime  # noqa: F401
except ImportError:
    exclude_patterns += ["docs/gallery/plot_term_set.py",]
    # Possibly add a warning here to make it easier to spot that the tutorial was skipped 
```
In this way we won't need any checks in the tutorial iteself to check if LinkML is available. If we want to make it clear that the tutorial is available but was not built, then we could add a "dummy" tutorial with the same tile, but the tutorial would simply say that the tutorial was not rendered because LinkML was not installed and have a link to the full code of the tutorial. I.e., if LinkML is available we would render the real tutorial and if LinkML is not available then we would render the "dummy" tutorial that would just point to the code of the full tutorial.

oruebel:
> 4\. @oruebel mentioned the idea of moving ExternalResources, ExternalResourceMangager, TermSet etc. codes into a separate module (folder) for external_resources.

@rly what do you think?

mavaylon1:
1. The patch coverage is solved with the updated reqs.
2. For the gallery tests, adding linkml to the requirements solves any test issues and so I removed the if statements to check for linkml. As for the plot_term_set.py in general, I think it's cleaner to "exit" if linkml is not installed. (refer to changes in plot_term_set.py)
3. The remaining issues are in regards to the docs. If linkml is not installed in the requirements_docs then it should be skipped in the conf.py as @oruebel suggested, but it doesn't seem to skip and throws the error that link_ml is not installed. If linkml is in the reqs_docs.txt file then it does a weird path for both the build and external links "/home/runner/work/hdmf/hdmf/docs/gallery/docs/gallery/example_term_set.yaml". It doubles the hdmf and docs/gallery. 

I saw in conf.py 
```
# If extensions (or modules to document with autodoc) are in another directory,
# add these directories to sys.path here. If the directory is relative to the
# documentation root, use os.path.abspath to make it absolute, like shown here.
# sys.path.insert(0, os.path.abspath('.'))
```
I'll try adding "sys.path.insert(0, os.path.abspath('../gallery'))". 
-->UPDATE: This does not help.
-->Update 2: Made some changes to how the path is acquired. Seems to work.

mavaylon1:
The termset.yaml probably shouldn't be in gallery since it's not a tutorial. Should we move it to source?

oruebel:
> 3\. The remaining issues are in regards to the docs

For sphinx gallery I think you will probably need to use:

sphinx_gallery_conf = {
    ...
    'filename_pattern': ...
    'ignore_pattern': ...
}

to skip a gallery instead of  the``exclude_patterns``, which I think is for Sphinx instead of sphinx gallery. See here for more details on this https://sphinx-gallery.github.io/stable/configuration.html#parsing-and-executing-examples-via-matching-patterns

mavaylon1:
@oruebel can you take a look at the changes. I went with a different approach.

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,40 @@ def __init__(self, **kwargs):
...
+    from hdmf.term_set import TermSet
+
+try:
+    dir_path = os.path.dirname(os.path.abspath(__file__))
+    yaml_file = os.path.join(dir_path, 'example_term_set.yaml')
+except NameError:
+    dir_path = os.path.dirname(os.path.abspath('.'))
+    yaml_file = os.path.join(dir_path, 'gallery/example_term_set.yaml')
+
+terms = TermSet(term_schema_path=yaml_file)
```
oruebel, line 343: I think this will fail if `LINKML_INSTALLED` is False, because `TermSet` is only imported if `LINKML_INSTALLED` is true.

oruebel:
> I went with a different approach.

Can you clarify. Are you refering to calling ``sys.exit` in the tutorial if LINKML is not installed?

mavaylon1:
The sys.exit and the different ways to handle the path for the yaml.

Everything passes, let me know what you think so we can merge.

docs/gallery/example_term_set.yaml
```diff
@@ -0,0 +1,24 @@
...
+  NCBI_TAXON: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=
+imports:
+  - linkml:types
+default_range: string
+
+enums:
+  Species:
+    permissible_values:
+      Homo sapiens:
+        description: description
```
rly:
```suggestion
        description: the species is human
```
Please add a description for these instead of a placeholder

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,40 @@ def __init__(self, **kwargs):
...
+# entity_id and entity_uri parameters. :py:class:`~hdmf.common.resources.Key` values will have
+# to match the name of the term in the :py:class:`~hdmf.TermSet`.
+try:
+    import linkml_runtime  # noqa: F401
+    LINKML_INSTALLED = True
+except ImportError:
+    LINKML_INSTALLED = False
+
+if LINKML_INSTALLED:
+    from hdmf.term_set import TermSet
```
rly:
```suggestion
import linkml_runtime
from hdmf.term_set import TermSet
```

Can't we just assume that linkml is installed for these gallery examples? If we put linkml in `requirements-doc.txt` then it will be available on readthedocs. It will also be available when devs build the documentation because they are supposed to install deps from `requirements-doc.txt`. Note that `requirements-doc.txt` contains other deps that are normally not installed like matplotlib.

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,34 @@ def __init__(self, **kwargs):
...
 os.remove('./er.zip')
+
+###############################################################################
+# Using TermSet with ExternalResources
+# ------------------------------------------------
+# :py:class:`~hdmf.TermSet` allows for an easier way to add references to
+# :py:class:`~hdmf.common.resources.ExternalResources`. These enumerations take place of the
+# entity_id and entity_uri parameters. :py:class:`~hdmf.common.resources.Key` values will have
+# to match the name of the term in the :py:class:`~hdmf.TermSet`.
+import linkml_runtime
```
rly:
```suggestion
```
I guess this import is not needed anymore.

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,34 @@ def __init__(self, **kwargs):
 
 er_read = ExternalResources.from_norm_tsv(path='./')
 os.remove('./er.zip')
+
+###############################################################################
+# Using TermSet with ExternalResources
+# ------------------------------------------------
+# :py:class:`~hdmf.TermSet` allows for an easier way to add references to
```
rly:
```suggestion
# :py:class:`~hdmf.term_set.TermSet` allows for an easier way to add references to
```
The link in the generated docs for this PR does not work using :py:class:`~hdmf.TermSet`. I think you need to add the name of the module.

docs/gallery/plot_external_resources.py
```diff
@@ -320,3 +323,34 @@ def __init__(self, **kwargs):
...
+# :py:class:`~hdmf.common.resources.ExternalResources`. These enumerations take place of the
+# entity_id and entity_uri parameters. :py:class:`~hdmf.common.resources.Key` values will have
+# to match the name of the term in the :py:class:`~hdmf.TermSet`.
```
rly:
```suggestion
# to match the name of the term in the :py:class:`~hdmf.term_set.TermSet`.
```

docs/gallery/example_term_set.yaml
```diff
@@ -0,0 +1,24 @@
...
+default_range: string
+
+enums:
+  Species:
+    permissible_values:
+      Homo sapiens:
+        description: the species is human
+        meaning: NCBI_TAXON:9606
+      Mus musculus:
+        description: description
```
rly:
```suggestion
        description: the species is a house mouse
```

rly: can you change the other ones too...?

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
+"""
+TermSet
+=======
+
+This is a user guide for interacting with the
+:py:class:`~hdmf.TermSet` class. The :py:class:`~hdmf.TermSet` type
```
rly: See other comment that :py:class:`~hdmf.TermSet` doesn't link correctly.

Also, if we move these into an external_resources.py module, all of these links will have to be updated...

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+-------------
+The :py:class:`~hdmf.TermSet` class provides a way for users to create their own
+set of terms from brain atlases, species taxonomies, and anatomical, cell, and
+gene function ontologies.
+
+:py:class:`~hdmf.TermSet` serves two purposes: data validation and external reference
+management. Users will be able to validate their data to their own set of terms, ensuring
+clean data to be used inline with the FAIR principles later on.
+The  :py:class:`~hdmf.TermSet` class allows for a reusable and sharable
+pool of metadata to serve as references to any dataset within the NWB ecosystem.
```
```suggestion
pool of metadata to serve as references to any dataset.
```

This is a general feature - not just for NWB (though that is our primary use case).

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+The :py:class:`~hdmf.TermSet` class is used closely with
+:py:class:`~hdmf.common.resources.ExternalResources` to more efficiently map terms
+to data. Please refer to the tutorial on ExternalResources to see how :py:class:`~hdmf.TermSet`
+is used with :py:class:`~hdmf.common.resources.ExternalResources`.
+
+:py:class:`~hdmf.TermSet` is built upon the resources from LinkmL, a modeling
```
```suggestion
:py:class:`~hdmf.TermSet` is built upon the resources from LinkML, a modeling
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+is used with :py:class:`~hdmf.common.resources.ExternalResources`.
+
+:py:class:`~hdmf.TermSet` is built upon the resources from LinkML, a modeling
+language to create YAML schemas, giving :py:class:`~hdmf.TermSet`
+a standardized structure and a variety of tools to help the user manage their references.
+
+How to make a TermSet Schema
+----------------------------
+Before the user can take advantage of all the wonders within the
+:py:class:`~hdmf.TermSet` class, the user needs to a LinkML schema (YAML) that provides
```
```suggestion
:py:class:`~hdmf.TermSet` class, the user needs to create a LinkML schema (YAML) that provides
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+all the permissible term values. Please refer to https://linkml.io/linkml/intro/tutorial06.html
+to learn more about how LinkML structures their schemas.
```
```suggestion
to learn more about how LinkML structures their schema.
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+clean data to be used inline with the FAIR principles later on.
+The  :py:class:`~hdmf.TermSet` class allows for a reusable and sharable
+pool of metadata to serve as references to any dataset within the NWB ecosystem.
+The :py:class:`~hdmf.TermSet` class is used closely with
+:py:class:`~hdmf.common.resources.ExternalResources` to more efficiently map terms
+to data. Please refer to the tutorial on ExternalResources to see how :py:class:`~hdmf.TermSet`
+is used with :py:class:`~hdmf.common.resources.ExternalResources`.
+
+:py:class:`~hdmf.TermSet` is built upon the resources from LinkML, a modeling
+language to create YAML schemas, giving :py:class:`~hdmf.TermSet`
```
```suggestion
language that uses YAML-based schema, giving :py:class:`~hdmf.TermSet`
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+# Viewing TermSet values
+# ----------------------------------------------------
+# :py:class:`~hdmf.TermSet` has methods to retrieve terms. The :py:func:`~hdmf.TermSet:view_set`
+# method will return a dictionary of all the terms and the corresponding information for each term.
+# Users can index specific terms from the :py:class:`~hdmf.TermSet`.
+try:
+    from hdmf.term_set import TermSet
+    terms = TermSet(term_schema_path=yaml_file)
+except ValueError:
+    sys.exit("linkml_runtime not installed.")
```
```suggestion
from hdmf.term_set import TermSet
terms = TermSet(term_schema_path=yaml_file)
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+except NameError:
+    dir_path = os.path.dirname(os.path.abspath('.'))
+    yaml_file = os.path.join(dir_path, 'gallery/example_term_set.yaml')
+
+######################################################
+# Viewing TermSet values
+# ----------------------------------------------------
+# :py:class:`~hdmf.TermSet` has methods to retrieve terms. The :py:func:`~hdmf.TermSet:view_set`
+# method will return a dictionary of all the terms and the corresponding information for each term.
+# Users can index specific terms from the :py:class:`~hdmf.TermSet`.
```
```suggestion
# Users can index specific terms from the :py:class:`~hdmf.TermSet`. The LinkML runtime will need to be installed.
# You can do so by first running ``pip install linkml-runtime``.
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+    sys.exit("linkml_runtime not installed.")
+print(terms.view_set)
+
+# Retrieve a specific term
+terms['Homo sapiens']
+
+######################################################
+# Validate Data with TermSet
+# ----------------------------------------------------
+# :py:class:`~hdmf.TermSet` has been integrated so that :py:class:`~hdmf.Data` and its
```
```suggestion
# :py:class:`~hdmf.term_set.TermSet` has been integrated so that :py:class:`~hdmf.container.Data` and its
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,146 @@
...
+)
+species = DynamicTable(name='species', description='My species', columns=[col1,col2])
+
+######################################################
+# Validate new rows in a DynamicTable with TermSet
+# ----------------------------------------------------
+# Validating new rows to :py:class:`~hdmf.common.table.DynamicTable` is simple. The
+# :py:func:`~hdmf.common.table.DynamicTable.add_row` method will automatically check each column for a
+# :py:class:`~hdmf.TermSet` (via the term_set attribute). If the attribute is set, the the data will be
+# validated for that column using that column's :py:class:`~hdmf.TermSet`. If their is invalid data, the
```
```suggestion
# validated for that column using that column's :py:class:`~hdmf.TermSet`. If there is invalid data, the
```

requirements-min.txt
```diff
@@ -2,8 +2,9 @@
 h5py==2.10  # support for selection of datasets with list of indices added in 2.10
 importlib-metadata==4.2.0; python_version < "3.8"  # TODO: remove when minimum python version is 3.8
 importlib-resources==5.12.0; python_version < "3.9"  # TODO: remove when when minimum python version is 3.9
-jsonschema==2.6.0
+jsonschema>=3.2.0
```
rly:
```suggestion
jsonschema==3.2.0
```
These minimum requirements should all be pinned to exact values. If a higher version is allowed, it will be used, and we will no longer be testing whether hdmf works with these minimum requirements.

requirements-min.txt
```diff
@@ -2,8 +2,9 @@
 h5py==2.10  # support for selection of datasets with list of indices added in 2.10
 importlib-metadata==4.2.0; python_version < "3.8"  # TODO: remove when minimum python version is 3.8
 importlib-resources==5.12.0; python_version < "3.9"  # TODO: remove when when minimum python version is 3.9
-jsonschema==2.6.0
+jsonschema==3.2.0
 numpy==1.16  # numpy>=1.16,<1.18 does not provide wheels for python 3.8 and does not build well on windows
 pandas==1.0.5  # when this is changed to >=1.5.0, see TODO items referenced in #762
 ruamel.yaml==0.16
 scipy==1.1  # scipy>=1.1,<1.4 does not provide wheels for python 3.8 and building scipy can fail due to incompatibilities with numpy
+linkml-runtime==1.5.0
```
rly, line 10:
```suggestion
linkml-runtime==1.5.0
tqdm==4.41.0
zarr==2.12.0
```
By convention, we have not been testing the minimum versions of optional dependencies. We can do so, but for consistency, we should also do so for tqdm and zarr.

requirements-dev.txt
```diff
@@ -13,3 +13,4 @@ python-dateutil==2.8.2
 ruff==0.0.265
 tox==4.5.1; python_version >= "3.8"
 tox==3.28.0; python_version < "3.8"  # tox 4+ has virtualenv requirements that are incompatible with other pkgs
+linkml-runtime==1.5.0
```
rly:
```suggestion
```
Since linkml-runtime is installed from `requirements.txt`, it is not needed here. This is just for any libraries that need to be installed for tests and checks to run.

requirements-doc.txt
```diff
@@ -4,3 +4,4 @@ sphinx>=4  # improved support for docutils>=0.17
 sphinx_rtd_theme>=1  # <1 does not work with docutils>=0.17
 sphinx-gallery
 sphinx-copybutton
+linkml-runtime==1.5.0
```
rly, line 7:
```suggestion
```
Similarly to the above, this is just for any libraries that need to be installed for docs to run that are not installed by `requirements.txt` or `pyproject.toml`

rly: Actually, never mind. I think it should not be in `requirements.txt`. Therefore, it should be here.

requirements.txt
```diff
@@ -11,3 +11,4 @@ pandas==1.3.5; python_version < "3.8"  # pandas 1.4 dropped python 3.7 support
 ruamel.yaml==0.17.24
 scipy==1.10.1; python_version >= "3.8"
 scipy==1.7.3; python_version < "3.8"   # scipy 1.8 dropped python 3.7 support
+linkml-runtime==1.5.0
```
rly:
```suggestion
```
Actually, never mind regarding my comments about the other requirements files. Sorry for the turnaround. I think it should not be in requirements.txt. Therefore, it should be in requirements-doc.txt. I'll add it back.

mavaylon1: It needs to be in requirements for testing, not just for docs. Otherwise codecov will throw an issue of coverage.

mavaylon1:
@rly everything is breaking. I'll revert some changes and start fresh tomorrow.

rly:
Let me take a quick try at it tonight. Setting up the requirements files right for testing can be tricky.

tests/unit/test_term_set.py
```diff
@@ -0,0 +1,46 @@
...
+    @unittest.skipIf(not LINKML_INSTALLED, "optional LinkML module is not installed")
+    def test_termset_validate_false(self):
+        termset = TermSet(term_schema_path='tests/unit/example_test_term_set.yaml')
+        self.assertEqual(termset.validate('missing_term'), False)
+
+    @unittest.skipIf(not LINKML_INSTALLED, "optional LinkML module is not installed")
+    def test_get_item(self):
+        termset = TermSet(term_schema_path='tests/unit/example_test_term_set.yaml')
+        self.assertEqual(termset['Homo sapiens'].id, 'NCBI_TAXON:9606')
+        self.assertEqual(termset['Homo sapiens'].description, 'description')
```
```suggestion
        self.assertEqual(termset['Homo sapiens'].description, 'the species is human')
```

rly:
> 4. @oruebel mentioned the idea of moving ExternalResources, ExternalResourceMangager, TermSet etc. codes into a separate module (folder) for external_resources.

This is fine to me.

rly:
Looks green to me!

rly (dismissed):
Looks good to me, but please wait for @oruebel to review as well before merging.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### New features and minor improvements
 - Updated `ExternalResources` to have EntityKeyTable with updated tests/documentation and minor bug fix to ObjectKeyTable. @mavaylon1 [#872](https://github.com/hdmf-dev/hdmf/pull/872)
 - Added warning for DynamicTableRegion links that are not added to the same parent as the original container object. @mavaylon1 [#891](https://github.com/hdmf-dev/hdmf/pull/891)
+- Added the `TermSet` class along with integrated validation methods for any child of `AbstractContainer`, e.g., `VectorData`, `Data`, `DynamicTable`. @mavaylon1 [#880](https://github.com/hdmf-dev/hdmf/pull/880)
+
```
oruebel, line 9:
```suggestion

### Documentation and tutorial enhancements:

- Added tutorial for the new `TermSet` class @mavaylon1 [#880](https://github.com/hdmf-dev/hdmf/pull/880)
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,143 @@
...
+1. The name of the schema is up to the user, e.g., the name could be "Species" if the term set will
+   contain species terms.
+2. The prefixes will be the standardized prefix of your source, followed by the URI to the terms.
+   For example, the NCBI Taxonomy is abbreviated as NCBI_TAXON, and Ensemble is simply Ensemble.
+   As mentioned prior, the URI needs to be to the terms; this is to allow the URI to later be coupled
+   with the source id for the term to create a valid link to the term source page.
+3. The schema uses LinkML enumerations to list all the possible terms. Currently, users will need to
+   manually outline the terms within the enumeration's permissible values.
+
+For a clear example, please refer to example_term_set.yaml within the tutorial gallery.
```
oruebel, line 47: On ReadTheDocs the files in the Git repo are not easily accessible for users. I would either add a link to point to the `example_term_set.yaml` file on GitHub or we could move line 55-60 (where you get the path to the file) before the "Creating an instance of a TermSet class" section and then simply do a pretty print of the YAML so that a user can see the content of the YAML.

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,143 @@
...
+data = VectorData(
+    name='species',
+    description='...',
+    data=['Homo sapiens'],
+    term_set=terms)
+
+######################################################
+# Validate on append with TermSet
+# ----------------------------------------------------
+# As mentioned prior, when the term_set attribute is set all new data is validated. This is true for both
```
oruebel:
```suggestion
# As mentioned prior, when the term_set attribute is set, then all new data is validated. This is true for both
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,143 @@
...
+# append and extend methods.
+data.append('Ursus arctos horribilis')
+data.extend(['Mus musculus', 'Myrmecophaga tridactyla'])
+
+######################################################
+# Validate Data in a DynamicTable with TermSet
+# ----------------------------------------------------
+# Validating data with :py:class:`~hdmf.common.table.DynamicTable` is determined by which columns were
+# initialized with the term_set attribute set. The data is validated when the columns are created and not
+# when set as columns to the table.
```
oruebel:
```suggestion
# initialized with the term_set attribute set. The data is validated when the columns are created or
# modified. Since adding the columns to a DynamicTable does not modify the data, validation is
# not being performed at that time.
```

src/hdmf/term_set.py
```diff
@@ -0,0 +1,96 @@
...
+from .utils import docval
+
+
+class TermSet():
+    """
+    Class for implementing term sets from ontologies.
```
oruebel:
```suggestion
    Class for implementing term sets from ontologies and other resources used to define the
    meaning and/or identify of terms.
```

src/hdmf/term_set.py
```diff
@@ -0,0 +1,97 @@
...
+class TermSet():
+    """
+    Class for implementing term sets from ontologies and other resources used to define the
+    meaning and/or identify of terms.
+
+    :ivar name: The name of the TermSet
+    :ivar term_schema_path: The LinkML YAML enumeration schema
+    :ivar ontology_version: The version of the ontology
+    :ivar ontology_source_name:  The name of the ontology
+    :ivar ontology_source_uri: The uri of the ontology
```
oruebel: Where are `ontology_version`, `ontology_source_name`, and `ontology_source_uri` define, I didn't see them in the code?

oruebel: Also `name` does not seem to be set in the code.

src/hdmf/term_set.py
```diff
@@ -0,0 +1,97 @@
...
+from .utils import docval
+
+
+class TermSet():
+    """
+    Class for implementing term sets from ontologies and other resources used to define the
+    meaning and/or identify of terms.
+
+    :ivar name: The name of the TermSet
+    :ivar term_schema_path: The LinkML YAML enumeration schema
```
oruebel, line 10: Add documentation for the `self.sources` instance variable

src/hdmf/term_set.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        try:
+            from linkml_runtime.utils.schemaview import SchemaView
+        except ImportError:
+            msg = "Install linkml_runtime"
+            raise ValueError(msg)
+        self.term_schema_path = term_schema_path
+        self.view = SchemaView(self.term_schema_path)
+        self.sources = self.view.schema.prefixes
+
```
oruebel, line 29: It would be useful to add instance variable or read-only properties for `name` and `version`. Possibly also for `prefixes` and `id` listed in the YAML.

src/hdmf/term_set.py
```diff
@@ -0,0 +1,97 @@
...
+    def __repr__(self):
+        re = "class: %s\n" % str(self.__class__)
+        re += "term_schema_path: %s\n" % self.term_schema_path
+        return re
+
+    def _perm_value_key_info(self, perm_values_dict: dict, key: str):
+        """
```
oruebel: Private functions should have a double `__` as prefix. A single underscore is "protected" in Python.

tests/unit/example_test_term_set.yaml
```diff
@@ -0,0 +1,24 @@
+id: pynert/termset/species_example
```
oruebel:
```suggestion
id: termset/species_example
```

docs/gallery/example_term_set.yaml
```diff
@@ -0,0 +1,24 @@
+id: pynert/termset/species_example
```
oruebel:
```suggestion
id: termset/species_example
```

docs/gallery/plot_term_set.py
```diff
@@ -0,0 +1,144 @@
...
+1. The name of the schema is up to the user, e.g., the name could be "Species" if the term set will
+   contain species terms.
+2. The prefixes will be the standardized prefix of your source, followed by the URI to the terms.
+   For example, the NCBI Taxonomy is abbreviated as NCBI_TAXON, and Ensemble is simply Ensemble.
+   As mentioned prior, the URI needs to be to the terms; this is to allow the URI to later be coupled
+   with the source id for the term to create a valid link to the term source page.
+3. The schema uses LinkML enumerations to list all the possible terms. Currently, users will need to
+   manually outline the terms within the enumeration's permissible values.
+
+For a clear example, please refer to example_term_set.yaml within the tutorial gallery.
```
oruebel:
```suggestion
For a clear example, please view the 
`example_term_set.yaml <https://github.com/hdmf-dev/hdmf/blob/dev/docs/gallery/example_term_set.yaml>`_ for this tutorial, which provides a concise example of how
a term set schema looks. 
```

On ReadTheDocs the files in the Git repo are not easily accessible for users. I would add a link to point to the file on GitHub. The link I added here is probably broken on the PR but should become valid once the PR is merged.

oruebel (commented):
Looks overall good to me. I made few edits and added a couple of suggestions. I requested a couple of minor changes on the TermSet class, but those should be quick fixes. Otherwise this looks like it is good to go.

src/hdmf/term_set.py
```diff
@@ -0,0 +1,95 @@
...
+from .utils import docval
+
+
+class TermSet():
+    """
+    Class for implementing term sets from ontologies and other resources used to define the
+    meaning and/or identify of terms.
+
+    :ivar term_schema_path: The LinkML YAML enumeration schema
+    :ivar sources: The prefixes for the ontologies used in the TermSet
```
oruebel, line 11:
```suggestion
    :ivar sources: The prefixes for the ontologies used in the TermSet
    :ivar view: SchemaView of the term set schema
```

oruebel (approved):
Thanks for the updates. Looks good to me. A couple of open items that can be part of a future PR (or this one, whichever you prefer):

- Moving ExternalResources, ExternalResourceMangager, TermSet etc. codes into a separate module (folder) for external_resources
- Add TermSet._repr_html_ function similar to #883 for to be able to render term sets nicely in the docs and Jupyter notebooks
- Add read-only properties to TermSet for “name”, “version”, and “prefixes” (which are probably just shortcuts for lookups in TermSet.view)

PR #882, opened 2023-06-28
add io attr

bendichter:
I'd like to be able to do `nwbfile.io.close()` and `pynwb.validate(nwbfile.io)`.

This would allow me to create a simple read function:

```python
nwbfile = read_nwb(...)

...

nwbfile.io.close()
```

As it currently stands, this is not possible because I can't recover and close the `io` object if my context only returns an `nwbfile`.


bendichter:
@oruebel and @rly, do you guys have thoughts on this? Do you see it causing any problems?

bendichter:
To give you an idea of what I am going for:

```python
from pathlib import Path

from typing import Optional, Literal

import h5py
from hdmf_zarr import NWBZarrIO
from pynwb import NWBHDF5IO


backend_classes = dict(hdf5=NWBHDF5IO, zarr=NWBZarrIO)


def _get_method(path: str):
    if path.startswith(("https://", "http://", "s3://")):
        return "fsspec"
    elif Path(path).is_file():
        return "local"
    else:
        raise ValueError(
            f"Unable to automatically determine method. Path {path} does not appear to be a URL and is not a file on "
            f"the local filesystem.")


def _init_fsspec(path):
    import fsspec
    if path.startswith(("https://", "http://")):
        return fsspec.filesystem("http")
    elif path.startswith("s3://"):
        return fsspec.filesystem("s3", anon=True)


def _get_backend(path: str, method: Literal["local", "fsspec", "ros3"]):
    if method == "ros3":
        return "hdf5"

    possible_backends = []
    if method == "fsspec":
        fs = _init_fsspec(path)
        with fs.open(path, "rb") as f:
            for backend, cls in backend_classes.items():
                if cls.can_read(f):
                    possible_backends.append(backend)
    else:
        for backend, cls in backend_classes.items():
            if cls.can_read(path):
                possible_backends.append(backend)

    if len(possible_backends) > 1:
        raise ValueError("more than one possible backend found: {possible_backends}")
    elif len(possible_backends) == 0:
        raise ValueError("No possible backend found.")

    return possible_backends[0]


def read_nwb(
    nwbfile_path: str,
    mode: str = "r",
    method: Optional[Literal["local", "fsspec", "ros3"]] = None,
    backend: Optional[Literal["hdf5", "zarr"]] = None
):
    method = method or _get_method(nwbfile_path)
    backend = backend or _get_backend(nwbfile_path, method)

    if method == "fsspec":
        fs = _init_fsspec(nwbfile_path)
        f = fs.open(nwbfile_path, "rb")
        file = h5py.File(f)
        io = backend_classes[backend](file=file, mode=mode, load_namespaces=True)
    else:
        io = backend_classes[backend](
            nwbfile_path, mode=mode, load_namespaces=True, driver="ros3" if method == "ros3" else None
        )

    return io.read()
```

oruebel:
Since the `io.read()` call is identical for all backends and since we need to be able to close the `io`, I'm wondering whether it would be easier to have `read_nwb` just return the `io` object and give the user the control when to call read and when to close. I.e., instead of:

```
nwbfile = read_nwb(...)
```

the pattern would be:

```
nwbio = get_nwb_io(...)   # same as read_nwb but returns the io object
nwbfile = nwbio.read()
nwbio.close()
```

CodyCBakerPhD:
> Since the io.read() call is identical for all backends and since we need to be able to close the io, I'm wondering whether it would be easier to have read_nwb just return the io object 

Alas, it is this very two-step process (determine IO, read file) that we are also hoping to alleviate with this convenience function

The goal is to have a simple one-liner

```python
nwbfile = read_nwbfile(path="path/to/my/file")
```

that anyone can use without knowledge of what type of file back-end to choose, without knowledge of streaming technicalities, and without needing to know anything about _how_ the file gets read in via multi-step operations

The trouble is, I think the `io` object itself needs to be at least set as an attribute on the class returned by `io.read()` to avoid garbage collection when leaving the the local scope of the read function, otherwise you get the behavior seen in https://github.com/NeurodataWithoutBorders/pynwb/issues/1711 where the file (HDF5, at least, IDK about Zarr) ends up closed and you can't access any dataset values

oruebel:
> The trouble is, I think the `io` object itself needs to be at least set as an attribute on the class returned by `io.read()` to avoid garbage collection

The issue is not just keeping the `io` object open so we can lazily load data on request, but we also need to make sure the `io` object is being closed correctly. This means the `read_nwb` methods will either need to return the `io` object to ensure the user can close the `io` or you need to "merge/link" the container with `io`. The current proposal in this PR always attaches the `io` object to the container. However, I don't think that is what you want here and it also does not currently solve the issue of closing the `io` object. I think you really would only want to attach the `io` object to the container (e.g., the NWBFile) if the container should actually own the `io` object and the `io` should always be closed and deleted at the same time the container is being deleted. I.e., instead of always setting the `io` object, I think you would: i) attach the `io` object to the NWBFile inside `read_nwb` instead of always attaching it in HDMFIO.read, and ii) the container would need to close and delete the `io` object when it is being deleted, e.g., in `__del__` and `__exit__`. If you would really want to always set the `io` object on the Container, then you would at the very least also need an additional flag on the Container to indicate whether the Container owns the `io` object so that it only closes the `io` when it actually owns it.

bendichter:
> we also need to make sure the `io` object is being closed correctly

Why? What if we open it in read mode? What problem does it cause if it doesn't close? Also, wouldn't making `io` an attribute of `NWBFile` make it _easier_ to properly close the `io` when you are finished with the file?

> the container would need to close and delete the io object when it is being deleted

Does closing happen automatically? I would imagine if an object is being deleted and the attribute of that object does not exist anywhere else in the namespace then it is also automatically deleted, but I could be wrong. But as it stands, all I am really advocating for is a change from

```python
io = nwbio(...)
nwbfile = io.read()
...
io.close()
```

to

```python
nwbfile = read_nwb(...)
...
nwbfile.io.close()
```

Are you worried that someone will `del nwbfile` and then not be able to call `nwbfile.io.close()`?

> I would attach the io object to the NWBFile inside read_nwb instead of always attaching it in HDMFIO.read

I think it would be confusing to have it so `NWBFile` objects that are produced by `read_nwb` have an attached `io` and `NWBFile` that are produced by `NWBIO` classes do not. I don't see the downside for attaching the `io` in all cases. Besides, there are other cases where having an `io` object linked is useful. What if I want to know what the backend of an `NWBFile` is, or somehow change my access pattern based on whether it is being streamed or read locally? Those parameters are in the `io` object, and I would find in more convenient of this were always available on the `NWBFile` object.

> If you would really want to always set the io object on the Container, then you would at the very least also need an additional flag on the Container to indicate whether the Container owns the io object so that it only closes the io when it actually owns it.

I don't really know what you mean by "own" here. Are you saying that in the `read_nwb` case the `NWBFile` object owns the `io` because there is no other `io` in the namespace, but in the `NWBIO` case the `NWBFile` object does not own the `io` object because one could call `io.read()` again and get another NWB file which you would not want to close?

oruebel:
@bendichter thanks for the additional thoughts and comments

> Does closing happen automatically? I would imagine if an object is being deleted and the attribute of that object does not exist anywhere else in the namespace then it is also automatically deleted

I think yes. This would rely on the automatic garbage collection to clean up the io, but I think that would work. I.e., if the io object is not being referenced anywhere else then calling `del nwbfile` should indeed also result in the `io` object being deleted and `HDF5IO` closes the file automatically when it is being deleted. I.e., I think the syntax indeed would be:

```
nwbfile = read_nwb(...)
...
del nwbfile
```

without having to call `nwbfile.io.close()` directly.


> I think it would be confusing to have it so `NWBFile` objects that are produced by `read_nwb` have an attached `io` and `NWBFile` that are produced by `NWBIO` classes do not.

If we rely on the Python garbage collector to close the io then I think always attaching the io object seems fine.

> I don't really know what you mean by "own" here.

With "own" I was referring to the question of who is responsible for closing the `io` object. When reading with `read_nwb` it is up to NWBFile to close the file whereas when a user uses NWBHDF5IO directly it is is up to the user. My concern then was, "How does NWBFile know if it should close the io?" However, if we rely on garabage collection, I think this should not be an issue. 

> But as it stands, all I am really advocating for is a change from

Long story short, overall, I think this approach is fine. One additional thought, I think with this we probably want to define `io` as an attribute on `AbstractContainer` so that we can set it on the container rather than using monkey patching to add it. The logic you propose would be the same, but in this way io is a documented property of the container that we can rely on, which will also make it easier to add logic around this in the future. For most containers the io would be `None`, i.e., only on the root Container that is being returned by `read` would the io be set, the same way you have it here. Since this PR currently just changes one line, why don't an take a quick crack at updating this PR in the next hour.

> What problem does it cause if it doesn't close?

For HDF5 closing files correctly is important (especially on Windows) because you can't open a file multiple times with colliding modes (e.g., open with 'r' and then 'w' at the same time won't work). For Zarr this is important when you want to use a different backend than the default directory store (e.g., one of Zarr's database backends). It mostly creeps up in tests or when you want to open a file again in the same program.

bendichter:
@oruebel Yes, I agree it would be best practice to initialize the `io` attribute in the class constructor

oruebel:
> Yes, I agree it would be best practice to initialize the `io` attribute in the class constructor

I can take a quick crack at updating the PR.

oruebel:
@bendichter I just pushed the changes for `AbstractContainer`. I still need to add tests.

oruebel:
> What problem does it cause if it doesn't close?

https://github.com/hdmf-dev/hdmf/actions/runs/5481164032/jobs/9985138446?pr=882#step:6:1548 this error on the new tests is an example of the kind of issues we see on Windows ;-)

oruebel:
I think I finally fixed the Windows issue. However, I would still recommend as a best practice (and in the docs) that we explicitly close the io object:

```
nwbfile = read_nwb(...)
...
nwbfile.read_io.close()
```

technically the following should work too, 


```
nwbfile = read_nwb(...)
...
del nwbfile
```

but it seems to be (at least on Windows) not reliable because it relies on the garbage collection to clean up the io.

oruebel:
@rly @bendichter this is ready for review

oruebel:
@bendichter I also added `AbstractContainer.get_read_io` method so you can get the io object used for reading from any child Container of the NWBFile as well. E.g.:

```
nwbfile = read_nwb(...)
ts = nwbfile.nwbfile.get_acquisition("vcs")
ts.read_io   # read_io will be None here. It is only set on the root nwbfile container
ts.get_read_io()  # this will return the read_io object set on the nwbfile
```

Alternatively, we could change the `read_io` property to behave like `get_read_io` but then we should add a `has_read_io` method, because otherwise we can't check of `read_io` is set on the Container or whether it is being retrieved from the parent.

bendichter:
Looks good! I prefer the way you have it, with get_* and set_* methods instead of attribute getters and setters when you are doing something nonstandard like this.

bendichter:
@oruebel I looked through the code and it looks great to me. However I am unable to formally approve it, since I created this PR

oruebel:
@rly can you please review and approve

rly (approved):
Looks good to me

PR #886, opened 2023-06-30
improve html output

bendichter:
@edeno 

current:

Summary expandable is rendered for fields that are empty. See the expanded groups at the bottom that are empty:

image

option 1: render them as empty

image

option 2: don't render them

image

I have a slight preference for  option 2, but I think option 1 would be fine and would be an improvement.

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -512,9 +512,10 @@ class Container(AbstractContainer):
         if isinstance(fields, dict):
             for key, value in fields.items():
                 current_access_code = f"{access_code}['{key}']"
+                if isinstance(value, (list, dict)) and not value:
+                    continue
                 if (
-                    isinstance(value, np.ndarray)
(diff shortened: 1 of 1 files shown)


bendichter:
Vote for Option 1

bendichter:
Vote for Option 2

edeno:
I don't personally mind the way it is because those exist, but are just empty. I guess there's the expectation that if something has the triangle that the expansion leads to an output. But hard for me to tell if that would freak people out or they would just understand that it could be expanded were there something there.

So Option 1 (or how it is) is fine, but I would not advocate for Option 2 (unless there's the situation in which there are a lot of these) because it lets the user know that something could be there but isn't.

bendichter:
Option 3: https://github.com/hdmf-dev/hdmf/pull/887 (could be combined with 1 or 2)

edeno:
Option 3 is the best I think.

rly:
Closed in favor of #887

PR #891, opened 2023-07-03
Warning for links

### LINKED ISSUE #885: [Feature]: Create a warning if linked objects are not already present in an NWBfile.

Create a warning if linked objects are not already present in an NWBfile.

_No response_

Create a warning if linked objects are not already present in an NWBfile.

Yes.

### Code of Conduct

mavaylon1:
## Motivation

Create a warning if linked objects are not already present in a parent when the original container's parent is set, stemming from the fact when objects that have links are added to containers (NWBFiles), the links are not added.

The Fix: Add a warning in the parent setter to search through the children of the original object for DynamicTableRegion and check the target table for a parent. If the parent is not set, raise a warning.

Original issue: https://github.com/NeurodataWithoutBorders/pynwb/issues/1701 

## How to test the behavior?
```
```
Run tests for container.py
## Checklist

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -331,7 +331,6 @@ class AbstractContainer(metaclass=ExtenderMeta):
 
     @parent.setter
     def parent(self, parent_container):
-        # breakpoint()
         if self.parent is parent_container:
             return
 
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -331,6 +331,7 @@ class AbstractContainer(metaclass=ExtenderMeta):
 
     @parent.setter
     def parent(self, parent_container):
+        # breakpoint()
         if self.parent is parent_container:
             return
 
@@ -354,6 +355,13 @@ class AbstractContainer(metaclass=ExtenderMeta):
             if isinstance(parent_container, Container):
                 parent_container.__children.append(self)
                 parent_container.set_modified()
+            for child in self.children:
+                if type(child).__name__ == "DynamicTableRegion":
+                    if child.table.parent is None:
+                        msg = "The table for this DynamicTableRegion has not been added to the parent."
+                        warn(msg)
+                    else:
+                        continue
 
     def _remove_child(self, child):
         """Remove a child Container. Intended for use in subclasses that allow dynamic addition of child Containers."""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### New features and minor improvements
 - Updated `ExternalResources` to have EntityKeyTable with updated tests/documentation and minor bug fix to ObjectKeyTable. @mavaylon1 [#872](https://github.com/hdmf-dev/hdmf/pull/872)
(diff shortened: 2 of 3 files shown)


mavaylon1:
Fix #885

src/hdmf/container.py
```diff
@@ -331,6 +331,7 @@ def parent(self):
 
     @parent.setter
     def parent(self, parent_container):
+        # breakpoint()
```
```suggestion
```

Remove breakpoint comment

oruebel:
@mavaylon1 in general for PR's it would be useful it you could fill in out the PR form to clarify the reasoning for the PR , checklist etc.

src/hdmf/container.py
```diff
@@ -354,6 +354,13 @@ def parent(self, parent_container):
             if isinstance(parent_container, Container):
                 parent_container.__children.append(self)
                 parent_container.set_modified()
+            for child in self.children:
+                if type(child).__name__ == "DynamicTableRegion":
+                    if child.table.parent is None:
+                        msg = "The table for this DynamicTableRegion has not been added to the parent."
+                        warn(msg)
+                    else:
+                        continue
```
oruebel, line 363: Could you clarify why this is specific for `DynamicTableRegion`?  I'm just wondering whether we should more generally check for: 1) any scalar attribute that is specified with a reference dtype in the schema (which would include the above case of DynamicTableRegion) and 2) any child that is specified as a Link in the schema. In principle we could also check for 3) any dataset that has a reference dtype, but that may be too expensive in terms of the number of objects we would need to check.

mavaylon1: In the issue ticket #885 I voiced/pinged the question of addressing how to tackle the issue of links needing a warning when the container parent is set. The original issue comes from pynwb ElectrodesSeries having a link the the ElectrodesTable. This is just a DynamicTableRegion. I looked more into hdmf and see that 
```
link_type = NamedTuple('DynamicTableLink',
--
1072 | [('source_table', DynamicTable),
1073 | ('source_column', Union[DynamicTableRegion, VectorIndex]),
1074 | ('target_table', DynamicTable)])

```
Seeing how the links are defined through DynamicTableRegion/VectorIndex and how VectorIndex is used in conjunction with DynamicTableRegion, I left the conditional to just check the former since the region will always be present (either stand alone or with the VectorIndex).

Anything beyond this I don't think falls within the original scope, but let's discuss further to see what else we can bring into the PR. @oruebel

oruebel: Thanks for the clarification. I'm fine with moving the discussion to a separate issue if that makes this PR too complex or would hold the PR back for too long from being merged. The PR as is I think is useful and addresses a key use case, so I think it is fine to merge and expand on it later. Up to you.

If you decide that we should leave the PR as is, then I think we should update the Changelog text to indicate that that this is for DynamicTableRegion. The current text refers to all links, which is really what led me to the question, because I was looking at the PR and not the original issue.

mavaylon1: I'll add the more general case to our discussions next week and let's merge this fix to tackle the issue.

oruebel: Sounds good.

oruebel (approved):

PR #895, opened 2023-07-07
ExternalResources I/O

mavaylon1:
## Motivation

This is to add logic to support ExternalResources on HDMFIO and HDF5IO. This was set up due to pynwb (https://github.com/NeurodataWithoutBorders/pynwb/actions/runs/5488375497?pr=1684). The update simply sets up external resources in the corresponding init methods.

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/backends/io.py
+++ b/src/hdmf/backends/io.py
@@ -12,9 +12,11 @@ class HDMFIO(metaclass=ABCMeta):
     @docval({'name': 'manager', 'type': BuildManager,
              'doc': 'the BuildManager to use for I/O', 'default': None},
             {"name": "source", "type": (str, Path),
-             "doc": "the source of container being built i.e. file path", 'default': None})
+             "doc": "the source of container being built i.e. file path", 'default': None},
+            {'name': 'external_resources', 'type': str,
+             'doc': 'The path to the ExternalResources', 'default': None},)
     def __init__(self, **kwargs):
-        manager, source = getargs('manager', 'source', kwargs)
+        manager, source, external_resources = getargs('manager', 'source', 'external_resources', kwargs)
         if isinstance(source, Path):
             source = source.resolve()
         elif (isinstance(source, str) and
@@ -26,6 +28,7 @@ class HDMFIO(metaclass=ABCMeta):
         self.__manager = manager
         self.__built = dict()
         self.__source = source
+        self.external_resources = external_resources
         self.open()
 
     @property
--- a/src/hdmf/backends/io.py
+++ b/src/hdmf/backends/io.py
@@ -53,6 +53,8 @@ class HDMFIO(metaclass=ABCMeta):
         if self.external_resources_path is not None:
             from hdmf.common import ExternalResources
             self.external_resources = ExternalResources.from_norm_tsv(path=self.external_resources_path)
+            if isinstance(container, ExternalResourcesManager):
+                container.link_resources(external_resources=self.external_resources)
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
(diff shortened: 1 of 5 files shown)


oruebel:
Overall, this looks good to me.  I'm wondering whether this should also include the reading of the ExternalResources file in `HDMFIO.read`:

```
if self.external_resources is not None:
            er_read = ExternalResources.from_norm_tsv(path=self.external_resources)
            file.link_resources(er_read)
```

mavaylon1:
> Overall, this looks good to me. I'm wondering whether this should also include the reading of the ExternalResources file in `HDMFIO.read`:
> 
> ```
> if self.external_resources is not None:
>             er_read = ExternalResources.from_norm_tsv(path=self.external_resources)
>             file.link_resources(er_read)
> ```

I did the local import, but without the link_resources since that exists only on ExternalResourcesManager.

mavaylon1:
I see there's a test missing. I'll add it.

mavaylon1:
Given that py3.7 has made it to the special farm upstate, can we ignore the conda 3.7 test issue? @rly 

We should also make the update to remove it on a new ticket.

mavaylon1:
@oruebel could you give this a gander so we can do a release and then merge the pynwb PR.

src/hdmf/backends/io.py
```diff
@@ -46,6 +50,9 @@ def read(self, **kwargs):
             # TODO also check that the keys are appropriate. print a better error message
             raise UnsupportedOperation('Cannot build data. There are no values.')
         container = self.__manager.construct(f_builder)
+        if self.external_resources_path is not None:
+            from hdmf.common import ExternalResources
+            self.external_resources = ExternalResources.from_norm_tsv(path=self.external_resources_path)
```
oruebel: Should the `ExternalResources` object not also be linked to the container if possible? Something along the lines of:

```suggestion
            self.external_resources = ExternalResources.from_norm_tsv(path=self.external_resources_path)
            if isinstance(container, ExternalResourcesManager):
                container.link_resources(external_resources=self.external_resources)
```

oruebel: Possibly also raise a warning if the `ExternalResources` was not linked to the Container.

src/hdmf/backends/io.py
```diff
@@ -46,6 +50,9 @@ def read(self, **kwargs):
             # TODO also check that the keys are appropriate. print a better error message
             raise UnsupportedOperation('Cannot build data. There are no values.')
         container = self.__manager.construct(f_builder)
+        if self.external_resources_path is not None:
```
oruebel, line 54: I think we should "safe-guard" reading the external resources with appropriate try/except blocks here so that a user can still get the container that is being read even if reading of the `ExternalResources` fails.

src/hdmf/backends/io.py
oruebel: Which approach are we using as default for saving the ExternalResources?

src/hdmf/backends/io.py
mavaylon1: raise warnings instead of errors as well?

oruebel: Yes, I think because this is in `HDMFIO.read` we'll need to raise warnings instead of errors here. The main reason is that we'd want the user to be able to read the Container even if reading the ExternalResources fails. Reading ExternalResources inside HDMFIO is in many ways a convenience and we would not want that functionality to break the core function of reading the main data file.

mavaylon1: I also saw your comment about moving ER before a release, I prefer if we did that after the dev days.

oruebel (approved):
Looks good to me

PR #906, opened 2023-07-18
Update bug_report.yml

bendichter:
Bugs do not always have tracebacks associated, so this field should be optional. When it is present it should be rendered with pytb, not shell


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -36,8 +36,6 @@ body:
 
         This section will be automatically formatted into code, so no need for backticks.
       render: shell
```
bendichter:
```suggestion
      render: pytb
```

oruebel (approved):

PR #910, opened 2023-07-19
Add ER write capability to IO backend

mavaylon1:
## Motivation

Give the ability to write ExternalResources if the path is provided and the container has a linked instance of ExternalResources.

## How to test the behavior?
```
```
Run the unit tests created.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### New features and minor improvements
 - Added the ability to write ExternalResources if the path is provided and the container has a linked instance of ExternalResources. @mavaylon1 [#910](https://github.com/hdmf-dev/hdmf/pull/910)
-  
+
 ### Bug fixes
 - Fixed bug on `add_ref_term_set` in which attributes that were not subscribtable returned an error. @mavaylon1 [#909](https://github.com/hdmf-dev/hdmf/pull/909)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # HDMF Changelog
 
-## HDMF 3.7.1 (Upcoming)
+## HDMF 3.8.0 (Upcoming)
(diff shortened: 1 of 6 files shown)


src/hdmf/container.py
```diff
@@ -43,7 +43,11 @@ def link_resources(self, **kwargs):
         self._external_resources = kwargs['external_resources']
 
     def get_linked_resources(self):
-        return self._external_resources if hasattr(self, "_external_resources") else None
+        if hasattr(self, "_external_resources"):
+            return self._external_resources
+        else:
+            msg = "Cannot find linked ExternalResources."
+            raise ValueError(msg)
```
oruebel: Since this is a get method, I think simply returning None is the more common approach and then a user would check ``if container.get_linked_resources() is None``

mavaylon1: I'll revert the change.

src/hdmf/backends/io.py
```diff
@@ -82,6 +82,14 @@ def write(self, **kwargs):
         f_builder = self.__manager.build(container, source=self.__source, root=True)
         self.write_builder(f_builder, **kwargs)
 
+        if self.external_resources_path is not None:
+            if container.get_linked_resources() is not None:
+                external_resources = container.get_linked_resources()
```
rly:
```suggestion
            external_resources = container.get_linked_resources()
            if external_resources is not None:
```
minor refactor to get one fewer call

oruebel (dismissed):

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## HDMF 3.7.1 (Upcoming)
```
oruebel: Since this adds a new feature, the version should be bumped to 3.8

oruebel (approved):

PR #917, opened 2023-07-25
Refactor DTR warning

rly:
## Motivation

Currently `AbstractContainer` performs a check for `type(child).__name__=='DynamicTableRegion'`. While this currently works, it is not ideal to have DTR-specific code in `AbstractContainer` and it only matches based on class name, not class type. This code is harder to find and does not work well for subclasses of DynamicTableRegion.

In addition, the check is not comprehensive in covering the possible situations in which it is useful. It only checks whether the referred-to table has a parent. The referred-to table may have a parent, but that parent may not have a parent. e.g. in NWB, a PlaneSegmentation may have an ImageSegmentation parent and that ImageSegmentation parent may have a ProcessingModule parent but that ProcessingModule may not have been added to the NWBFile.

See #891 and #885

## Checklist


mavaylon1:
@rly this slipped by. I'll review today.

src/hdmf/common/table.py
```diff
@@ -1421,6 +1421,26 @@ def __repr__(self):
                                                               id(self.table))
         return template
 
+    def _validate_on_set_parent(self):
+        # when this DynamicTableRegion is added to a parent, check:
+        # 1) if the table was read from a written file, no need to validate further
+        p = self.table
+        while p is not None:
+            if p.container_source is not None:
```
mavaylon1, line 1429: @rly what's the difference between parent and container source? Is the source meant to be the root?

mavaylon1: This checks if the table being used in the DTR has a source, if it does then pass, if not then it will check the parent for a source (and so on)

src/hdmf/common/table.py
```diff
@@ -1421,6 +1421,26 @@ def __repr__(self):
...
+                return super()._validate_on_set_parent()
+            p = p.parent
+
+        # 2) if none of the ancestors are ancestors of the linked-to table, then when this is written, the table
+        # field will point to a table that is not in the file
+        table_ancestor_ids = [id(x) for x in self.table.get_ancestors()]
+        self_ancestor_ids = [id(x) for x in self.get_ancestors()]
+
+        if set(table_ancestor_ids).isdisjoint(self_ancestor_ids):
```
mavaylon1, line 1438: This will check to see of the table and the DTR have some shared ancestor, aka check that its in the same file. If not then it'll be an external link and a warning will be thrown but it will be allowed to write. Correct?

src/hdmf/common/table.py
mavaylon1: Container_source ==> path to the file ==>Does zarrio set container_source (confirm?)

oruebel: container_source is set in `BuildManager.build` . ZarrIO deals with read/write of Builders, but the logic for creating Containters is the same as in HDMF, i.e., HDMF is setting the container_source and the logic to do so is independent of whether HDF5IO or ZarrIO is being used.

mavaylon1 (approved):

PR #923, opened 2023-07-28
DynamicEnums and Schemasheets integration 

mavaylon1:
## Motivation

Currently to create a TermSet a user has to manually add the permissible values to the enumeration. This is not scalable and so we want to take advantage of the LinkML tools that should make creating schemas easier. 
The goal of this PR is to add support for 

- [x] LinkML schemasheets in TermSet
- [x] LinkML integration for Dynamic TermSet 
- [x] Updated documentation

## How to test the behavior?
```
```
Run the unit tests.
## Checklist


mavaylon1:
@rly I know we don't like to go heavy handed on warnings, as they provide possible changes we should look into before it changes to an error. That being said, importing oaklib brings almost 10 deprecation warnings for pkg_resources. I talked to Sujay and he confirmed it was on their side. I will talk with him some more in regards to any plans on their end for updates. Till then I ignored warnings on importing oaklib.

mavaylon1:
This was ready for review a week ago (hence the green check) but it seems after the multiple merges to dev things are now outdated. I will fix it this weekend. @rly @oruebel 

UPDATE: linkml-runtime updated to 1.5.6 which breaks some stuff on our end. I capped everything at 1.5.5 for now and will talk to Sujay.

docs/gallery/plot_term_set.py
```diff
@@ -41,25 +41,59 @@
...
 
 For a clear example, please view the
 `example_term_set.yaml <https://github.com/hdmf-dev/hdmf/blob/dev/docs/gallery/example_term_set.yaml>`_
 for this tutorial, which provides a concise example of how a term set schema looks.
+
+For more information on how to properly format the Google spreadsheet to be compatible with LinkMl, please
+refer to https://linkml.io/schemasheets/#examples.
+
+For more information how to properly format the schema to support LinkML Dynamic Enumerations, please
+refer to https://linkml.io/linkml/schemas/enums.html#dynamic-enums.
```
oruebel:
```suggestion

.. note::
    For more information on how to properly format the Google spreadsheet to be compatible with LinkMl, please
    refer to https://linkml.io/schemasheets/#examples.

.. note::
    For more information how to properly format the schema to support LinkML Dynamic Enumerations, please
    refer to https://linkml.io/linkml/schemas/enums.html#dynamic-enums.
```

src/hdmf/term_set.py
```diff
@@ -12,19 +15,37 @@ class TermSet():
...
+                self.term_schema_path = self.schemasheets_convert()
+                self.view = SchemaView(self.term_schema_path)
+        else:
+            self.view = SchemaView(self.term_schema_path)
+
+        if dynamic:
+            # reset view to now include the dynamically populated term_set
+            self.expanded_term_set_path = self.enum_expander()
+            self.view = SchemaView(self.expanded_term_set_path)
+
```
oruebel:
```suggestion
        self.expanded_term_set_path = None
        if dynamic:
            # reset view to now include the dynamically populated term_set
            self.expanded_term_set_path = self.enum_expander()
            self.view = SchemaView(self.expanded_term_set_path)

```

Having the existence of public  object members depend on input parameters is problematic, because it requires users to check if an attribute exists. Instead, we should set `self.expanded_term_set_path` to `None` if it is not being used. If this is strictly and internal variable, then it may also be useful to make the variable private or protected by adding the `_` or `___` prefix.

src/hdmf/term_set.py
```diff
@@ -12,19 +15,37 @@ class TermSet():
     :ivar view: SchemaView of the term set schema
```
oruebel, line 15: Please add `:ivar ...:` docs to the class docstring for new instance variables ``self.schemasheets_folder`` and ``self.expanded_term_set_path``

src/hdmf/term_set.py
```diff
@@ -94,3 +115,39 @@ def __getitem__(self, term):
         except KeyError:
             msg = 'Term not in schema'
             raise ValueError(msg)
+
+    def schemasheets_convert(self):
```
oruebel: Please add a docstring for `` schemasheets_convert`` to indicate what the function does

src/hdmf/term_set.py
```diff
@@ -94,3 +115,39 @@ def __getitem__(self, term):
...
+        schema = schema_maker.create_schema(tsv_file_paths)
+        schema_dict = schema_as_dict(schema)
+        schemasheet_schema_path = os.path.join(self.schemasheets_folder, f"{schema_dict['name']}.yaml")
+
+        with open(schemasheet_schema_path, "w") as f:
+            yaml.dump(schema_dict, f)
+
+        return schemasheet_schema_path
+
+    def enum_expander(self):
```
oruebel: Pleas add a docstring for ``enum_expander`` to indicate what the function does.

oruebel: If I understand the function correctly, then this is creating a new file with the expanded version of the dynamic enumeration and places it next to the original term set. If so then maybe calling the function ``expand_term_set`` would be more intuitive.

src/hdmf/term_set.py
```diff
@@ -94,3 +115,39 @@ def __getitem__(self, term):
...
+            from oaklib.utilities.subsets.value_set_expander import ValueSetExpander
+        except ImportError:
+            msg = 'Install oaklib.'
+            raise ValueError(msg)
+        expander = ValueSetExpander()
+        # TODO: linkml should raise a warning if the schema does not have dynamic enums
+        enum = list(self.view.all_enums())
+        schema_dir = os.path.dirname(self.term_schema_path)
+        file_name = os.path.basename(self.term_schema_path)
+        output_path = os.path.join(schema_dir, f"expanded_{file_name}")
```
oruebel, line 167: Can you clarify why the term set is being expanded in the same place and why a user should not provide a path? If either option is fine, then I would add `output_path = None`  as an input argument and then set the output_path to be in place if the user does not provide a path.

src/hdmf/term_set.py
```diff
@@ -94,3 +115,39 @@ def __getitem__(self, term):
...
+        expander.expand_in_place(self.term_schema_path, enum, output_path)
+
+        return output_path
```
oruebel, line 170: Is the expanded term set being used directly by this instance of TermSet or would the user create a new instance of TermSet to open the expanded term set? If the later, then I'm wondering whether it would be more useful to directly return a new TermSet:
```suggestion
        return  TermSet(term_schema_path=output_path)
```

docs/gallery/plot_term_set.py
```diff
@@ -41,25 +41,61 @@
...
+# refer to the Schemasheets tutorial link above for more details on the required syntax structure within the sheets.
+# Once the tsv files are in a folder, the user simply provides the path to the folder with ``schemasheets_folder``.
+termset = TermSet(schemasheets_folder=schemasheets_folder)
+
+# Use Dynamic Enumerations to populate TermSet
+# --------------------------------------------
+# The :py:class:`~hdmf.term_set.TermSet` class allows user to skip manually defining permissible values, by pulling from
+# a LinkML supported source. These sources contain multiple ontologies. A user can select a node from an ontology,
+# in which all the elements on the branch, starting from the chosen node, will be used as permissible values.
+# Please refer to the LinkMl Dynamic Enumeration tutorial for more information on these sources and how to setup Dynamic
```
oruebel, line 99: Add link for the text "LinkMl Dynamic Enumeration tutorial"

src/hdmf/term_set.py
oruebel: Looking at the tutorial it seems that users are not supposed to call this function directly, but are instead supposed to construct ``TermSet(...., dynamic=True)``. If that is the case, then I would add  the `__` prefix to the function name to make it private.

src/hdmf/term_set.py
```diff
@@ -94,3 +115,39 @@ def __getitem__(self, term):
...
             raise ValueError(msg)
+
+    def schemasheets_convert(self):
+        try:
+            import yaml
+            from linkml_runtime.utils.schema_as_dict import schema_as_dict
+            from schemasheets.schemamaker import SchemaMaker
+        except ImportError:
+            msg="Install schemasheets."
+            raise ValueError(msg)
```
oruebel:
```suggestion
        except ImportError:   # pragma: no cover
            msg="Install schemasheets."   # pragma: no cover
            raise ValueError(msg)   # pragma: no cover
```

src/hdmf/term_set.py
```diff
@@ -94,3 +115,39 @@ def __getitem__(self, term):
...
+
+        return schemasheet_schema_path
+
+    def enum_expander(self):
+        try:
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            from oaklib.utilities.subsets.value_set_expander import ValueSetExpander
+        except ImportError:
+            msg = 'Install oaklib.'
+            raise ValueError(msg)
```
oruebel:
```suggestion
        except ImportError:   # pragma: no cover
            msg = 'Install oaklib.'  # pragma: no cover
            raise ValueError(msg)  # pragma: no cover
```

docs/gallery/plot_term_set.py
```diff
@@ -41,25 +41,61 @@
...
 except NameError:
     dir_path = os.path.dirname(os.path.abspath('.'))
     yaml_file = os.path.join(dir_path, 'gallery/example_term_set.yaml')
+    schemasheets_folder = os.path.join(dir_path, 'gallery/schemasheets')
+    dynamic_schema_path = os.path.join(dir_path, 'gallery/example_dynamic_term_set.yaml')
+
+# Use Schemasheets to create TermSet schema
+# -----------------------------------------
+# The :py:class:`~hdmf.term_set.TermSet` class builds off of LinkML Schemasheets, allowing users to convert between
+# a Google spreadsheet to a complete LinkML schema. Once the user has defined the necessary LinkML metadata within the
```
oruebel, line 88: Can we add screenshot of an example Google spreadsheet here to illustrate how this looks? Or maybe even a small graphic that show the spreadsheet and the corresponding TermSet that LinkML creates?

oruebel (commented):
Overall the PR looks good. I added a few comments below, but those should all be easy/quick to fix.

src/hdmf/term_set.py
mavaylon1: The output of the method is a path that will be used to be create a SchemaView and not a Termset. I will clarify this in the docstring.

src/hdmf/term_set.py
mavaylon1: This is an internal method. In the spirit of minimizing fields the user has to specify, the new schema will be saved in the directory of the original schema (which is why I don't allow an output path).

docs/gallery/plot_term_set.py
mavaylon1: With how LinkML uses multiple tabs in the sheets, I think the link I added to look at schemasheets for users to have a look at an example sheets will be more helpful and give a better picture.

src/hdmf/term_set.py
mavaylon1: Will do.

mavaylon1:
> Overall the PR looks good. I added a few comments below, but those should all be easy/quick to fix.

I really appreciate it! I made most of the changes, but also I left some notes to clarify some the methods both here and in the docstring as to why it is setup the way it is.

src/hdmf/term_set.py
```diff
@@ -1,30 +1,56 @@
...
     """
     Class for implementing term sets from ontologies and other resources used to define the
     meaning and/or identify of terms.
 
-    :ivar term_schema_path: The LinkML YAML enumeration schema
+    :ivar term_schema_path: The path to LinkML YAML enumeration schema
     :ivar sources: The prefixes for the ontologies used in the TermSet
     :ivar view: SchemaView of the term set schema
+    :ivar schemasheets_folder: The path to the folder containing the LinkML TSV files
+    :ivar self.expanded_term_set_path: The path to the schema with the expanded enumerations
```
oruebel:
```suggestion
    :ivar expanded_term_set_path: The path to the schema with the expanded enumerations
```

oruebel (approved):

PR #924, opened 2023-07-30
Add base methods for iterator serialization

CodyCBakerPhD:
## Motivation

Zarr supports efficient parallelization, and a sister PR (https://github.com/hdmf-dev/hdmf-zarr/pull/111) adds this capability but requires the argument `number_of_jobs` to be propagated from a top level of the `HDMFIO.write` due to the super call at [this line](https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/backend.py#L203)

It also has a subtle condition in that our iterators (which in this role primarily define how to access data from a source) must be pickleable, so I've defined methods related to that for the `GenericDataChunkIterator` (which for several reasons should be the only type of iterator allowed for usage in parallelization)

An example in practice can be found downstream: https://github.com/catalystneuro/neuroconv/pull/536

## Checklist

- [X] Did you update CHANGELOG.md with your changes?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,6 @@ dependencies = [
     "ruamel.yaml>=0.16",
     "scipy>=1.4",
     "importlib-resources; python_version < '3.9'",  # TODO: remove when minimum python version is 3.9
-    "threadpoolct"
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ dependencies = [
     "ruamel.yaml>=0.16",
     "scipy>=1.4",
     "importlib-resources; python_version < '3.9'",  # TODO: remove when minimum python version is 3.9
+    "threadpoolct"
 ]
 dynamic = ["version"]
 
--- a/src/hdmf/backends/io.py
+++ b/src/hdmf/backends/io.py
@@ -75,6 +75,10 @@ class HDMFIO(metaclass=ABCMeta):
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
+            {'name': 'number_of_jobs', 'type': int, 'doc': 'Number of jobs to use in parallel during write (only operates on GenericDataChunkIterator-wrapped datasets).', 'default': 1},
+            {'name': 'number_of_jobs', 'type': int,
+         'doc': "Number of jobs to use in parallel during write (only operates on GenericDataChunkIterator-wrapped datasets).",
+         'default': 1},
             allow_extra=True)
     def write(self, **kwargs):
         """Write a container to the IO source."""
--- a/src/hdmf/backends/io.py
+++ b/src/hdmf/backends/io.py
@@ -75,10 +75,15 @@ class HDMFIO(metaclass=ABCMeta):
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
-            {'name': 'number_of_jobs', 'type': int, 'doc': 'Number of jobs to use in parallel during write (only operates on GenericDataChunkIterator-wrapped datasets).', 'default': 1},
-            {'name': 'number_of_jobs', 'type': int,
-         'doc': "Number of jobs to use in parallel during write (only operates on GenericDataChunkIterator-wrapped datasets).",
-         'default': 1},
+            {
+                "name": "number_of_jobs",
+                "type": int,
+                "doc": (
+                    "Number of jobs to use in parallel during write "
+                    "(only works with GenericDataChunkIterator-wrapped datasets)."
+                ),
+                "default": 1,
+            },
             allow_extra=True)
     def write(self, **kwargs):
         """Write a container to the IO source."""
--- a/src/hdmf/backends/io.py
+++ b/src/hdmf/backends/io.py
@@ -74,17 +74,7 @@ class HDMFIO(metaclass=ABCMeta):
 
         return container
 
-    @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
-            {
-                "name": "number_of_jobs",
-                "type": int,
-                "doc": (
-                    "Number of jobs to use in parallel during write "
-                    "(only works with GenericDataChunkIterator-wrapped datasets)."
-                ),
-                "default": 1,
-            },
-            allow_extra=True)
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "ruamel.yaml>=0.16",
     "scipy>=1.4",
     "importlib-resources; python_version < '3.9'",  # TODO: remove when minimum python version is 3.9
+    "threadpoolct"
```
CodyCBakerPhD: Do you like to have lower bound versions pinned here?

src/hdmf/common/hdmf-common-schema
CodyCBakerPhD, line 1: Unsure why this file got changed; maybe I just need to update it from the submodule

src/hdmf/data_utils.py
```diff
@@ -241,7 +242,7 @@ def __init__(self, **kwargs):
                 for buffer_axis, maxshape_axis in zip(self.buffer_shape, self.maxshape)
             ],
         )
-        self.buffer_selection_generator = (
+        self.buffer_selections = tuple(  # temporary solution to cast as tuple to allow indexing
```
CodyCBakerPhD: This may be able to be reverted back to the previous state depending on how the actual selection iteration is performed on the Zarr side (TBD)

pyproject.toml
oruebel: @rly thoughts?

src/hdmf/backends/io.py
```diff
@@ -75,6 +75,10 @@ def read(self, **kwargs):
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
+            {'name': 'number_of_jobs', 'type': int, 'doc': 'Number of jobs to use in parallel during write (only operates on GenericDataChunkIterator-wrapped datasets).', 'default': 1},
+            {'name': 'number_of_jobs', 'type': int,
+         'doc': "Number of jobs to use in parallel during write (only operates on GenericDataChunkIterator-wrapped datasets).",
```
oruebel: It looks like the `number_of_jobs` argument is duplicated here. I assume one of these should be removed.

oruebel: Why is this argument needed on HDMFIO? HDMFIO should only expose arguments that we expect all backends to support. If this argument is specific for ZarrIO then we should probably only add it in hdmf-zarr

pyproject.toml
rly: It would be nice to set a lower bound version here unless all versions work fine. You can test it in requirements-min.txt which is used to make sure tests pass with the lower bound version.

rly: Also I think this should be `threadpoolctl`

CodyCBakerPhD: Moved to https://github.com/hdmf-dev/hdmf-zarr/pull/111

src/hdmf/common/hdmf-common-schema
CodyCBakerPhD: Fixed now, guess it was a problem with submodule on fork not syncing

src/hdmf/backends/io.py
CodyCBakerPhD: From the lines

```python
with NWBZarrIO(path=..., mode="w") as io:
    io.write(nwbfile, number_of_jobs=number_of_jobs)
```

leads to

https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/nwb.py#L14

which since it does not redefine `write` leads to

https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/backend.py#L200

which then quickly directs to the `super` (which is `HDMFIO`) at

https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/backend.py#L203

I suppose one way around this could be to set it as an instance attribute on the HDMF-Zarr side; but that feels weird since this is not a property of the `NWBZarrIO` class or object, but is a control parameter over how we want the operation `write` to be performed

src/hdmf/backends/io.py
```diff
@@ -75,6 +75,15 @@ def read(self, **kwargs):
         return container
 
     @docval({'name': 'container', 'type': Container, 'doc': 'the Container object to write'},
+            {
```
rly: @CodyCBakerPhD can you move this to hdmf-zarr?

With `docval(..., allow_extra=True)` and `**kwargs` being passed around, it will make its way to `write_builder`

https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/backend.py#L200

rly: Since this arg is specific to Zarr at the moment, I think it makes the most sense to move it to hdmf_zarr

CodyCBakerPhD: Ahhh, because of `allow_extra=True`, I see

src/hdmf/backends/io.py
CodyCBakerPhD: Or I suppose `allow_extra=True` is what allows it to not be specified at the HDMF level, but nonetheless passed implicitly throughout the nested calls

src/hdmf/data_utils.py
CodyCBakerPhD: Yep, no longer needed

CodyCBakerPhD:
@rly @oruebel Ready for review

src/hdmf/data_utils.py
```diff
@@ -372,6 +365,11 @@ def __next__(self):
                 self.progress_bar.write("\n")  # Allows text to be written to new lines after completion
             raise StopIteration
 
+    def __reduce__(self) -> Tuple[Callable, Iterable]:
+        instance_constructor = self._from_dict
+        initialization_args = (self._to_dict(),)
+        return (instance_constructor, initialization_args)
```
rly, line 370: I know this might seem silly, but could you add a test where you create a dummy GenericDCI class that defines `_from_dict` and `_to_dict` and test that this `__reduce__` method does what it should?

CodyCBakerPhD: Done

rly (approved):

PR #934, opened 2023-08-07
set_data_io method and associated test

bendichter:
## Motivation

A potential replacement for https://github.com/NeurodataWithoutBorders/pynwb/pull/1752

## How to test the behavior?
```
```

## Checklist


rly:
I think this is fine. The only minor concern is that this will not raise an error or do anything if a user calls `set_data_io` on a non-scalar attribute. But that is rare and has no bad side effect. There is no straightforward way to detect at the container level whether a field is a dataset or an attribute. I think this is good then. Please add a changelog. Thanks!

bendichter:
@rly I just noticed there is a (now defuct) method on `Data` called `set_dataio` (no second underscore). Do you think it would be better to rename this to be consistent?

rly:
I kind of prefer `set_data_io` because the underscores separate words. `wrap_data_io` might be better, but either is fine.

rly (approved):

Issue #927, opened 2023-08-01
[Feature]: TermSet Restructure (Stage 1): add from root_container 


mavaylon1:
### What would you like to see added to HDMF?

This will be part 1 of restructuring TermSet to be more general by supporting validation for all "python attributes" of HDMF classes.

`add_ref_termset` will shift to take only a container object, in our case this will be a nwbfile but in HDMF this will be any container, and will find all children that have a TermSet (check for Data as instance) in order to bulk add to ExternalResources.

pseudo-code:
```
for containers in nwbfile.objects
... for attributes in containers
......<check>
```

Eventually this set up will be altered to not check for TermSet, but whether there is a `TermSetWrapper` class in place. This is defined in Part 2.

### What solution would you like?

Refer above

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/hdmf-dev/hdmf/issues)?

mavaylon1:
Note: Think about how termsets are used with compound data

PR #935, opened 2023-08-08
TermSet Update #1

mavaylon1:
## Motivation

This is the first stage in the TermSet updates. Update `add_ref_term_set` so that it takes in only a root container, in PyNWB it will be a NWBFile, and search for objects that have a `TermSet`. Currently, it is only data objects that can have termsets.

TODO Items:

- [x] Update add_ref_term_set
- [x] Updates test_resources.py
- [x] Update container.py to support `all_children` search
- [x] Update test_container.py to test `all_children`
- [x] Removed `add_ref_termset_ from the gallery for HERD to be updated once TermSet is more finalized with the wrapper.

Changes to Note:
I removed the validation step in `add_ref_termset` (and corresponding tests). This is due to the fact that when a user adds data to `HERD` the data is validated prior using the termset and so we do not need a double validation.

## How to test the behavior?
```
```
Run unit tests.
## Checklist


mavaylon1:
Fix #927

mavaylon1:
@oruebel Proof of concept is done but some questions are floating in my head.
1. I need to think about how this all works with compound data and TermSets
2. Looking at the plot_external_resources.py I have to set the parent but I wonder if that's not the most user friendly line to have there. I technically could just use the table as a root_container. Thoughts? (I do need to re-word that section)

Update:
1. We discussed and made a ticket #938 
2. I removed it per consulting with Ryan

src/hdmf/container.py
```diff
@@ -302,6 +303,37 @@ def get_ancestor(self, **kwargs):
...
+            else: # pragma: no cover
+                # warn that a child does not have an object_id, which is unusual
+                warn('%s "%s" does not have an object_id' % (type(n).__class__, n.name))
+            if hasattr(n, 'children'):
+                for c in n.children:
+                    stack.append(c)
+        return ret
+
+    @property
+    def all_objects(self):
```
rly, line 331:
```suggestion
    def all_objects(self):
        """Get a LabelledDict that indexed all child objects recursively by object ID."""
```

src/hdmf/container.py
```diff
@@ -302,6 +303,38 @@ def get_ancestor(self, **kwargs):
             p = p.parent
         return None
 
+    def all_children(self):
+        """
+        This method will search through all children within the object and their children.
```
rly:
```suggestion
        """Get a list of all child objects and their child objects recursively.
```

rly: In general, the first line of a method/function docstring should be a succinct description of what the method/function does / what it returns

rly (approved):

Issue #928, opened 2023-08-01
[Feature]: TermSet Restructure (Stage 2 & 3): Wrapper class and "write" compatibility 


mavaylon1:
### What would you like to see added to HDMF?
Stage 1: #927 
Stage 2:
This will introduce the `TermSetWrapper` class. The point of this class is act similarly to DataIO in which it can be wrapped around objects to provide a TermSet to HDF5 groups, datastets, and attributes. 

This wrapper needs to be integrated such that it does not disrupt existing mechanisms. For example, if I wrap a VectorData instance and I want to access the data, then usually I can do `instance.data`. However, if I simply wrap the instance then I would have to do `wrap.data.data`, breaking existing code and workflows. 

We also have to be aware of `docval`. We would most likely need to expose the `dtype` as a property

Stage 3:
With the wrapper class, adjustments may need to be made such that all containers for PyNWB pass this wrapper update along with adding the unwrapping functionality so that the wrapper does not interfere with write. I would expect that the main issues here are likely: 1) passing docval (which should be a simple fix in HDMF), and 2) passing the various validation and data processing that the PyNWB classes want to do, e.g., when we check in __init__ that data values are correct or aligned with other fields.

This may lead to adjustments to the BuildManager. The idea here is that the I/O backend does not need to know about TermSet's because this is only relevant for HERD files, which are written separately. So when creating builders for I/O we could simply "unwrap" the data so that nothing changes for the I/O backend.

Alternatively, the HDF5IO and ZarrIO backends would need to be updated to handle the case of wrapped datasets correctly.
```
@property
def dtype(self):
...return self.data.dtype
```
We may also need to make changes to docval, but this will come with experimenting. 

Much like `DataIO`, we will need to support all forms of getting and setting, such that nothing breaks and is wrapped only to provide a TermSet.

We would also remove TermSet from Data at this stage.

- [ ] Update docs for ExternalResources that points to pynwb for real world example

### What solution would you like?

Refer above

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/hdmf-dev/hdmf/issues)?

PR #950, opened 2023-08-29
TermSetWrapper and write support

mavaylon1:
## Motivation

This will introduce the `TermSetWrapper` class. The point of this class is act similarly to DataIO in which it can be wrapped around objects to provide a TermSet to HDF5 datasets, and attributes.

The wrapper will hold the termset for validation for any dataset or attribute. Some concerns:
1. This will require major changes in NWB so that docval still passes
2. We need to be aware of `write`. The termset wrapper is not supposed to be written and so needs to be unwrapped. 
3. A lot of things will break and edge cases will come up with development. Adjust accordingly.

TODO:

- [x] Remove prior termset integration from Data.
- [x] Update old tests that use TermSet: container.py, table.py
- [x] Add tests for TermSetWrapper append/extend
- [x] Add tests for TermSetWrapper attributes and data (this is in test_resources)
- [x] Update tests for resources.py with the TermSetWrapper and new helper functions
- [x] Add tests for roundtrip/write/write with HERD (Partially Done)
- [x] Add tests for write_dataset
- [x] Add parameters on write to: This required changes to `HDMFIO` (in order to be later ported to HDMF_Zarr), changes to how `HERD` interacts with `TermSet` due to the wrapper, updates on `add_ref_term_set`, and assorted helper functions
- [x] Consider how to validate when users append new data to VectorData and DynamicTable: Currently, Termsets are integrated as a field on Data. We want to remove this and now only use the wrapper, such that we wrap the data within the VectorData. This requires us to have append and extend in the `TermSetWrapper`. This will also require changes to `add_column` and `add_row` in `DynamicTable` and `add_row` in `VectorData` (all require test changes).
- [x] Finalize Validation 
- [x] Write compatible
- [x] Docval compatible
- [x] Tutorial
- [x] Explore how the wrapper would work on a dataset that also wants to use DataIO (Created ticket: https://github.com/hdmf-dev/hdmf/issues/954)

## How to test the behavior?
```
```

## Checklist


mavaylon1:
Fix #928

src/hdmf/utils.py
```diff
@@ -273,6 +274,12 @@ def __parse_args(validator, args, kwargs, enforce_type=True, enforce_shape=True,
                 type_errors.append("missing argument '%s'" % argname)
             else:
                 if enforce_type:
+                    from .container import TermSetWrapper # circular import fix
+                    termset = False
+                    if isinstance(argval, TermSetWrapper):
+                        # kwargs is the dict that stores the object names and the values
+                        # we can use this to unwrap the dataset/attribute to use the "item" for docval to validate the type.
+                        argval = kwargs[argname].item
```
oruebel: Should this be:

```suggestion
                        argval = argval.item
```

I think in: https://github.com/hdmf-dev/hdmf/blob/561e2797e5b488e90640ccc08d978e5c6b8b89b5/src/hdmf/utils.py#L260-L271

``__parse_args`` already extracts the value of the argument and looking up the argument and assigns it to ``argval``. I think looking up the argument via ``kwargs[argname].item`` would fail for positional arguments.

oruebel: I think this check to "unwrap" `argval` may need to go before: 

https://github.com/hdmf-dev/hdmf/blob/561e2797e5b488e90640ccc08d978e5c6b8b89b5/src/hdmf/utils.py#L276

since the other if conditions to check `shape` and `enum` that appear afterwards also need the unwrapped argument value, e.g., here:

https://github.com/hdmf-dev/hdmf/blob/561e2797e5b488e90640ccc08d978e5c6b8b89b5/src/hdmf/utils.py#L290-L291

mavaylon1: @oruebel it remains unwrapped throughout

mavaylon1: The order of operations when the code runs is 
1. The wrapper validates the item
2. Then the docval is called
That way we actually just need to unwrap the item once.

oruebel: > @oruebel it remains unwrapped throughout

Yes, but what happens if `enforce_type` is `False` and `enforce_shape` is `True`. In that case the code to unwrap would not be called and `argval` should still be wrapped.

mavaylon1: Ahh I see. In practice, why would we ever not enforce type?

oruebel: > Ahh I see. In practice, why would we ever not enforce type?

I'm not sure if/were this might happen right now  But I think all that is needed is to move the few lines of code that do the unwrapping before the `if enforce_type` statement so that the unwrapping is always done before any of the checks.

mavaylon1: Makes sense and will do. Sorry for that miss.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1099,6 +1100,8 @@ def write_dataset(self, **kwargs):  # noqa: C901
             dataio = data
             link_data = data.link_data
             data = data.data
+        if isinstance(data, TermSetWrapper):
+            data = data.item
```
oruebel: I think you may also need to do the same in ``set_attributes`` https://github.com/hdmf-dev/hdmf/blob/a0fdb244727b622de17dc470eec9bc5d1252a510/src/hdmf/backends/hdf5/h5tools.py#L929-L931

at the beginning of the loop do ``isinstance(data, TermSetWrapper):  value = value.item``

src/hdmf/build/objectmapper.py
```diff
@@ -564,6 +565,8 @@ def get_attr_value(self, **kwargs):
                 msg = ("%s '%s' does not have attribute '%s' for mapping to spec: %s"
                        % (container.__class__.__name__, container.name, attr_name, spec))
                 raise ContainerConfigurationError(msg)
+            if isinstance(attr_val, TermSetWrapper):
+                    attr_val = attr_val.item
```
oruebel: If the data is being unwrapped here in the ObjectMapper, are the changes in`H5DataIO` actually needed?

mavaylon1: I don't believe I've made any changes to `H5DataIO` @oruebel

mavaylon1: Here was my approach to handling the write compatibility:
1. Datasets 
Write -> write_builder -> write_group and write_dataset (write_group also calls write_dataset) In write_dataset:
```
if isinstance(data, H5DataIO):
            options['io_settings'] = data.io_settings
            dataio = data
            link_data = data.link_data
            data = data.data
        if isinstance(data, TermSetWrapper):
            data = data.item
```
With DataIO we flag it after the data after the builder has been made and unwrap it. We do the same for TermSet.

2. Attributes
Attributes had to go a bit differently. If we followed the workflow of DataIo and the wrapper for the Termset, we would have to modify the attributes after the builder has been made. However, when you look at the builder the attribute has been converted to a string. So instead of <hdmf.container.TermSetWrapper object at 0x11fde7850> it was giving ‘<hdmf.container.TermSetWrapper object at 0x11fde7850>’. 

The fix I came up with was to tackle it before the builder was made via the object mapper. As you see in the lines you highlighted.

mavaylon1: Now could we change how we handle data such that it is also flagged in the object_mapper. I suppose we could. I just followed how DataIO was handled and followed suit. @oruebel

mavaylon1:
Order of Operations for appending data:

1. VectorData
add_row ->self.append -> append in `Data`. 
We want to call the self.append in the wrapper if the wrapper is present. 

2. DynamicTable `add_row`
We validate before going through the process of adding the data to the rows. This is a bulk check. Once that passes, the column add_row is called, which should work with the wrapper with the changes in VectorData. This does do another validation (not ideal). This can be avoided by adding a field that dictates the data already being validated, but it doesn't really matter: the user doesn't ever interact with this level of validation and it does not add that much time to have a second validation.

4. DynamicTable `add_column`
Can revert back to pre-termset integration.

mavaylon1:
Tests that need to be fixed:

- [ ] test_container.py
- [ ] test_io_hdf5_h5tools.py
- [ ] unit/test_table.py
- [ ] test_term_set.py
- [ ] test_io_map_data.py
- [ ] test_build.py
- [ ] test_build_quantity.py
- [ ] test_alignedtable.py
- [ ] test_generate_table.py
- [ ] test_linkedtables.py
- [ ] test_multi.py
- [ ] test_resources.py
- [ ] unit/common/test_table.py

mavaylon1:
The use of "fields" is not consistent, which makes searching for the wrapper more difficult.

mavaylon1:
The meaning of "attribute" is starting to take on multiple meanings. When we are sloppy we say attribute and field, but that is not always interchangeable. Attribute in HDF5 is not always a field. For example, in data we have a description attribute but the data itself is not, even though you can think of it as a python field. Now that you can wrap hdf5 dataset and hdf5 attributes.

src/hdmf/build/objectmapper.py
```diff
@@ -564,7 +565,11 @@ def get_attr_value(self, **kwargs):
                 msg = ("%s '%s' does not have attribute '%s' for mapping to spec: %s"
                        % (container.__class__.__name__, container.name, attr_name, spec))
                 raise ContainerConfigurationError(msg)
-            if attr_val is not None:
+            if isinstance(attr_val, TermSetWrapper):
+                # This is when the wrapped item is an attribute
+                # Refer to h5tools.py for wrapped datasets
+                attr_val = attr_val.value
+            elif attr_val is not None:
```
rly:
```suggestion
            if attr_val is not None:
```
Why `elif` here? I think it should be `if` so that the attribute value is processed normally after unwrapping

mavaylon1: It's elif because otherwise the value, in the case of data for Foo, will take a list and convert it to a string.

src/hdmf/common/resources.py
```diff
@@ -408,7 +409,29 @@ def _get_file_from_container(self, **kwargs):
...
+        Takes a list of objects and checks the fields for TermSetWrapper.
+        :return: [[object1, attribute_name1, wrapper1], [object2, attribute_name2, wrapper2], ...]
+        """
+        objects = kwargs['objects']
+
+        ret = [] # list to be returned with the objects, attributes and corresponding termsets
+
+        for obj in objects:
+            # Get all the fields, parse out the methods and internal variables
+            obj_fields = [a for a in dir(obj) if not a.startswith('_') and not callable(getattr(obj, a))]
```
rly, line 428: This coarse approach still unsettles me, but I cannot think of a specific real-world example of a problem that this might cause.

mavaylon1: I can open an enhancement ticket for us to look at down the road?

src/hdmf/common/table.py
```diff
@@ -596,23 +585,9 @@ def add_row(self, **kwargs):
...
-            else:
-                if col.term_set is not None:
-                    if col.term_set.validate(term=data[colname]):
-                        continue
-                    else:
-                        bad_data.append(data[colname])
-
-        if len(bad_data)!=0:
-            msg = ('"%s" is not in the term set.' % ', '.join([str(item) for item in bad_data]))
-            raise ValueError(msg)
```
rly: What is nice about having this validation code here is that all the values of the row are checked before any values in the row are added. If we leave VectorData.add_row to handle the validation, one value at a time, the user might end up in a situation where the value of the first column is valid and added and the value of the second column is invalid and raises an error. This will result in an invalid table that is difficult to fix. Can you add this validation code back plus a comment about why it's here.

mavaylon1: This was the double validation we talked about yesterday. I agree the value here is probably worth the double validation.

src/hdmf/common/table.py
```diff
@@ -708,19 +681,7 @@ def add_column(self, **kwargs):  # noqa: C901
...
-            bad_data = []
-            for val in data:
-                if term_set.validate(term=val):
-                    continue
-                else:
-                    bad_data.append(val)
-            if len(bad_data)!=0:
-                bad_data_string = str(bad_data)[1:-1]
-                msg = ("%s is not in the term set." % bad_data_string)
-                raise ValueError(msg)
```
rly, line 723: Same here. Either all the values get added or an error is raised and none of them get added. This means that data will be validated twice, but I think that's OK.

mavaylon1: Agreed. Will do

rly: Alternative, we could "roll back" the new additions, but I think that would be messy.

src/hdmf/container.py
```diff
@@ -847,18 +824,7 @@ def extend(self, arg):
...
-        else:
-            bad_data = []
-            for item in arg:
-                try:
-                    self.append(item)
-                except ValueError:
-                    bad_data.append(item)
-            if len(bad_data)!=0:
-                msg = ('"%s" is not in the term set.' % ', '.join([str(item) for item in bad_data]))
-                raise ValueError(msg)
```
rly, line 861: I think we should keep this validation as well for the same reasons mentioned above.

src/hdmf/term_set.py
```diff
@@ -169,3 +171,102 @@ def __enum_expander(self):
...
+        """
+        bad_data = []
+        for item in arg:
+            try:
+                self.append(item)
+            except ValueError:
+                bad_data.append(item)
+        if len(bad_data)!=0:
+            msg = ('"%s" is not in the term set.' % ', '.join([str(item) for item in bad_data]))
+            raise ValueError(msg)
```
rly: Here also I think we should validate all items before adding any items

src/hdmf/container.py
mavaylon1: I don't think we would need this here. This was moved to the wrapper. The logic remains the same. Now it calls the wrapper append, which is essentially this.

src/hdmf/term_set.py
```diff
@@ -169,3 +171,102 @@ def __enum_expander(self):
...
+        """
+        This is used when we want to index items.
+        """
+        return self.__value[val]
+
+    def __next__(self):
+        """
+        We want to make sure all iterators are still valid.
+        """
+        return self.__value.__next__()
```
rly: Can you add a test that includes running this line?

tests/unit/helpers/utils.py
```diff
@@ -203,7 +207,7 @@ def foo_ref_attr(self, value):
             raise ValueError("can't reset foo_ref_attr attribute")
 
 
-def get_foo_buildmanager():
+def get_foo_buildmanager(dtype="int"):
```
rly:
```suggestion
def get_foo_buildmanager(my_data_dtype="int"):
```
`dtype` is very ambiguous since there are multiple objects with dtypes.

tests/unit/helpers/utils.py
```diff
@@ -215,7 +219,7 @@ def get_foo_buildmanager():
         datasets=[
             DatasetSpec(
                 "an example dataset",
-                "int",
+                dtype,
```
rly:
```suggestion
                my_data_dtype,
```

tests/unit/helpers/utils.py
```diff
@@ -72,6 +72,10 @@ def __str__(self):
     def my_data(self):
         return self.__data
 
+    # @property
+    # def data(self):
+    #     return self.__data
+
```
rly:
```suggestion
```

src/hdmf/build/objectmapper.py
```diff
@@ -564,7 +565,11 @@ def get_attr_value(self, **kwargs):
                 msg = ("%s '%s' does not have attribute '%s' for mapping to spec: %s"
                        % (container.__class__.__name__, container.name, attr_name, spec))
                 raise ContainerConfigurationError(msg)
-            if attr_val is not None:
+            if isinstance(attr_val, TermSetWrapper):
+                # This is when the wrapped item is an attribute
+                # Refer to h5tools.py for wrapped datasets
```
rly:
```suggestion
```
Actually this is used for some DatasetSpecs too (I think just those without a data_type). That is evidenced by the fact that this function was called when you changed the "my_data" DatasetSpec

docs/gallery/plot_term_set.py
```diff
@@ -114,71 +118,85 @@
...
-# Validate Data in a DynamicTable with TermSet
+# Validate Data in a DynamicTable
 # ----------------------------------------------------
-# Validating data with :py:class:`~hdmf.common.table.DynamicTable` is determined by which columns were
-# initialized with the term_set attribute set. The data is validated when the columns are created or
-# modified. Since adding the columns to a DynamicTable does not modify the data, validation is
-# not being performed at that time.
+# Validating data for :py:class:`~hdmf.common.table.DynamicTable` is determined by which columns were
+# initialized with a :py:class:`~hdmf.term_set.TermSetWrapper`. The data is validated when the columns
+# created.
```
```suggestion
# are created and modified using ``DynamicTable.add_row``.
```

docs/gallery/plot_term_set.py
```diff
@@ -14,15 +15,18 @@
...
 The :py:class:`~hdmf.term_set.TermSet` class is used closely with
-:py:class:`~hdmf.common.resources.ExternalResources` to more efficiently map terms
-to data. Please refer to the tutorial on ExternalResources to see how :py:class:`~hdmf.term_set.TermSet`
-is used with :py:class:`~hdmf.common.resources.ExternalResources`.
+:py:class:`~hdmf.common.resources.HERD` to more efficiently map terms
+to data.
+
+In order to actually use a :py:class:`~hdmf.term_set.TermSet`, users will use the
+:py:class:`~hdmf.term_set.TermSetWrapper` to wrap data and attributes. The
+:py:class:`~hdmf.term_set.TermSetWrapper` uses a user provided :py:class:`~hdmf.term_set.TermSet`
```
```suggestion
:py:class:`~hdmf.term_set.TermSetWrapper` uses a user-provided :py:class:`~hdmf.term_set.TermSet`
```

docs/gallery/plot_term_set.py
```diff
@@ -114,71 +118,85 @@
...
-# Validate new rows in a DynamicTable with TermSet
-# ----------------------------------------------------
+##########################################################
+# Validate new rows in a DynamicTable with TermSetWrapper
+# --------------------------------------------------------
 # Validating new rows to :py:class:`~hdmf.common.table.DynamicTable` is simple. The
 # :py:func:`~hdmf.common.table.DynamicTable.add_row` method will automatically check each column for a
-# :py:class:`~hdmf.term_set.TermSet` (via the term_set attribute). If the attribute is set, the the data will be
-# validated for that column using that column's :py:class:`~hdmf.term_set.TermSet`. If there is invalid data, the
+# :py:class:`~hdmf.term_set.TermSetWrapper`. If a wrapper is being used, the the data will be
```
```suggestion
# :py:class:`~hdmf.term_set.TermSetWrapper`. If a wrapper is being used, then the data will be
```

docs/gallery/plot_term_set.py
```diff
@@ -114,71 +118,85 @@
...
-# As mentioned prior, validating in a :py:class:`~hdmf.common.table.DynamicTable` is determined
-# by the columns. The :py:func:`~hdmf.common.table.DynamicTable.add_column` method has a term_set attribute
-# as if you were making a new instance of :py:class:`~hdmf.common.table.VectorData`. When set, this attribute
-# will be used to validate the data. The column will not be added if there is invalid data.
+#############################################################
+# Validate new columns in a DynamicTable with TermSetWrapper
+# -----------------------------------------------------------
+# To add a column that has been validated using :py:class:`~hdmf.term_set.TermSetWrapper`,
+# users would wrap the data in the :py:func:`~hdmf.common.table.DynamicTable.add_column`
+# method as if you were making a new instance of :py:class:`~hdmf.common.table.VectorData`.
```
```suggestion
# To add a column that is validated using :py:class:`~hdmf.term_set.TermSetWrapper`,
# wrap the data in the :py:func:`~hdmf.common.table.DynamicTable.add_column`
# method as if you were making a new instance of :py:class:`~hdmf.common.table.VectorData`.
```

I think we can also remove the `VectorData` example below.

tests/unit/test_term_set.py
```diff
@@ -101,3 +103,83 @@ def test_folder_output(self):
...
+        self.assertEqual(data_obj.data.value, ['Homo sapiens', 'Mus musculus'])
+
+    def test_wrapper_extend_error(self):
+        data_obj = VectorData(name='species', description='...', data=self.wrapped_list)
+        with self.assertRaises(ValueError):
+            data_obj.extend(['bad_data'])
+
+    def test_wrapper_next(self):
+        with self.assertRaises(AttributeError):
+            next(self.wrapped_list)
```
```suggestion
    def test_wrapper_next(self):
        data_obj = VectorData(name='species', description='...', data=self.wrapped_list)
        data_obj.extend(['Mus musculus'])
        it = iter(data_obj.data)
        self.assertEqual(next(it), "Homo sapiens"))
        self.assertEqual(next(it), "Mus musculus"))
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## HDMF 3.9.1 (Upcoming)
 
+### Enhancements
+- Updated `TermSet` to be used with `TermSetWrapper`, allowng for general use of validation for datasets and attributes. This also brings updates to `HERD` integration and updates on `write` to easily add references for wrapped datasets/attributes. @mavaylon1 [#950](https://github.com/hdmf-dev/hdmf/pull/950)
```
```suggestion
- Updated `TermSet` to be used with `TermSetWrapper`, allowing for general use of validation for datasets and attributes. This also brings updates to `HERD` integration and updates on `write` to easily add references for wrapped datasets/attributes. @mavaylon1 [#950](https://github.com/hdmf-dev/hdmf/pull/950)
```

src/hdmf/common/resources.py
oruebel: Instead of using `dir` to determine `obj_fields` I think you should be able to just use``__dict__`` https://docs.python.org/3/library/stdtypes.html#object.__dict__  :

```
for attribute_name, attribute_value in obj.__dict__.items():
    if isinstance(attribute_value, TermSetWrapper):
          ret.append([obj, attribute_name, attribute_value])
```

src/hdmf/common/resources.py
```diff
@@ -408,7 +409,29 @@ def _get_file_from_container(self, **kwargs):
...
+
+            for attribute in obj_fields:
+                attr = getattr(obj, attribute)
+                if isinstance(attr, TermSetWrapper):
+                    # Search objects that are wrapped
+                    ret.append([obj, attribute, attr])
```
oruebel: Instead of a list of lists, a slightly more Pythonic way would be to use a list of `namedtuple` https://docs.python.org/3/library/collections.html#collections.namedtuple The main advantage being is that is enforces the type of objects the list contains and it documents what each item in the list and item within the tuple means.

src/hdmf/common/resources.py
mavaylon1: It's either dir(obj) or obj.__dict__.keys(). Either way I have to still clean out the "_". I don't think there's a tangible benefit on switching.

src/hdmf/common/resources.py
mavaylon1: Done

src/hdmf/term_set.py
```diff
@@ -169,3 +171,103 @@ def __enum_expander(self):
...
+
```
```suggestion
```

@mavaylon1 as you mentioned, this is only used for iterator objects and not needed here. It is also not needed/used  and is broken in DataIO and should be removed from there.

tests/unit/test_term_set.py
```diff
@@ -101,3 +103,86 @@ def test_folder_output(self):
...
+        data_obj = VectorData(name='species', description='...', data=self.wrapped_list)
+        with self.assertRaises(ValueError):
+            data_obj.extend(['bad_data'])
+
+    def test_wrapper_next(self):
+        data_obj = VectorData(name='species', description='...', data=self.wrapped_list)
+        data_obj.extend(['Mus musculus'])
+        it = iter(data_obj.data)
+        self.assertEqual(next(it), "Homo sapiens")
+        self.assertEqual(next(it), "Mus musculus")
```
rly: We can remove this too.
```suggestion
```

src/hdmf/term_set.py
rly: Actually, never mind. DataChunkIterator is one such iterator that might be wrapped by DataIO and by TermSetWrapper. We need to be able to call `__next__` on that. `__next__` will fail for things that are not iterators, but that is to be expected.

```suggestion
    # uncomment when DataChunkIterator objects can be wrapped by TermSet
    # def __next__(self):
    #     """
    #     Return the next item of a wrapped iterator.
    #     """
    #     return self.__value.__next__()
    #
```
TermSetWrapper does not currently support DataChunkIterators or DataIOs to be wrapped. Let's comment this out for now.

tests/unit/test_term_set.py
rly: This ended up not calling `__next__`. We would need to wrap an iterator which is currently not supported.

rly (approved):

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -1017,39 +1071,35 @@ def test_io_read_herd_value_warn(self):
...
-        data = Data(name="species", data=['Homo sapiens', 'Mus musculus'])
-        er.add_ref(file=self.foofile,
-                   container=data,
-                   key='key1',
-                   entity_id='entity_id1',
-                   entity_uri='entity1')
-        with HDF5IO(self.path, manager=self.manager, mode='w', herd_path='./HERD.zip') as io:
-            with self.assertWarns(Warning):
-                io.write(self.foofile)
+        self.remove_er_files()
```
rly: Can you test/assert whether the ref gets written in the HERD file

Can you also test whether passing HERD to `io.write` works?

rly (changes requested):

tests/unit/test_io_hdf5_h5tools.py
mavaylon1: The extend is in the tests for write. I did add more asserts now to check the HERD itself.

rly: > The extend is in the tests for write.

Can you point me to the test?

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -1017,39 +1071,35 @@ def test_io_read_herd_value_warn(self):
...
-        self.foofile.link_resources(er)
-
-        data = Data(name="species", data=['Homo sapiens', 'Mus musculus'])
-        er.add_ref(file=self.foofile,
-                   container=data,
-                   key='key1',
-                   entity_id='entity_id1',
-                   entity_uri='entity1')
+    @unittest.skipIf(not LINKML_INSTALLED, "optional LinkML module is not installed")
+    def test_io_write_extend_herd(self):
```
mavaylon1, line 1075: @rly This is the extend test

rly: Got it. That's what I was pointing to, but for some reason github showed a weird location in the diff.

rly (approved):

rly:
image

This is what I like to see. One of the U24 milestones is to reach 95% coverage by year 5 (ending Jan 2026). So we should strive to reach 100% patch coverage when possible.

Issue #893, opened 2023-07-06
[Feature]: Enhance TermSet


rly:
### What would you like to see added to HDMF?

See https://github.com/hdmf-dev/hdmf/pull/880#pullrequestreview-1517181868

- Moving `ExternalResources`, `ExternalResourceMangager`, `TermSet` etc. codes into a separate module (folder) for external_resources
- Add `TermSet._repr_html_` function similar to #883 for to be able to render term sets nicely in the docs and Jupyter notebooks
- Add read-only properties to `TermSet` for “name”, “version”, and “prefixes” (which are probably just shortcuts for lookups in `TermSet.view`)

### What solution would you like?

See above

### Do you have any interest in helping implement the feature?

No.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/hdmf-dev/hdmf/issues)?

oruebel:
I think at least moving ExternalResources, ExternalResourceMangager, TermSet etc. codes into a separate module would be useful to do before the release

mavaylon1:
I'd prefer not. I want to get everything out there now and focus on finalizing tutorials and the talk. @oruebel

PR #967, opened 2023-10-17
TermSet minor test updates, _repr_html_, name field

mavaylon1:
## Motivation

- [x] Add _repr_html_ to Termset
- [x] Add name to Termset
- [x] Update tests to include name

## How to test the behavior?
```
```
Run notebook for html representation and then the unit tests.

## Checklist


mavaylon1:
Fix #893

mavaylon1:
It's been a while since #893 but we don't have versions anymore, the prefixes can be called by "termset.sources", I added the name, I added _repr_html_

mavaylon1:
I did not move termset and herd to a folder as the issue ticket suggested. That would require a lot of import changes for something that doesn't provide much value to the user, nor me as the main person who works on this.

src/hdmf/term_set.py
```diff
@@ -35,6 +37,7 @@ def __init__(self,
             msg = "Install linkml_runtime"
             raise ValueError(msg)
 
+        self.name = name
```
oruebel: @mavaylon1 what is `TermSet.name` being used for? I didn't see where the variable is actually being used in this PR. Is this preparing for some additional changes to come?

mavaylon1: I thought it could be useful for the user to name their termsets to more easily identify them without needing to open the schema.

rly:
The YAML schema in `TermSet.term_schema_path` already has a name. I think it is required but I am not sure. Could we extract the name field from the schema? I think that is what I meant in #893. Add convenience properties `name`, `version`, and `prefixes` that allow the user to get those values from the YAML schema.

src/hdmf/term_set.py
```diff
@@ -55,9 +58,14 @@ def __init__(self,
...
-        re = "class: %s\n" % str(self.__class__)
-        re += "term_schema_path: %s\n" % self.term_schema_path
+    def __repr__(self): # pragma: no cover
+        re = "Class: %s\n" % str(self.__class__.__name__)
+        re += "Schema Path: %s\n" % self.term_schema_path
+        return re
+
+    def _repr_html_(self): # pragma: no cover
+        re = "<p><b>" + "Schema Path: " + "</b>" + self.term_schema_path + "<br>" + "</p>"
+        re += "<p><b>" + "Sources: " + "</b>" + ", ".join(list(self.sources.keys())) + "</p>"
```
rly: 1. Could you add a screenshot in the PR showing how this looks?
2. This and `__repr__` are testable. Please do not exclude lines of code from coverage checks unless there is a very good reason why that code will not ever run, e.g., some code exists as an additional just-in-case error check but if everything is configured by HDMF correctly, that error should never run. In this case, the code is very likely to run and should be tested to make sure there is not a silly error within it.
3. I think `__repr__` and `_repr_html_` should display the same information. Showing the sources is nice - why not add that to `__repr__`?

mavaylon1: 1. I can add a screenshot.
2. We didn't have tests before for them but we should? It seems a bit unnecessary for these methods.
3. I can make them the same, but I treated them differently based on where I interact with the class. In terminal it was more useful for me to see the path than the sources themselves. I think what I'll do is make sure both representations have sources and path.

mavaylon1:
> The YAML schema in `TermSet.term_schema_path` already has a name. I think it is required but I am not sure. Could we extract the name field from the schema? I think that is what I meant in #893. Add convenience properties `name`, `version`, and `prefixes` that allow the user to get those values from the YAML schema.

That makes sense to extract.

src/hdmf/term_set.py
rly: re: 2. I think we should. If the code were as simple as just returning the name, then maybe not. Here, you have slightly more complex code where you could have had a typo, used += instead of =, the `join` used the wrong syntax, self.sources might not have been initialized, etc. Better to spend the 3 minutes to write at least one test to make sure the code runs as expected.

rly: re: 2. I think we should. If the code were as simple as just returning the name, then maybe not. Here, you have slightly more complex code where you could have had a typo, used `=` instead of `+=`, the `join` used the wrong syntax, `self.sources` might not have been initialized, etc. Better to spend the 3 minutes to write at least one test to make sure the code runs as expected than to have a user discover the bug down the road.

rly:
@mavaylon1 thanks for the PR. I think printing "[item 1] [item 2] ... [item 3]" and "[item 1] [item 2] ... [item 3] [item 4]" when there are only 3 or 4 items in the list is misleading, so I changed the format a bit so that the ellipses only show up if there are more items present than are shown. Super minor, but I think it is better this way.

Also, the panda that you added to your example is a genus, not a species, so I updated that accordingly.

rly (approved):

Issue #961, opened 2023-10-05
[Feature]: Reusing Keys  and Entities should be automatic 


mavaylon1:
### What would you like to see added to HDMF?

When we first made HERD, it was not built for "bulk" adding. You could loop add_ref; however, when reusing keys and entities you need to change the parameters of add_ref. For keys, you would need to use `get_key` to use the key object in add_ref. For entity, you would need to remove the uri parameter. 

Say we wanted to use a DANDI set of nwbfiles and add references for subject and experimenter. It's not user friendly to have to have a try/except set up based on whether they key or entity exists. 
```
from pynwb.resources import HERD
from pynwb import NWBHDF5IO, NWBFile
from glob import glob
from tqdm import tqdm

# Path to all the files
path = '/Users/mavaylon/Research/NWB/000015/sub*'

# Create HERD
herd = HERD()

# populate iteratively
folders = glob(path)
for folder in folders:
    for file in tqdm(glob(folder+'/*')):
        io = NWBHDF5IO(file, mode='r')
        read_file = io.read()
        #Add HERD for Subject
        try:
            entity = herd.get_entity(entity_id='NCBI_TAXON:10090')
            if entity is not None:
                raise ValueError()
            else:
                herd.add_ref(file=read_file,
                             container=read_file.subject,
                             key=read_file.subject.species,
                             entity_id = 'NCBI_TAXON:10090', # this assumes the same species for each file
                             entity_uri = 'https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=NCBI_TAXON:10090'
                             )
        except ValueError: # after the first use of an entity_id and key, you are required to reuse them
            herd.add_ref(file=read_file,
                         container=read_file.subject,
                         key=read_file.subject.species,
                         entity_id = 'NCBI_TAXON:10090'
                         )


        # Add HERD for Experimenter
        try:
            if len(read_file.experimenter)>1:
                breakpoint()
            herd.get_entity(entity_id='0000-0001-6782-3819')
            if entity is not None:
                raise ValueError()
            else:
                herd.add_ref(file=read_file,
                             container=read_file,
                             attribute="experimenter",
                             key=read_file.experimenter[0], # this assumes the experimenter is the same for each file
                             entity_id = '0000-0001-6782-3819',
                             entity_uri = 'https://orcid.org/0000-0001-6782-3819'
                             )

        except ValueError:
            herd.add_ref(file=read_file,
[...]

PR #968, opened 2023-10-18
HERD Updates

mavaylon1:
## Motivation

- [x] HERD should resolve the reuse of keys when it is being associated with the same object. This will allow us to bulk add but also gives the user to still manually choose a key that exists if they want to use that key later on.
- [x] HERD should do checks before populating the tables. Currently, the checks are spread out (some we may not be able to move so TBD). This leads to some tables being populated and some are not when an error is thrown. If the checks can't be moved as easily as I think, then we would need to have something that removes what was added. I prefer the former to the latter.
- [x] HERD should rename add_ref_termset to maybe "add_ref_container" since it is only called in the I/O for write.
- [x] HERD should have a add_ref_termset that actually uses termsets instead of entity id or uri. We had this prior but this changed when we introduced the wrapper. This would bring the essence of that back.
- [x] HERD should resolve existing entities with a warning and use the uri of current entity and not the parameter.
- [x] Add tests for coverage
- [x] Update tutorial to account for add_ref_termset

## How to test the behavior?
```
```
Unit Tests

## Checklist


mavaylon1:
Fix #961

mavaylon1:
This PR has grown to cover multiple items.

- add_ref_termset
- resolving keys within the same object
- resolving entities to ignore new a uri

I remembered a question about `get_key` and I want to get better at limiting the scope of PRs.  I will continue off of the original PR: https://github.com/hdmf-dev/hdmf/pull/964 
The idea is that `get_key` should: 

- If no key is found return None instead of raising a ValueError
- If multiple matching keys are found, then return a list of Key objects rather than raising a ValueError
I don't think changing to None is more informative. I like the fact the error tells you why. As for the second one, I think that's fine.

docs/gallery/plot_external_resources.py
```diff
@@ -306,6 +307,52 @@ def __init__(self, **kwargs):
     entity_uri='https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=10090'
 )
 
+###############################################################################
+# Using add_ref_termset
+# ------------------------------------------------------
+# :py:class:`~hdmf.common.resources.HERD` has multiple ways for users to add
+# external references. The :py:func:`~hdmf.common.resources.HERD.add_ref_termset`
+# method allows users to not only validate terms, i.e., keys, but also adds the
+# ability to iteratively add references for entire datasets, lists, arrays, etc.
```
rly:
```suggestion
# method allows users to not only validate terms, i.e., keys, but also 
# add references for entire datasets, lists, arrays, etc.
```
Fix grammar and improve clarity

docs/gallery/plot_external_resources.py
```diff
@@ -306,6 +307,52 @@ def __init__(self, **kwargs):
...
+
+# The :py:func:`~hdmf.common.resources.HERD.add_ref_container` method is directly
+# used for populating :py:class:`~hdmf.common.resources.HERD` when writing a file.
+# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` can be used for new files;
+# however, it is also the best practice when adding references for existing files.
```
rly: I'm a bit confused here. When should a user use `add_ref_container`? Or should that be used only internally?

docs/gallery/plot_external_resources.py
```diff
@@ -306,6 +307,52 @@ def __init__(self, **kwargs):
...
+
+# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` has many optional fields,
+# giving the user a range of control when adding references. Let's see an example.
+er = HERD()
```
rly:
```suggestion
herd = HERD()
```
Can we name the variable `er` to `herd` throughout these examples? I don't think `er` would mean anything to a new user, now that we have renamed everything to HERD.

docs/gallery/plot_external_resources.py
```diff
@@ -306,6 +307,52 @@ def __init__(self, **kwargs):
...
+# Using add_ref_termset for an entire dataset
+# ------------------------------------------------------
+# As mentioned above, :py:func:`~hdmf.common.resources.HERD.add_ref_termset`
+# supports iteratively validating and populating :py:class:`~hdmf.common.resources.HERD`.
+
+# When populating :py:class:`~hdmf.common.resources.HERD`, users may have some terms
+# that are not in the :py:class:`~hdmf.term_set.TermSet`. As a result,
+# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` will return an all the missing
+# terms in a dictionary. It is up to the user to either add these terms to the
+# :py:class:`~hdmf.term_set.TermSet` or remove them from the dataset.
```
rly: I thought `add_ref_termset` would raise a validation error if a user tries to add terms not in the term set. Is that not correct?

docs/gallery/plot_external_resources.py
```diff
@@ -306,6 +307,52 @@ def __init__(self, **kwargs):
...
+# The :py:func:`~hdmf.common.resources.HERD.add_ref_container` method is directly
+# used for populating :py:class:`~hdmf.common.resources.HERD` when writing a file.
+# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` can be used for new files;
+# however, it is also the best practice when adding references for existing files.
+
+# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` has many optional fields,
+# giving the user a range of control when adding references. Let's see an example.
+er = HERD()
+terms = TermSet(term_schema_path='example_term_set.yaml')
+file = HERDManagerContainer(name='file')
```
rly: Why do we have to create a new file here? If we create a new file, should we also create a new container? Right now the species Data container isn't even in this file. Can we add a sanity check in `add_ref_termset` that if `file` and `container` are both passed that `container` is actually part of the `file`? This could be an easy way for users to accidentally create nonsense references. I think this example would create a nonsense reference.

src/hdmf/common/resources.py
```diff
@@ -358,16 +361,17 @@ def _check_object_field(self, **kwargs):
         relative_path = kwargs['relative_path']
         field = kwargs['field']
         create = kwargs['create']
+
         file_object_id = file.object_id
         files_idx = self.files.which(file_object_id=file_object_id)
 
         if len(files_idx) > 1:
```
```suggestion
        if len(files_idx) > 1:  # pragma: no cover
```

src/hdmf/common/resources.py
```diff
@@ -378,10 +382,16 @@ def _check_object_field(self, **kwargs):
...
-            return self._add_object(files_idx=files_idx, container=container, relative_path=relative_path, field=field)
+            # Used for add_ref
+            return {'file_object_id': file_object_id,
+                    'files_idx': files_idx,
+                    'container': container,
+                    'relative_path': relative_path,
+                    'field': field}
         elif len(objecttable_idx) == 0 and not create:
             raise ValueError("Object not in Object Table.")
         else:
```
```suggestion
        else:  # pragma: no cover
```

src/hdmf/common/resources.py
```diff
@@ -466,63 +476,69 @@ def add_ref_term_set(self, **kwargs):
...
+             'doc': 'The TermSet to be used if the container/attribute does not have one.'}
+            )
+    def add_ref_termset(self, **kwargs):
         """
-        Return a Key.
-
-        If container, relative_path, and field are provided, the Key that corresponds to the given name of the key
-        for the given container, relative_path, and field is returned.
+        This method allows users to take advantage of using the TermSet class to provide the entity information
+        for add_ref, while also validating the data. This method supports adding a single key and an entire dataset
```
```suggestion
        for add_ref, while also validating the data. This method supports adding a single key or an entire dataset
```

src/hdmf/common/resources.py
```diff
@@ -466,63 +476,69 @@ def add_ref_term_set(self, **kwargs):
...
+        to the HERD tables. For both cases, the term, i.e., key, will be validated against the permissible values
+        in the TermSet. If valid, it will proceed to call add_ref. Otherwise, the method will return a dict of
+        missing terms.
```
```suggestion
        missing terms (terms not found in the TermSet).
```

docs/gallery/plot_external_resources.py
oruebel: As far as I can tell, I think `add_ref_container` in the code is right now used only internally. I don't think `add_ref_container` applies to existing files, as that would require wrapping fields in Containers that have been read from disk with TermSetWrapper, which I don't think is supported right now. For existing files, one would use `add_ref_termset`.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## HDMF 3.11.1 (Upcoming)
 
+### Enhancements
+- Added `add_ref_termset`, updated helper methods for `HERD`, and revised `add_ref` to support validations prior to populating the tables. @mavaylon1 [#968](https://github.com/hdmf-dev/hdmf/pull/968)
```
oruebel: I believe `add_ref_container` is also new and should be mentioned here.

docs/gallery/plot_external_resources.py
oruebel: I'm not sure if/when a user would need to call ``add_ref_container`` directly.

src/hdmf/common/resources.py
```diff
@@ -466,63 +476,69 @@ def add_ref_term_set(self, **kwargs):
...
-            return self.entities.row[entity[0]]
-        else:
-            return None
+                data_object = getattr(container, attribute)
+            if isinstance(data_object, (Data, DataIO)):
+                data = data_object.data
+            elif isinstance(data_object, (list, tuple, np.ndarray)):
+                data = data_object
+        missing_terms = []
+        for term in data:
```
rly, line 528: There is a possibility that `data` is not set in the code above. e.g., if `data` is a DataChunkIterator or an h5py.Dataset. Can you add an else statement below 522 that raises an error that says something like "data_object is not a Data, DataIO, list, tuple, or numpy ndarray and cannot be used in add_ref_termset"?

src/hdmf/common/resources.py
```diff
@@ -466,63 +476,69 @@ def add_ref_term_set(self, **kwargs):
...
+            entity_uri = term_info[2]
+            self.add_ref(file=file,
+                         container=container,
+                         attribute=attribute,
+                         key=term,
+                         field=field,
+                         entity_id=entity_id,
+                         entity_uri=entity_uri)
+        if len(missing_terms)>0:
+            return {"missing_terms": missing_terms}
```
rly, line 545: I forget - why do we return a dict here with a single key-value pair instead of just the list of missing terms?

src/hdmf/common/resources.py
```diff
@@ -605,69 +667,141 @@ def add_ref(self, **kwargs):
...
-                if not obj_key_check:
-                    self._add_object_key(object_field, key)
-            else:
-                msg = "Cannot find key object. Create new Key with string."
-                raise ValueError(msg)
-            # check if the key and object have been related in the ObjectKeyTable
+                        obj_key_exists = True
+                        # this means there is already a object-key relationship recorded
+                if not obj_key_exists:
+                    # this means that though the key is there, there is not object-key relationship
```
```suggestion
                    # this means that though the key is there, there is no object-key relationship
```

src/hdmf/common/resources.py
```diff
@@ -605,69 +667,141 @@ def add_ref(self, **kwargs):
...
                     entity_idx = self.entity_keys['entities_idx', row_idx]
                     if entity_idx == entity.idx:
                         entity_key_check = True
-                        # this means there is already a key-entity relationship recorded
+                        # this means there is already a entity-key relationship recorded
                 if not entity_key_check:
-                    # this means that though the key is there, there is not key-entity relationship
-                    # a.k.a add it now
-                    self._add_entity_key(entity, key)
+                    # this means that though the key is there, there is not entity-key relationship
```
```suggestion
                    # this means that though the key is there, there is no entity-key relationship
```

rly:
Thanks for all this work and sorry for my delayed review. Overall it looks good. I made some comments above.

I also think we should change the examples/tutorial to be more realistic. Right now, you create a `HERDManagerContainer` object to represent a file, but in the later examples, the file object and the file of the data object do not match each other, and I think that is confusing.

```python
# Class to represent a file
class HERDManagerContainer(Container, HERDManager):
    def __init__(self, **kwargs):
        kwargs['name'] = 'HERDManagerContainer'
        super().__init__(**kwargs)
```

Instead, I suggest using a `SimpleMultiContainer` as the base for the file, and in each example, create a new file, create a new data object, and add the data object to the file before using it:
```python
# Class to represent a file
from hdmf.common import SimpleMultiContainer
class HdmfFile(SimpleMultiContainer, HERDManager):
    def __init__(self, **kwargs):
        kwargs['name'] = 'SimpleMultiContainer'
        super().__init__(**kwargs)

...

file = HdmfFile()
species = DynamicTable(name='species', description='My species', columns=[col1])
file.add_container(species)
```

Then in the examples, the `file` argument does not need to be passed because it can be retrieved from the `data` object.
```python
er.add_ref_termset(
	container=species,
	attribute='Species_Data',
	key='Ursus arctos horribilis',
	termset=terms
)
```
This is a more realistic use case, in my opinion. This structure also has the advantage of making the code examples being more standalone / not dependent on earlier objects, except for the definition of `HdmfFile`.

All of this would help make this tutorial more analogous to its use in PyNWB, so that we could copy and paste relevant parts of this tutorial to create a PyNWB tutorial with minimal modifications, e.g., `HdmfFile` -> `NWBFile`. `add_container` -> `add_acquisition`. Perhaps all of this should be a separate PR though. Up to you. Happy to help with these suggested changes.

mavaylon1:
> Thanks for all this work and sorry for my delayed review. Overall it looks good. I made some comments above.
> 
> I also think we should change the examples/tutorial to be more realistic. Right now, you create a `HERDManagerContainer` object to represent a file, but in the later examples, the file object and the file of the data object do not match each other, and I think that is confusing.
> 
> ```python
> # Class to represent a file
> class HERDManagerContainer(Container, HERDManager):
>     def __init__(self, **kwargs):
>         kwargs['name'] = 'HERDManagerContainer'
>         super().__init__(**kwargs)
> ```
> 
> Instead, I suggest using a `SimpleMultiContainer` as the base for the file, and in each example, create a new file, create a new data object, and add the data object to the file before using it:
> 
> ```python
> # Class to represent a file
> from hdmf.common import SimpleMultiContainer
> class HdmfFile(SimpleMultiContainer, HERDManager):
>     def __init__(self, **kwargs):
>         kwargs['name'] = 'SimpleMultiContainer'
>         super().__init__(**kwargs)
> 
> ...
> 
> file = HdmfFile()
> species = DynamicTable(name='species', description='My species', columns=[col1])
> file.add_container(species)
> ```
> 
> Then in the examples, the `file` argument does not need to be passed because it can be retrieved from the `data` object.
> 
> ```python
> er.add_ref_termset(
> 	container=species,
> 	attribute='Species_Data',
> 	key='Ursus arctos horribilis',
> 	termset=terms
> )
> ```
> 
> This is a more realistic use case, in my opinion. This structure also has the advantage of making the code examples being more standalone / not dependent on earlier objects, except for the definition of `HdmfFile`.
> 
> All of this would help make this tutorial more analogous to its use in PyNWB, so that we could copy and paste relevant parts of this tutorial to create a PyNWB tutorial with minimal modifications, e.g., `HdmfFile` -> `NWBFile`. `add_container` -> `add_acquisition`. Perhaps all of this should be a separate PR though. Up to you. Happy to help with these suggested changes.

I think this is a great idea. I'll make it a separate ticket.

docs/gallery/plot_external_resources.py
mavaylon1: No. `add_ref_termset` returns all the missing terms. This way when you bulk add, you see all the bad data vs having to correct one bad data point at a time.

You are thinking of the error being raised within the TermSet class itself.

src/hdmf/common/resources.py
mavaylon1: My personal preference in that it gives information that it is indeed missing terms

docs/gallery/plot_external_resources.py
mavaylon1: good idea

mavaylon1:
@rly A note on your comment of checking that the container is in the file is a good idea. It avoids bad references. I added that change and it would require some major reworking of the test suite. I'd like to make this into a PR that includes the "HDMF_File" as it would go in line with updating the tests. I added the actual code already and a test, these will be flushed out in the next PR. This would also include updates to the doc.

I don't want to manually add the parent in the tests and the documentation then remove them with the HERDFile update.

docs/gallery/plot_external_resources.py
```diff
@@ -166,13 +176,13 @@ def __init__(self, **kwargs):
...
-# to the closest parent, so that ``er.add_ref(container=genotypes, attribute='genotype_name')`` and
-# ``er.add_ref(container=genotypes.genotype_name, attribute=None)`` will ultimately both use the ``object_id``
+# to the closest parent, so that ``herd.add_ref(container=genotypes, attribute='genotype_name')`` and
+# ``herd.add_ref(container=genotypes.genotype_name, attribute=None)`` will ultimately both use the ``object_id``
 # of the ``genotypes.genotype_name`` :py:class:`~hdmf.common.table.VectorData` column and
 # not the object_id of the genotypes table.
 
 ###############################################################################
-# Using the add_ref method without the file parameter.
+# Using the add_ref method without the file parametherd.
```
rly:
```suggestion
# Using the add_ref method without the file parameter.
```
Be careful with find and replace all!

src/hdmf/common/resources.py
```diff
@@ -466,64 +476,72 @@ def add_ref_term_set(self, **kwargs):
...
-            return self.entities.row[entity[0]]
-        else:
-            return None
+                data_object = getattr(container, attribute)
+            if isinstance(data_object, (Data, DataIO)):
+                data = data_object.data
+            elif isinstance(data_object, (list, tuple, np.ndarray)):
+                data = data_object
+            else:
+                msg = "The data object being used is not supported. Please review the documentation for support types."
```
```suggestion
                msg = "The data object being used is not supported. Please review the documentation for supported types."
```

docs/gallery/plot_external_resources.py
```diff
@@ -306,14 +316,53 @@ def __init__(self, **kwargs):
...
+
+###############################################################################
+# Using add_ref_termset for an entire dataset
+# ------------------------------------------------------
+# As mentioned above, :py:func:`~hdmf.common.resources.HERD.add_ref_termset`
+# supports iteratively validating and populating :py:class:`~hdmf.common.resources.HERD`.
+
+# When populating :py:class:`~hdmf.common.resources.HERD`, users may have some terms
+# that are not in the :py:class:`~hdmf.term_set.TermSet`. As a result,
+# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` will return an all the missing
```
rly:
```suggestion
# :py:func:`~hdmf.common.resources.HERD.add_ref_termset` will return all of the missing
```

rly (approved):

PR #971, opened 2023-10-24
Add `target_tables` argument to `DynamicTable.__init__`

### LINKED ISSUE #970: [Bug]: Auto-generating subclass of subclass of DynamicTable with DTR fails
### What happened?

`EventsTable` is a subtype of `DynamicTable` and specifies a `DynamicTableRegion` that should point to an `EventTypesTable`.
`TtlsTable` is a subtype of `EventsTable` and specifies a `DynamicTableRegion` that should point to an `TtlTypesTable`..

Auto-generating a class for `TtlsTable` results in the error below:

### Steps to Reproduce

```shell
- neurodata_type_def: EventsTable
  neurodata_type_inc: DynamicTable
  doc: ...
  datasets:
  - name: event_type
    neurodata_type_inc: DynamicTableRegion
    dims:
    - num_events
    shape:
    - null
    doc: The type of event that occurred. This is represented as a reference to a
      row of the EventTypesTable.
- neurodata_type_def: TtlsTable
  neurodata_type_inc: EventsTable
  doc: ...
  datasets:
  - name: event_type
    neurodata_type_inc: DynamicTableRegion
    dims:
    - num_events
    shape:
    - null
    doc: The type of TTL that occurred. This is represented as a reference to a row
      of the TtlTypesTable.
```

### Traceback

```pytb
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/runner.py:341: in from_call
    result: Optional[TResult] = func()
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/runner.py:372: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/python.py:531: in collect
    self._inject_setup_module_fixture()
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/python.py:545: in _inject_setup_module_fixture
    self.obj, ("setUpModule", "setup_module")
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/python.py:310: in obj
    self._obj = obj = self._getobj()
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/python.py:528: in _getobj
    return self._importtestmodule()
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/python.py:617: in _importtestmodule
    mod = import_path(self.path, mode=importmode, root=self.config.rootpath)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/pathlib.py:567: in import_path
    importlib.import_module(module_name)
../../../mambaforge/envs/ndx-events/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1204: in _gcd_import
    ???
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1147: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:690: in _load_unlocked
    ???
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/_pytest/assertion/rewrite.py:178: in exec_module
    exec(co, module.__dict__)
src/pynwb/tests/unit/test_events.py:5: in <module>
    from ndx_events import EventsTable, EventTypesTable, Task
src/pynwb/ndx_events/__init__.py:29: in <module>
    TtlTypesTable = get_class('TtlTypesTable', 'ndx-events')
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/utils.py:668: in func_call
    return func(**pargs)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/pynwb/__init__.py:199: in get_class
    return __TYPE_MAP.get_dt_container_cls(neurodata_type, namespace)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/utils.py:664: in func_call
    return func(args[0], **pargs)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/build/manager.py:527: in get_dt_container_cls
    cls = self.__class_generator.generate_class(data_type, spec, parent_cls, attr_names, self)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/utils.py:664: in func_call
    return func(args[0], **pargs)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/build/classgenerator.py:75: in generate_class
    class_generator.set_init(classdict, bases, docval_args, not_inherited_fields, spec.name)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/common/io/table.py:133: in set_init
    cls._add_to_docval_args(docval_args_local, target_tables_dvarg, err_if_present=True)
../../../mambaforge/envs/ndx-events/lib/python3.11/site-packages/hdmf/build/classgenerator.py:258: in _add_to_docval_args
    raise ValueError("Argument %s already exists in docval args." % arg["name"])
E   ValueError: Argument target_tables already exists in docval args.
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.11

### Package Versions

_No response_

### Code of Conduct

rly:
## Motivation

In a previous PR, we added the ability to specify the target table for a pre-defined `DynamicTableRegion` column in `DynamicTable.__init__` but this was only in generated subclasses of `DynamicTable`. This functionality works well, but could be useful to build in to `DynamicTable` in case users want to write their own custom subclass of `DynamicTable` and want to reuse this generic behavior (such as in pynwb and in extensions). Though in PyNWB, we often have a more specific constructor argument for this: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/misc.py#L160 and some hacky code in `add_row` to handle setting the target table when the first row is added: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/misc.py#L207

So I moved the functionality from the custom class generator for generated `DynamicTable` subclasses to the base `DynamicTable` itself.

This also has the benefit of fixing #970 when there are multiple chains of hierarchy below `DynamicTable`. Fixes #970. 

We can optionally adjust pynwb to reuse this behavior in the Units table and elsewhere.
 
## How to test the behavior?
```
pytest
```

## Checklist

### DIFF
--- a/docs/gallery/plot_dynamictable_howto.py
+++ b/docs/gallery/plot_dynamictable_howto.py
@@ -321,13 +321,20 @@ table3 = DynamicTable(
 ###############################################################################
 # Setting the target table of a DynamicTableRegion column of a DynamicTable
 # -------------------------------------------------------------------------
-# A subclass of DynamicTable might have a pre-defined DynamicTableRegion column
-# that references rows of another table. The target table of this column might
-# not be initialized. In this case, the target table can be set using the
-# ``target_tables`` argument of the DynamicTable constructor. This argument
-# should be a dictionary mapping the name of the DynamicTableRegion column to
-# the target table. The target table can also be set after the DynamicTable
-# has been initialized using ``my_table.my_column.table = other_table``.
+# A subclass of DynamicTable might have a pre-defined DynamicTableRegion column.
+# To write this column correctly, the "table" attribute of the column must be set so
+# that users know to what table the row index values reference. Because the target
+# table could be any table, the "table" attribute must be set explicitly. There are three
+# ways to do so. First, you can use the ``target_tables`` argument of the
+# DynamicTable constructor as shown below. This argument
+# is a dictionary mapping the name of the DynamicTableRegion column to
+# the target table. Secondly, the target table can be set after the DynamicTable
+# has been initialized using ``my_table.my_column.table = other_table``. Finally,
+# you can create the DynamicTableRegion column and pass the ``table``
+# attribute to `DynamicTableRegion.__init__` and then pass the column to
+# `DynamicTable.__init__` using the `columns` argument. However, this approach
+# is not recommended for columns defined in the schema, because it is up to
+# the user to ensure that the column is created in accordance with the schema.
 
 class SubTable(DynamicTable):
     __columns__ = (
--- a/docs/gallery/plot_dynamictable_howto.py
+++ b/docs/gallery/plot_dynamictable_howto.py
@@ -319,6 +319,34 @@ table3 = DynamicTable(
 )
 
 ###############################################################################
+# Setting the target table of a DynamicTableRegion column of a DynamicTable
+# -------------------------------------------------------------------------
+# A subclass of DynamicTable might have a pre-defined DynamicTableRegion column
+# that references rows of another table. The target table of this column might
+# not be initialized. In this case, the target table can be set using the
+# ``target_tables`` argument of the DynamicTable constructor. This argument
+# should be a dictionary mapping the name of the DynamicTableRegion column to
+# the target table. The target table can also be set after the DynamicTable
+# has been initialized using ``my_table.my_column.table = other_table``.
+
+class SubTable(DynamicTable):
+    __columns__ = (
+        {'name': 'dtr', 'description': 'required region', 'required': True, 'table': True},
+    )
+
+referenced_table = DynamicTable(
+    name='referenced_table',
+    description='an example table',
+)
+
+sub_table = SubTable(
+    name='sub_table',
+    description='an example table',
+    target_tables={'dtr': referenced_table},
+)
+# now the target table of the DynamicTableRegion column 'dtr' is set to `referenced_table`
+
+###############################################################################
 # Creating an expandable table
 # ----------------------------
 # When using the default HDF5 backend, each column of these tables is an HDF5 Dataset,
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -292,9 +292,15 @@ class DynamicTable(Container):
             {'name': 'colnames', 'type': 'array_data',
              'doc': 'the ordered names of the columns in this table. columns must also be provided.',
              'default': None},
+            {'name': 'target_tables',
+             'doc': ('dict mapping DynamicTableRegion column name to the table that the DTR points to. The column is '
+                     'added to the table if it is not already present (i.e., when it is optional).'),
+             'type': dict,
+             'default': None},
             allow_positional=AllowPositional.WARNING)
     def __init__(self, **kwargs):  # noqa: C901
         id, columns, desc, colnames = popargs('id', 'columns', 'description', 'colnames', kwargs)
+        target_tables = popargs('target_tables', kwargs)
         super().__init__(**kwargs)
         self.description = desc
 
@@ -468,6 +474,10 @@ class DynamicTable(Container):
         self.__colids = {name: i + 1 for i, name in enumerate(self.colnames)}
         self._init_class_columns()
 
+        if target_tables:
+            self._set_dtr_targets(target_tables)
+
+
     def __set_table_attr(self, col):
         if hasattr(self, col.name) and col.name not in self.__uninit_cols:
             msg = ("An attribute '%s' already exists on %s '%s' so this column cannot be accessed as an attribute, "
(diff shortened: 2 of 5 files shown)


src/hdmf/common/table.py
```diff
@@ -468,6 +474,10 @@ def __init__(self, **kwargs):  # noqa: C901
         self.__colids = {name: i + 1 for i, name in enumerate(self.colnames)}
         self._init_class_columns()
 
+        if target_tables:
+            self._set_dtr_targets(target_tables)
```
oruebel, line 478: Since a user may specify the DTR via the `columns` argument as well, do we need to check that the `target_table` and `columns` are consistent, i.e., what if the target specified in columns is different from what is listed in target_tables?

docs/gallery/plot_dynamictable_howto.py
```diff
@@ -318,6 +318,34 @@
     columns=[dtr_idx, indexed_dtr_col],
 )
 
+###############################################################################
+# Setting the target table of a DynamicTableRegion column of a DynamicTable
+# -------------------------------------------------------------------------
+# A subclass of DynamicTable might have a pre-defined DynamicTableRegion column
+# that references rows of another table. The target table of this column might
+# not be initialized. In this case, the target table can be set using the
```
oruebel: What does ``The target table of this column might not be initialized.`` refer to? Does this mean the table it is pointing to is not initialized or that the attribute `target_table` of the column itself has not been set? I assume it is probably the latter but would be nice to clarify.

docs/gallery/plot_dynamictable_howto.py
```diff
@@ -318,6 +318,34 @@
...
+    description='an example table',
+)
+
+sub_table = SubTable(
+    name='sub_table',
+    description='an example table',
+    target_tables={'dtr': referenced_table},
+)
+# now the target table of the DynamicTableRegion column 'dtr' is set to `referenced_table`
+
```
oruebel, line 355: It would be useful to contrast this approach with how this would look when using the `columns` argument of `DynamciTable.__init__`. E.g., we could add a note to explain why this is useful. Maybe something like:

```suggestion
# now the target table of the DynamicTableRegion column 'dtr' is set to `referenced_table`

.. note::

     Using `target_tables` to specify the target means that we do not need to create 
     `DynamicTableRegion` columns that are specified in the schema manually, but  we can have
     `DynamicTable.__init__` create the column on our behalf. The alternative would be to create
     the `DynamicTableRegion` column and provide it as part of the `columns` argument of 
      `DynamicTable.__init__`, however, this means it is up to the user to ensure that the column is
      created in accordance to the schema of the specific `DynamicTable` type. 
     column  

```

oruebel (commented):
Overall, the approach looks fine to me. However, the specific use I think may be a bit tricky to understand at first, so I think we should make sure to clearly define when to use ``target_table`` vs. when to use the ``columns`` argument of ``DynamicTable.__init__``.

src/hdmf/common/table.py
rly: Yes, that is checked on line 558.

docs/gallery/plot_dynamictable_howto.py
rly: Good point. I will clarify.

docs/gallery/plot_dynamictable_howto.py
```diff
@@ -318,6 +318,34 @@
...
+###############################################################################
+# Setting the target table of a DynamicTableRegion column of a DynamicTable
+# -------------------------------------------------------------------------
+# A subclass of DynamicTable might have a pre-defined DynamicTableRegion column
+# that references rows of another table. The target table of this column might
+# not be initialized. In this case, the target table can be set using the
+# ``target_tables`` argument of the DynamicTable constructor. This argument
+# should be a dictionary mapping the name of the DynamicTableRegion column to
+# the target table. The target table can also be set after the DynamicTable
+# has been initialized using ``my_table.my_column.table = other_table``.
```
rly:
```suggestion
# A subclass of DynamicTable might have a pre-defined DynamicTableRegion column.
# To write this column correctly, the "table" attribute of the column must be set so
# that users know to what table the row index values reference. Because the target
# table could be any table, the "table" attribute must be set explicitly. There are three
# ways to do so. First, you can use the ``target_tables`` argument of the
# DynamicTable constructor as shown below. This argument
# is a dictionary mapping the name of the DynamicTableRegion column to
# the target table. Secondly, the target table can be set after the DynamicTable
# has been initialized using ``my_table.my_column.table = other_table``. Finally,
# you can create the DynamicTableRegion column and pass the ``table``
# attribute to `DynamicTableRegion.__init__` and then pass the column to
# `DynamicTable.__init__` using the `columns` argument. However, this approach
# is not recommended for columns defined in the schema, because it is up to
# the user to ensure that the column is created in accordance with the schema.
```

docs/gallery/plot_dynamictable_howto.py
rly: I tried to address this in the revised text above. Please review.

rly:
Thanks for the quick review!

oruebel (approved):
Thanks for the clarifications in the text. The PR looks good to me.

Issue #959, opened 2023-10-03
[Bug]: Auto-generating MultiContainerInterface subclasses


mavaylon1:
### What happened?

First alerted in #948, there is a bug from auto-generating class that inherits from `MultiContainerInterface`. In this example, `PlaneExtension` extends `PlaneSegmentation`, which inherits from `MultiContainerInterface` (you can see this in the schema with `reference_images` having quantity `*`. With extensions, NWBInspector does not see the python class definitions and will autogenerate the class from the schema. In this example, `PlaneExtension` is auto-generated. `PlaneExtension` inherits from both `MultiContainerInterface` and from `PlaneSegmentation`, which is a `DynamicTable`. 

Within `DynamicTable` is a pre-init method `_gather_columns` that uses the base classes in a conditional to create pre-defined columns. When autogenerating a class, the order of these classes are determined in lines 377-385 of classgenerator.py
```
if issubclass(MultiContainerInterface, bases[0]):
    # if bases[0] is Container or another superclass of MCI, then make sure MCI goes first
    # otherwise, MRO is ambiguous
    bases.insert(0, MultiContainerInterface)
else:
    # bases[0] is not a subclass of MCI and not a superclass of MCI. place that class first
    # before MCI. that class __init__ should call super().__init__ which will call the
    # MCI init
    bases.insert(1, MultiContainerInterface)
```
Since `MultiContainerInterface` is not a sub-class of `DynamicTable` (bases[0]), we insert at index 1. 

Within `gather_columns`, we check
```
if (len(bases) and 'DynamicTable' in globals() and issubclass(bases[-1], Container)
                and bases[-1].__columns__ is not cls.__columns__):
```
Such that bases[-1] (the last base in this case would be `MultiContainerInterface`) is assumed to be a `DynamicTable` and checks for columns. This returns the error: 
`AttributeError: type object 'MultiContainerInterface' has no attribute '__columns__'`

To fix this, the search for DynamicTable should be agnostic to the order of the bases. 

### Steps to Reproduce

```shell
# Run the extension or un-register a class.
```

### Operating System

macOS

### Python Executable

Python

### Python Version

3.10

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf/blob/dev/.github/CODE_OF_CONDUCT.md)
[...]

PR #991, opened 2023-10-31
_gather_columns agnostic search

mavaylon1:
## Motivation

Within DynamicTable is a pre-init method _gather_columns that uses the base classes in a conditional to create pre-defined columns. When autogenerating a class, the order of these classes are determined in classgenerator.py
```
if '__clsconf__' in classdict:
            # do not add MCI as a base if a base is already a subclass of MultiContainerInterface
            for b in bases:
                if issubclass(b, MultiContainerInterface):
                    break
            else:
                if issubclass(MultiContainerInterface, bases[0]):
                    # if bases[0] is Container or another superclass of MCI, then make sure MCI goes first
                    # otherwise, MRO is ambiguous
                    bases.insert(0, MultiContainerInterface)
                else:
                    # bases[0] is not a subclass of MCI and not a superclass of MCI. place that class first
                    # before MCI. that class __init__ should call super().__init__ which will call the
                    # MCI init
                    bases.insert(1, MultiContainerInterface)
```

This results in an issue for an extension that use the nwbinspector to validate the NWBFile. The best fix is to make the search in _gather_columns agnostic to the order.

## How to test the behavior?
```
```
## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -268,7 +268,7 @@ class DynamicTable(Container):
 
     @ExtenderMeta.pre_init
     def __gather_columns(cls, name, bases, classdict):
-        """
+        r"""
         Gather columns from the *\_\_columns\_\_* class attribute and add them to the class.
 
         This classmethod will be called during class declaration in the metaclass to automatically
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -268,7 +268,7 @@ class DynamicTable(Container):
 
     @ExtenderMeta.pre_init
     def __gather_columns(cls, name, bases, classdict):
-        r"""
+        """
         Gather columns from the *\_\_columns\_\_* class attribute and add them to the class.
 
         This classmethod will be called during class declaration in the metaclass to automatically
@@ -281,14 +281,11 @@ class DynamicTable(Container):
         if len(bases) and 'DynamicTable' in globals():
             for item in bases[::-1]: # reverse the bases tuple as the code suggest it should be last
                 if issubclass(item, Container):
-                    try:
-                        if item.__columns__ is not cls.__columns__:
-                            new_columns = list(cls.__columns__)
-                            new_columns[0:0] = item.__columns__  # prepend superclass columns to new_columns
-                            cls.__columns__ = tuple(new_columns)
-                            break
-                    except AttributeError:  # raises error when "__columns__" is not an attr of item
-                        continue
+                    if item.__columns__ is not cls.__columns__:
+                        new_columns = list(cls.__columns__)
+                        new_columns[0:0] = item.__columns__  # prepend superclass columns to new_columns
+                        cls.__columns__ = tuple(new_columns)
+                        break
 
     @docval({'name': 'name', 'type': str, 'doc': 'the name of this table'},  # noqa: C901
             {'name': 'description', 'type': str, 'doc': 'a description of what is in this table'},
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -286,7 +286,7 @@ class DynamicTable(Container):
                             new_columns = list(cls.__columns__)
                             new_columns[0:0] = item.__columns__  # prepend superclass columns to new_columns
                             cls.__columns__ = tuple(new_columns)
-                    except AttributeError:
+                    except AttributeError:  # raises error when "__columns__" is not an attr of item
                         continue
 
     @docval({'name': 'name', 'type': str, 'doc': 'the name of this table'},  # noqa: C901
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -281,11 +281,14 @@ class DynamicTable(Container):
         if len(bases) and 'DynamicTable' in globals():
             for item in bases[::-1]: # reverse the bases tuple as the code suggest it should be last
                 if issubclass(item, Container):
-                    if item.__columns__ is not cls.__columns__:
-                        new_columns = list(cls.__columns__)
-                        new_columns[0:0] = item.__columns__  # prepend superclass columns to new_columns
-                        cls.__columns__ = tuple(new_columns)
-                        break
+                    try:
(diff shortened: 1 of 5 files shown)


mavaylon1:
Fix #959

mavaylon1:
@rly Proof of concept. This passes the "test" of the extension now works.

src/hdmf/common/table.py
```diff
@@ -278,11 +278,16 @@ def __gather_columns(cls, name, bases, classdict):
...
-            new_columns = list(cls.__columns__)
-            new_columns[0:0] = bases[-1].__columns__  # prepend superclass columns to new_columns
-            cls.__columns__ = tuple(new_columns)
+        if len(bases) and 'DynamicTable' in globals():
+            for item in bases:
+                if issubclass(item, Container):
+                    try:
+                        if item.__columns__ is not cls.__columns__:
+                            new_columns = list(cls.__columns__)
+                            new_columns[0:0] = bases[-1].__columns__  # prepend superclass columns to new_columns
```
```suggestion
                            new_columns[0:0] = item.__columns__  # prepend superclass columns to new_columns
```

src/hdmf/common/table.py
```diff
@@ -278,11 +278,16 @@ def __gather_columns(cls, name, bases, classdict):
...
+                            cls.__columns__ = tuple(new_columns)
+                    except AttributeError:
```
```suggestion
                    except AttributeError:  # raises error when "__columns__" is not an attr of item
```

rly:
I think the for loop should break upon finding a class that has `__columns__`. Otherwise the same columns might be added multiple times. I also think the order should be last first. The current code suggests that we should look only at the last base class. Can you create a test where extension Class A extends DynamicTable and has `__columns__`, extension Class B extends Class A and has `__columns__`, and extension Class C extends Class B. The generated `__columns__` of Class C should prepend only the columns defined in Class B and not touch the columns defined in Class A (which should be in Class B). 

This also raises an issue where a column in Class B might overwrite a column in Class A. We don't want to add the same column twice during the prepending. We should add logic to check for that in this block of code, but perhaps in a separate PR.

mavaylon1:
> I think the for loop should break upon finding a class that has `__columns__`. Otherwise the same columns might be added multiple times. I also think the order should be last first. The current code suggests that we should look only at the last base class. Can you create a test where extension Class A extends DynamicTable and has `__columns__`, extension Class B extends Class A and has `__columns__`, and extension Class C extends Class B. The generated `__columns__` of Class C should prepend only the columns defined in Class B and not touch the columns defined in Class A (which should be in Class B). 
> 
> 
> 
> This also raises an issue where a column in Class B might overwrite a column in Class A. We don't want to add the same column twice during the prepending. We should add logic to check for that in this block of code, but perhaps in a separate PR. 

I'll give it a gander. This was just to get our initial idea out in practice.

rly:
Understood and sounds good.

mavaylon1:
@rly Taking a gander and wanted to confirm some thoughts. PyNWB has PlaneSegmentation and that is a subclass of DynamicTable. The extension has PlaneExtension, which extends PlaneSegmentation. When writing a NWBFile from the extension, the class defined in the extension is used. However, when running the nwbinspector, it does not see the class but just the schema (not clear how). This leads hdmf to autogenerate the class. 

In a testing case, Class A is PlaneSegmentation, Class B is PlaneExtension, and Class C is an extension of Class B with generated columns. You say "prepend" meaning I should use add_column to Class C and verify that it doesn't affect the columns in class A?

By not defining the __columns__ in class C, that should trigger them to be autogenerated to match B correct? But that would require a schema

rly:
> You say "prepend" meaning I should use add_column to Class C and verify that it doesn't affect the columns in class A?

By prepend, I am referring to line 287 in the new code. I mean that the gather_columns method should add the columns in class A once not twice because those columns were also added to class B. 

> By not defining the **columns** in class C, that should trigger them to be autogenerated to match B correct? But that would require a schema

Yes, if class C does not add or change any columns, then the `C.__columns__` should match `B.__columns__`

mavaylon1:
> > You say "prepend" meaning I should use add_column to Class C and verify that it doesn't affect the columns in class A?
> 
> By prepend, I am referring to line 287 in the new code. I mean that the gather_columns method should add the columns in class A once not twice because those columns were also added to class B.
> 
> > By not defining the **columns** in class C, that should trigger them to be autogenerated to match B correct? But that would require a schema
> 
> Yes, if class C does not add or change any columns, then the `C.__columns__` should match `B.__columns__`

Do you mean add the columns "from" class A to class B?

rly:
> Do you mean add the columns "from" class A to class B?

Gather columns should copy the columns from class A to class B, and from class B to class C, but because class B contains class A's columns, gather_columns should not additionally copy columns from class A to class C, or else class A's columns will be in class C twice.

mavaylon1:
> > Do you mean add the columns "from" class A to class B?
> 
> Gather columns should copy the columns from class A to class B, and from class B to class C, but because class B contains class A's columns, gather_columns should not additionally copy columns from class A to class C, or else class A's columns will be in class C twice.

Class A's columns would be in Class C via Class B.

src/hdmf/common/table.py
```diff
@@ -278,11 +278,17 @@ def __gather_columns(cls, name, bases, classdict):
             msg = "'__columns__' must be of type tuple, found %s" % type(cls.__columns__)
             raise TypeError(msg)
 
-        if (len(bases) and 'DynamicTable' in globals() and issubclass(bases[-1], Container)
-                and bases[-1].__columns__ is not cls.__columns__):
-            new_columns = list(cls.__columns__)
-            new_columns[0:0] = bases[-1].__columns__  # prepend superclass columns to new_columns
-            cls.__columns__ = tuple(new_columns)
+        if len(bases) and 'DynamicTable' in globals():
+            for item in bases[::-1]: # reverse the bases tuple as the code suggest it should be last
```
```suggestion
            for item in bases[::-1]: # look for __columns__ in the base classes, closest first
```

rly (approved):

rly:
Looks good to me. Thanks!

PR #996, opened 2023-11-08
propose alternative chunk shape algorithm

### LINKED ISSUE #995: [Feature]: Define partial chunk shape for GenericDataChunkIterator

Right now, for the `GenericDataChunkIterator`, it's possible to define `chunk_mb` or `chunk_shape`. I would like to enable a hybrid approach, where a user could input `chunk_mb=10.0, chunk_shape=(None, 64)`, and the `GenericDataChunkIterator` would identify the remaining dimension that gets you close to the target chunk size.

It is pretty common for users to have some insight into the likely read patterns of a dataset.

I would like `GenericDataChunkIterator` to find the maximum size (prod of dims) that is <= the target size. I also would like the chunk to be as cube-like as possible, so I would like to minimize the sum of the dimensions of the array. Previously, we tried building chunks that were scaled down versions of the data shape, similar to h5py, but experience with Jeremy has shown that this approach is poorly suited for common data reading routines, and I think a better naive assumption would be that (hyper-) cube chunks are a good default.

Yes.

### Code of Conduct

bendichter:
## Motivation

Fix #995 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -3,13 +3,99 @@ import math
 from abc import ABCMeta, abstractmethod
 from collections.abc import Iterable
 from warnings import warn
-from typing import Tuple, Callable
+from typing import Tuple, Callable, List, Optional, Union
 from itertools import product, chain
 
 import h5py
 import numpy as np
 
 from .utils import docval, getargs, popargs, docval_macro, get_data_shape
+
+
+def find_nth_none(lst, n):
+    """
+    Finds the index of the nth None in a list.
+
+    Parameters
+    ----------
+    lst : list
+        The list in which to search for None values.
+    n : int
+        The occurrence of None to find (1-based).
+
+    Returns
+    -------
+    int or None
+        The index of the nth None in the list, or None if there aren't enough None values.
+    """
+    count = 0
+    for index, value in enumerate(lst):
+        if value is None:
+            count += 1
+            if count == n:
+                return index
+    return None
+
+
+def array_with_desired_product(
+    desired_product: int,
+    upper_limits: Iterable[Optional[int]],
+    defined_elements: List[Optional[int]],
+) -> List[int]:
+    """
+    Adjusts the undefined elements in a list such that their product equals the desired product,
+    respecting the upper limits for each element.
+
+    Parameters
+    ----------
+    desired_product : int
+        The desired product of all elements in the list.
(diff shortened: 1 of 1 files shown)


src/hdmf/data_utils.py
```diff
@@ -12,6 +12,92 @@
...
+    count = 0
+    for index, value in enumerate(lst):
+        if value is None:
+            count += 1
+            if count == n:
+                return index
+    return None
+
+
+def array_with_desired_product(
```
oruebel, line 40: Since the intent is for this to be used to determine chunk shape, I think naming the function in a way that describes what it is used for may be more intuitive for users.

src/hdmf/data_utils.py
```diff
@@ -12,6 +12,92 @@
 from .utils import docval, getargs, popargs, docval_macro, get_data_shape
 
 
+def find_nth_none(lst, n):
```
oruebel, line 15: Is this function intended for external use? If we will only use this for ``array_with_desired_product`` method, then I would suggest to make the method private.
```suggestion
def __find_nth_none(lst, n):
```

oruebel:
In general, I think this is reasonable. @CodyCBakerPhD can you also take a look.

bendichter:
Another approach could be to work with the prime factorization of the residual product of the size. That would get us closer to the desired chunk size but we could end up with very uneven shapes

oruebel:
> Another approach could be to work with the prime factorization of the residual product of the size. That would get us closer to the desired chunk size but we could end up with very uneven shapes

Uneven shape is not necessarily bad, but I think you probably want the ratio to be informed by the overall size of the dataset. I.e., for ``(time,  electrodes)`` where time is much longer than electrodes, then you'd also want the chunks to be larger in the time dimension.

src/hdmf/data_utils.py
bendichter: You're right, both of these should be private, though @CodyCBakerPhD 's solution makes this function unnecessary.

src/hdmf/data_utils.py
bendichter: I actually prefer it this way: separating the math from the application. What we want here is a vector that has a product that is near to a target and has minimal sum, given constraints on the vector. This function is then called in an effort to determine the shape of a chunk.

PR #997, opened 2023-11-08
Constrained chunk shape estimation

### LINKED ISSUE #995: [Feature]: Define partial chunk shape for GenericDataChunkIterator

CodyCBakerPhD:
## Motivation

fix #995

Potential replacement of #996 

This is actually really good timing; the latest stage of Zarr integration in NeuroConv (https://github.com/catalystneuro/neuroconv/pull/569) benefits greatly from having these types of methods become static methods of the iterator (and its children) for broader usage, much as @oruebel suggests in https://github.com/hdmf-dev/hdmf/issues/995#issuecomment-1802509057

@bendichter Anyway, I did some mathematical reduction and refactoring of the suggested approach from #996; it's actually quite similar to the original axis ratio method in that one could imagine an altogether more general algorithm that places 'weight factors' on each axis to determine how strongly to bias the distribution of bytes throughout the process...

In the recently suggested update these weight factors would all be `1`; in the original, they are something proportional to the relative length of each axis

This PR also addresses @oruebel's comments by no longer requiring secondary methods

Two other changes:

I strongly suggest keeping the previous method accessible 'just in case' something unexpected arises in the new implementation; a lesson from learned from previous bugs in this iterator. The other fallback would be to force installation of a previous HDMF version, but that could be less desirable and cause other problems (such as incompatibilities) on top of that

Another lesson learned from the past, always use `math.prod` for these operations. Eventually these methods will be used on shapes with very large values, and `np.prod` overflows easily and silently (well, occasionally there might be a warning thrown but not always)

## How to test the behavior?

Existing tests may require updating to the newly predicted default values

A few new tests should be added for the static usage and showcase of the expected results from each strategy

## Checklist

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -182,6 +182,12 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             doc="Dictionary of keyword arguments to be passed directly to tqdm.",
             default=None,
         ),
+        dict(
+            name="chunking_strategy",
+            type=str,
+            doc="One of ['axis_ratio', 'uniform'].",
+            default=None,
+        ),
     )
 
     @docval(*__docval_init)
@@ -196,8 +202,23 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
         HDF5 recommends chunk size in the range of 2 to 16 MB for optimal cloud performance.
         https://youtu.be/rcS5vt-mKok?t=621
         """
-        buffer_gb, buffer_shape, chunk_mb, chunk_shape, self.display_progress, progress_bar_options = getargs(
-            "buffer_gb", "buffer_shape", "chunk_mb", "chunk_shape", "display_progress", "progress_bar_options", kwargs
+        (
+            buffer_gb,
+            buffer_shape,
+            chunk_mb,
+            chunk_shape,
+            self.display_progress,
+            progress_bar_options,
+            chunking_strategy,
+        ) = getargs(
+            "buffer_gb",
+            "buffer_shape",
+            "chunk_mb",
+            "chunk_shape",
+            "display_progress",
+            "progress_bar_options",
+            "chunking_strategy",
+            kwargs
         )
         self.progress_bar_options = progress_bar_options or dict()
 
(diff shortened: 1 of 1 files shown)


src/hdmf/data_utils.py
```diff
@@ -286,69 +314,140 @@ def __init__(self, **kwargs):
...
+                constrained_bytes = math.prod(estimated_chunk_shape[constrained_axes]) * itemsize
+                estimated_fill_factor = math.floor((chunk_bytes / constrained_bytes) ** (1 / number_of_free_axes))
+
+                # Update axes and constraints in the event that the fill factor pushed some axes beyond their maximum
+                capped_axes = none_axes[np.where(maxshape[none_axes] <= estimated_fill_factor)[0]]
+                estimated_chunk_shape[capped_axes] = maxshape[capped_axes]  # Cap the estimate at the max
+                chunk_shape_constraints[capped_axes] = maxshape[capped_axes]  # Consider capped axis a constraint
+                none_axes = np.where([x is None for x in chunk_shape_constraints])[0]
+                constrained_axes = np.where([x is not None for x in chunk_shape_constraints])[0]
+            estimated_chunk_shape[none_axes] = estimated_fill_factor
```
bendichter, line 387: OK so if I understand it's:

1. try to set unconstrained axes to the nth root where n in the number of unconstrained
2. for any axes where this exceeds the limit, constrain it to that limit
3. repeat

bendichter, line 387: OK, I think this seems like a good alternative solution, which should achieve the same answer in far fewer loops, but I agree with @oruebel that it could go in a separate function. I liked the way I broke it apart where the math was in a separate function that was just trying to achieve a vector with a certain product (not knowing that it was related to the size and shape of an array).

Also, a minor point: I would remove the `np.where`s. If you are going this way you can just index with the boolean np arrays, which will make the syntax a bit cleaner.

src/hdmf/data_utils.py
```diff
@@ -286,69 +314,140 @@ def __init__(self, **kwargs):
...
-            ) ** (1 / len(self.chunk_shape))
-        )
-        return tuple(
-            [
-                min(max(k * x, self.chunk_shape[j]), self.maxshape[j])
-                for j, x in enumerate(self.chunk_shape)
-            ]
+            (buffer_gb * 1e9 / (math.prod(chunk_shape) * itemsize)) ** (1 / len(chunk_shape))
         )
+        return tuple([min(max(k * x, chunk_shape[j]), maxshape[j]) for j, x in enumerate(chunk_shape)])
```
bendichter, line 450: Nice, pulling these attributes out made the code much more readable.

bendichter:
> I think you probably want the ratio to be informed by the overall size of the dataset. I.e., for (time,  electrodes) where time is much longer than electrodes, then you'd also want the chunks to be larger in the time dimension. (@oruebel https://github.com/hdmf-dev/hdmf/pull/996#issuecomment-1802574509)

@oruebel, this was our original guess, and it was conveniently also the approach h5py uses, but our experience with @magland has shown that this can cause big problems for certain common applications and at this point I am leaning towards the hypercube approach, as I don't expect it to totally blow up in most cases. The (time, electrode) is the perfect example- it's actually much more common for users to want many channels over a short time than one channel over a long time, particularly in streaming applications.

> This PR also addresses @oruebel's comments by no longer requiring secondary methods

See the way I separated the function out in my PR. I think this the function as-is is a bit too long and should be broken up, but that's more of a style thing.

I'm fine with keeping the old method for now. Let's see how often we or anyone else tends to use it.

And yes, using `math.prod` is a good idea. I do remember that being a particularly sneaky bug we definitely do not want to reintroduce.

PR #998, opened 2023-11-09
improve html display for tables

bendichter:
* improve the title code
* render tables correctly

I have not yet updated the tests

## Motivation

## How to test the behavior?
```
from pynwb.testing.mock.ecephys import mock_ElectricalSeries
from pynwb.testing.mock.ophys import mock_ImageSegmentation
ex1 = mock_ElectricalSeries()
ex2 = mock_ImageSegmentation()
```

image

image

## Checklist


src/hdmf/common/table.py
```diff
@@ -1192,6 +1192,17 @@ def to_dataframe(self, **kwargs):
...
         return ret
 
+    def generate_html_repr(self, level=0, access_code=""):
+        out = ""
+        for key, value in self.fields.items():
+            if key not in ("id", "colnames", "columns"):
+                out += self._generate_field_html(key, value, level, access_code)
+        out += (
+            f'<details><summary style="display: list-item; margin-left: {level * 20}px;" '
+            f'class="container-fields field-key" title="{access_code}"><b>table</b></summary>{self.to_dataframe().to_html()}</details>'
```
oruebel: My only concern here is that this will load the whole data into memory, and as such, rendering the tables may be slow, in particular if a file is opened remotely (e.g., on S3) or when dealing with large tables. I'm wondering whether we can have some form of configuration option to define: a) whether tables should be rendered, and b) when rendering tables, how many rows should be loaded.

rly: What if we do the following: If there are more than 4 rows, we slice the table to be the first four rows (returns a dataframe) and add text at the bottom of the repr saying how many rows there are? 

```
{self[0:min(4, len(self))].to_dataframe().to_html()}<br/>
{len(self)} rows x {len(self.columns)} columns
```

bendichter: Good point, @oruebel . @rly, this sounds like a good solution!

bendichter:
@rly , @oruebel ok I did the restriction to the first 4 rows for each table.

image

image

rly:
Looks good to me!

oruebel:
Looks good to me too.

rly (approved):

rly:
Looks good. There are some lines of code that we could cover with tests so we're not reducing our code coverage, but I think we can defer that to a separate PR so that this one can make it to users sooner.

PR #999, opened 2023-11-15
Update get_key to support multiple keys

mavaylon1:
## Motivation

This is the continuation of PR #964. Update the `get_key` method to return all the keys if there are multiple (instead of returning an error. 

## How to test the behavior?
```
```

## Checklist


src/hdmf/common/resources.py
```diff
@@ -510,8 +510,7 @@ def get_key(self, **kwargs):
                 # the key has never been used before
                 raise ValueError("key '%s' does not exist" % key_name)
             elif len(key_idx_matches) > 1:
-                msg = "There are more than one key with that name. Please search with additional information."
-                raise ValueError(msg)
+                return [self.keys.row[x] for x in key_idx_matches]
```
rly, line 515: Please add to the docstring something that says something like: if multiple matching Keys are found, a list of Keys is returned. else, a single Key is returned.

rly (approved):

PR #1013, opened 2023-12-10
Add set_data_io to Data

### LINKED ISSUE #1012: [Feature]: Deprecate Data.set_dataio and replace with Container.set_data_io logic

`Data.set_dataio`: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/container.py#L783-L790

`Container.set_data_io`: https://github.com/hdmf-dev/hdmf/blob/f95b1efa0e890d35dd69d74d496ec6f2fa1e716d/src/hdmf/container.py#L749-L753

These are a bit inconsistent in their usage

I have a mild preference the previous `Data` method be soft deprecated in favor of the `Container` approach (passing both the class and the args to be passed to that class)

CodyCBakerPhD:
fix #1012

## Motivation

Outlined in issue

## How to test the behavior?

- [x] add test
- [x] add deprecation warning

## Checklist

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -791,7 +791,7 @@ class Data(AbstractContainer):
         self.__data = dataio
 
     @docval(
-        {'name': 'data_io_class', 'type': Type[DataIO], 'doc': 'The DataIO to apply to the data held by this Data.'}
+        {'name': 'data_io_class', 'type': Type[DataIO], 'doc': 'The DataIO to apply to the data held by this Data.'},
         {'name': 'data_io_kwargs', 'type': dict, 'doc': 'The keyword arguments to pass to the DataIO.'}
     )
     def set_data_io(self, **kwargs):
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -747,11 +747,11 @@ class Container(AbstractContainer):
         out += '\n' + indent + right_br
         return out
 
-    def set_data_io(self, dataset_name, data_io_class, **kwargs):
+    def set_data_io(self, dataset_name: str, data_io_class: Type[DataIO], data_io_kwargs: dict):
         data = self.fields.get(dataset_name)
         if data is None:
             raise ValueError(f"{dataset_name} is None and cannot be wrapped in a DataIO class")
-        self.fields[dataset_name] = data_io_class(data=data, **kwargs)
+        self.fields[dataset_name] = data_io_class(data=data, **data_io_kwargs)
 
 
 class Data(AbstractContainer):
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -4,6 +4,7 @@ from collections import OrderedDict
 from copy import deepcopy
 from uuid import uuid4
 from warnings import warn
+from typing import Type
 
 import h5py
 import numpy as np
@@ -788,6 +789,15 @@ class Data(AbstractContainer):
         dataio = getargs('dataio', kwargs)
         dataio.data = self.__data
         self.__data = dataio
+
+    @docval(
+        {'name': 'data_io_class', 'type': Type[DataIO], 'doc': 'The DataIO to apply to the data held by this Data.'}
+        {'name': 'data_io_kwargs', 'type': dict, 'doc': 'The keyword arguments to pass to the DataIO.'}
+    )
+    def set_data_io(self, **kwargs):
+        """Apply DataIO object to the data held by this Data object."""
+        data_io_class, data_io_kwargs = getargs('data_io_class', 'data_io_kwargs', kwargs)
+        self.__data = data_io_class(data=self.__data, **data_io_kwargs)
 
     @docval({'name': 'func', 'type': types.FunctionType, 'doc': 'a function to transform *data*'})
     def transform(self, **kwargs):
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -757,7 +757,7 @@ class Container(AbstractContainer):
             Name of dataset to wrap in DataIO
         data_io_class: Type[DataIO]
             Class to use for DataIO, e.g. H5DataIO or ZarrDataIO
-        data_io_kwargs: dict, optional
+        data_io_kwargs: dict
             keyword arguments passed to the constructor of the DataIO class.
         **kwargs:
             DEPRECATED. Use data_io_kwargs instead.
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -809,6 +809,11 @@ class Data(AbstractContainer):
         """
         Apply DataIO object to the data held by this Data object
         """
+        warn(
+            "Data.set_dataio() is deprecated. Please use Data.set_data_io() instead.",
+            DeprecationWarning,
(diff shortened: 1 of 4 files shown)


src/hdmf/container.py
```diff
@@ -746,11 +747,11 @@ def __smart_str_dict(d, num_indent):
         out += '\n' + indent + right_br
         return out
 
-    def set_data_io(self, dataset_name, data_io_class, **kwargs):
+    def set_data_io(self, dataset_name: str, data_io_class: Type[DataIO], data_io_kwargs: dict):
```
CodyCBakerPhD: This change is also inspired for consistency; usually the `**kwargs` usage across HDMF refers to the ones that get popped and passed on by docval

bendichter: a bit of a backwards compatibility break though. Could you do a a soft deprecation?

bendichter: yes, I see how this could be confusing. Can we try to do a soft deprecation, since this technically breaks backwards compatibility? Maybe something like 

```
def set_data_io(self, dataset_name: str, data_io_class: Type[DataIO], data_io_kwargs: dict, **kwargs):
    if kwargs:
        DeprecationWarning("Using of kwargs is deprecated. Please pass the kwargs for the DataIO class in `data_io_kwargs` as a dictionary instead.")
        data_io_kwargs = kwargs
    ...
```

bendichter:
Thanks for working on this, @CodyCBakerPhD. I think this feature is really important. After this is integrated, I would like to also add a convenience method on `DynamicTable`. Something like `DynamicTable.make_expandable`, but let's save that for a separate PR after this one is merged.

bendichter:
I'll take a stab and finishing this

src/hdmf/container.py
```diff
@@ -747,22 +747,34 @@ def __smart_str_dict(d, num_indent):
...
+        Apply DataIO object to a dataset field of the Container.
 
         Parameters
         ----------
         dataset_name: str
             Name of dataset to wrap in DataIO
         data_io_class: Type[DataIO]
             Class to use for DataIO, e.g. H5DataIO or ZarrDataIO
+        data_io_kwargs: dict, optional
+            kwargs passed to the constructor of the DataIO class.
```
bendichter:
```suggestion
            keyword arguments passed to the constructor of the DataIO class.
```

src/hdmf/container.py
```diff
@@ -747,22 +747,34 @@ def __smart_str_dict(d, num_indent):
...
         data_io_class: Type[DataIO]
             Class to use for DataIO, e.g. H5DataIO or ZarrDataIO
+        data_io_kwargs: dict, optional
+            keyword arguments passed to the constructor of the DataIO class.
         **kwargs:
-            kwargs are passed to the constructor of the DataIO class
+            DEPRECATED. Use data_io_kwargs instead.
+            kwargs are passed to the constructor of the DataIO class.
         """
+        if kwargs or (data_io_kwargs is None):
```
rly, line 766: The warning triggers when `data_io_kwargs` is not provided, which I think is not intended.
However, I can't imagine why someone would wrap a dataset with a `DataIO` without any kwargs. So I suggest that `data_io_kwargs` is required.

rly: However, that would make a hard breaking change, because `obj.set_data_io("data2", H5DataIO, chunks=True)` would raise an error, so never mind.

src/hdmf/container.py
```diff
@@ -747,22 +747,34 @@ def __smart_str_dict(d, num_indent):
...
         """
+        Apply DataIO object to a dataset field of the Container.
 
         Parameters
         ----------
         dataset_name: str
             Name of dataset to wrap in DataIO
         data_io_class: Type[DataIO]
             Class to use for DataIO, e.g. H5DataIO or ZarrDataIO
+        data_io_kwargs: dict, optional
```
bendichter:
```suggestion
        data_io_kwargs: dict
```

@rly this should help a bit

src/hdmf/container.py
bendichter: Yes, that's the dilemma.

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -3671,3 +3671,54 @@ def test_hdf5io_can_read():
...
+
+    def test_set_data_io_old_api(self):
+        """Test that using the kwargs still works but throws a warning."""
+        msg = (
+            "Use of **kwargs in Container.set_data_io() is deprecated. Please pass the DataIO kwargs as a dictionary to"
+            " the `data_io_kwargs` parameter instead."
+        )
+        with self.assertWarnsWith(DeprecationWarning, msg):
+            self.obj.set_data_io("data1", H5DataIO, chunks=True)
+        self.assertIsInstance(self.obj.data1, H5DataIO)
```
rly, line 3712:
```suggestion
        self.assertIsInstance(self.obj.data1, H5DataIO)
        self.assertTrue(self.obj.data1.chunks)
```

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -3671,3 +3671,55 @@ def test_hdf5io_can_read():
...
+            def __init__(self, **kwargs):
+                super().__init__(name=kwargs["name"])
+                self.data1 = kwargs["data1"]
+                self.data2 = kwargs["data2"]
+
+        self.obj = ContainerWithData("name", [1, 2, 3, 4, 5], None)
+
+    def test_set_data_io(self):
+        self.obj.set_data_io("data1", H5DataIO, data_io_kwargs=dict(chunks=True))
+        assert isinstance(self.obj.data1, H5DataIO)
```
rly, line 3696:
```suggestion
        assert isinstance(self.obj.data1, H5DataIO)
        assert self.obj.data1.chunks
```

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -3671,3 +3671,56 @@ def test_hdf5io_can_read():
...
+
+    def setUp(self):
+        class MyData(Data):
+            pass
+
+        self.data = MyData("my_data", [1, 2, 3])
+
+    def test_set_data_io(self):
+        self.data.set_data_io(H5DataIO, dict(chunks=True))
+        assert isinstance(self.data.data, H5DataIO)
```
rly, line 3726:
```suggestion
        assert isinstance(self.data.data, H5DataIO)
        assert self.data.data.chunks
```

rly (dismissed):

src/hdmf/container.py
```diff
@@ -801,6 +813,19 @@ def set_dataio(self, **kwargs):
         dataio.data = self.__data
         self.__data = dataio
 
+    def set_data_io(self, data_io_class: Type[DataIO], data_io_kwargs: dict) -> None:
```
rly, line 821: Can you add a deprecation warning to `Data.set_dataio` and tell people to use `Data.set_data_io` instead?

bendichter: done

bendichter (approved):

PR #1016, opened 2023-12-11
Configuration File for TermSet validations

mavaylon1:
## Motivation

TLDR: Create a configuration file that sets which datatypes are associated with default NWB created termsets. This will be used for validation fields.

Note: This wrapping is done in the setter, requiring the field to be defined in the schema. If not, user the wrapper manually. 
TODO:

- [x] Rename TermSetConfigurator to TypeConfigurator
- [x] Set the config file to use keys
- [x] use ruamel and not yaml
- [x] deep copy fix
- [ ] doc tutorial

## How to test the behavior?
```
```

## Checklist


mavaylon1:
@oruebel Give a gander to the workflow above to see where my ideas are at.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## HDMF 3.13.0 (Upcoming)
 
 ### Enhancements
+- Added `TermSetConfigurator` to automatically wrap fields with `TermSetWrapper` according to a configuration file. @mavaylon1 [1016](https://github.com/hdmf-dev/hdmf/pull/1016)
```
```suggestion
- Added `TermSetConfigurator` to automatically wrap fields with `TermSetWrapper` according to a configuration file. @mavaylon1 [#1016](https://github.com/hdmf-dev/hdmf/pull/1016)
```

src/hdmf/common/__init__.py
```diff
@@ -20,6 +20,34 @@
...
+        is_method=False)
+def load_termset_config(**kwargs):
+    """
+    This method will either load the default config or the config provided by the path.
+    """
+    config_path, return_map = kwargs['config_path'], kwargs['return_map']
+    __TYPE_MAP.ts_config.load_termset_config(config_path)
+
+    if return_map: # This helps with testing
+        return __TYPE_MAP
```
rly: I think this is not needed. You can call `get_type_map` to get a copy of the global `__TYPE_MAP`. It's potentially dangerous to give the user direct access to the global `__TYPE_MAP`. Hence, all the functions below that allow a user to do X to `__TYPE_MAP` instead of just making `__TYPE_MAP` public.

src/hdmf/common/__init__.py
```diff
@@ -37,10 +65,14 @@ def register_class(**kwargs):
...
             _set_exp(cls)
             __TYPE_MAP.register_container_type(namespace, data_type, cls)
+            cls.type_map = __TYPE_MAP
+            cls.namespace = namespace
             return cls
     else:
         def _dec(cls):
             __TYPE_MAP.register_container_type(namespace, data_type, cls)
+            cls.type_map = __TYPE_MAP
+            cls.namespace = namespace
```
rly: I feel uneasy making these public variables that a user can manipulate. Also it seems like this should happen within `register_container_type` instead of having it here and duplicated above.

src/hdmf/common/__init__.py
mavaylon1: So the copy of the type map does not actually have the same configuration loaded. This is why I went with that. Theoretically dangerous, but practically I doubt the users know what to do with it. That is also why the default is false.

src/hdmf/common/__init__.py
mavaylon1: For efficiency yes they could be there. However, having them here gives us:
1. Easy debugging since everything is there.
2. (More importantly) it uses __TYPe_MAP and not a copy. The passing around of copies has proven to be an obstacle for me in that the configuration is not preserved.

mavaylon1: I do agree I could make a way to avoid the duplication.

rly: Please read the answers here on why one should make variables private/protected: https://softwareengineering.stackexchange.com/questions/143736/why-do-we-need-private-variables

> (More importantly) it uses __TYPe_MAP and not a copy. The passing around of copies has proven to be an obstacle for me in that the configuration is not preserved.

This seems like a problem that can be solved in a different way than making a variable that is very intentionally private suddenly very public. When a type map is copied, can the term set config be copied as well?

src/hdmf/container.py
```diff
@@ -85,10 +86,78 @@ def setter(self, val):
...
+
+    def _field_config(self, arg_name, val):
+        """
+        This method will be called in the setter. The termset configuration will be used (if loaded)
+        to check for a defined TermSet associated with the field. If found, the value of the field
+        will be wrapped with a TermSetWrapper.
+
+        Even though the path field in the configurator can be a list of paths, the config
+        itself is only one file. When a user loads custom configs, the config is appended/modified.
+        The modificiations are not written to file, avoiding permanent modifications.
```
rly:
```suggestion
        The modifications are not written to file, avoiding permanent modifications.
```

src/hdmf/container.py
```diff
@@ -85,10 +86,78 @@ def setter(self, val):
...
+
+        Even though the path field in the configurator can be a list of paths, the config
+        itself is only one file. When a user loads custom configs, the config is appended/modified.
+        The modifications are not written to file, avoiding permanent modifications.
+        """
+        # load termset configuration file from global Config
+        try:
+            configurator = self.type_map.ts_config
+            if len(configurator.path)>0:
+                # The type_map has a config always set; however, when toggled off, the config path path is empty.
```
rly:
```suggestion
                # The type_map has a config always set; however, when toggled off, the config path is empty.
```

src/hdmf/common/__init__.py
rly: `register_container_type` also already sets the `namespace` attribute here: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/build/manager.py#L736  (though admittedly 1) that is not very clear and 2) it should probably also not be publicly settable)

src/hdmf/container.py
```diff
@@ -85,10 +86,78 @@ def setter(self, val):
...
+
+        # check to see that the namespace for the container is in the config
+        if self.namespace not in self.type_map.container_types:
+            msg = "%s not found within loaded configuration." % self.namespace
+            warn(msg)
+            return val
+        else:
+            # check to see that the container type is in the config under the namespace
+            config_namespace = termset_config['namespaces'][self.namespace]
+            object_name = self.__class__.__name__
```
rly: Because of how `register_class` works, the class name is not guaranteed to be the same as the data type name. You can get the type name through `getattr(self, self._data_type_attr)`. It would probably be useful to put that into its own `get_data_type` method.

rly: Also I suggest changing the variable name from `object_name` to `data_type`. I would interpret the name of the object as the name of the container.

src/hdmf/container.py
```diff
@@ -85,10 +86,78 @@ def setter(self, val):
...
+                        termset_path = os.path.join(CUR_DIR, config_namespace['data_types'][object_name][constr_name])
+                        termset = TermSet(term_schema_path=termset_path)
+                        # If the val has been manually wrapped then skip checking the config for the attr
+                        if type(val) == TermSetWrapper:
+                            msg = "Field value already wrapped with TermSetWrapper."
+                            warn(msg)
+                            return val
+                        else:
+                            val = TermSetWrapper(value=val, termset=termset)
+                            return val
```
rly:
```suggestion
                        # If the val has been manually wrapped then skip checking the config for the attr
                        if type(val) == TermSetWrapper:
                            msg = "Field value already wrapped with TermSetWrapper."
                            warn(msg)
                            return val
                        else:
                            # From the spec, get the corresponding constructor name
                            constr_name = obj_mapper.get_const_arg(spec)
                            termset_path = os.path.join(CUR_DIR, config_namespace['data_types'][object_name][constr_name])
                            termset = TermSet(term_schema_path=termset_path)
                            val = TermSetWrapper(value=val, termset=termset)
                            return val
```

rly: Moved the if code up because it did not depend on the code above it.

src/hdmf/term_set.py
```diff
@@ -304,3 +297,83 @@ def extend(self, arg):
         else:
             msg = ('"%s" is not in the term set.' % ', '.join([str(item) for item in bad_data]))
             raise ValueError(msg)
+
+class TermSetConfigurator:
+    """
+    This class allows users to toggle on/off a global configuration for defined data types.
+    When toggled on, every instance of a configuration file supported data type will be validated
+    according the corresponding TermSet.
```
```suggestion
    according to the corresponding TermSet.
```

src/hdmf/term_set.py
```diff
@@ -304,3 +297,83 @@ def extend(self, arg):
...
+        if kwargs['path'] is None:
+            self.path = []
+        else:
+            self.path = [kwargs['path']]
+            self.load_termset_config(config_path=self.path[0])
+
+    @docval({'name': 'data_type', 'type': str,
+             'doc': 'The desired data type within the configuration file.'},
+            {'name': 'namespace', 'type': str,
+             'doc': 'The namespace for the data type. '})
```
```suggestion
             'doc': 'The namespace for the data type.'})
```

src/hdmf/term_set.py
```diff
@@ -304,3 +297,83 @@ def extend(self, arg):
...
+    @docval({'name': 'config_path', 'type': str, 'doc': 'Path to the configuration file.'})
+    def load_termset_config(self,config_path):
+        """
+        Load the configuration file for validation on the fields defined for the objects within the file.
+        """
+        try:
+            import yaml
+        except ImportError: # pragma: no cover
+            msg = "Install yaml."
+            raise ValueError(msg)
```
rly: `ruamel.yaml` is a dependency of hdmf. https://github.com/hdmf-dev/hdmf/blob/dev/pyproject.toml#L36

Please add `import ruamel.yaml as yaml` at the top of the file and use that yaml library instead.

src/hdmf/term_set.py
```diff
@@ -304,3 +297,83 @@ def extend(self, arg):
...
+                        if namespace not in self.config['namespaces']: # append namespace config if not present
+                            self.config['namespaces'][namespace] = termset_config['namespaces'][namespace]
+                        else: # check for any needed overrides within existing namespace configs
+                            for data_type in termset_config['namespaces'][namespace]['data_types']:
+                                if data_type in self.config['namespaces'][namespace]['data_types']:
+                                    replace_config = termset_config['namespaces'][namespace]['data_types'][data_type]
+                                    self.config['namespaces'][namespace]['data_types'][data_type] = replace_config
+                                else: # append to config
+                                    new_config = termset_config['namespaces'][namespace]['data_types'][data_type]
+                                    self.config['namespaces'][namespace]['data_types'][data_type] = new_config
```
rly, line 367: It looks like lines 365-366 and 368-369 do the same thing but with a different local variable name. Can we get rid of the `if` statement here and add a comment that says something like `# copy all configs from termset_config to self.config, overriding existing values`. If you really want to keep the if statement, then please add a comment that says that `# NOTE: these two branches effectively do the same thing, but are split for clarity.` because it looks odd to have

src/hdmf/container.py
```diff
@@ -85,10 +86,78 @@ def setter(self, val):
...
+                    # In the case of dealing with datasets directly or not defined in the spec.
+                    # (Data/VectorData/DynamicTable/etc)
+                    if spec is None:
+                        msg = "Spec not found for %s." % attr
+                        warn(msg)
+                        return val
+                    else:
+                        # From the spec, get the corresponding constructor name
+                        constr_name = obj_mapper.get_const_arg(spec)
+                        termset_path = os.path.join(CUR_DIR, config_namespace['data_types'][object_name][constr_name])
```
rly:
```suggestion
                        # From the spec, get the mapped attribute name
                        mapped_attribute_name = obj_mapper.get_attribute(spec)
                        termset_path = os.path.join(CUR_DIR, config_namespace['data_types'][object_name][mapped_attribute_name])
```

It looks odd to see the constructor argument name being used here. In 100% of custom object mappings that I am aware of, the mapped constructor arg and the mapped attribute name are the same, so it does not matter, but in the event of a discrepancy, it seems better to use attribute names in the config file than constructor arg names.

tests/unit/test_config.yaml
```diff
@@ -0,0 +1,15 @@
+namespaces:
```
rly: Are this file and the one below it used? It does not look like it. And they use relative paths on your machine which won't work for everyone.

tests/unit/test_term_set.py
```diff
@@ -215,3 +217,125 @@ def test_wrapper_extend_error(self):
...
+        tc = TermSetConfigurator(path=path)
+        with self.assertRaises(ValueError):
+            tc.load_termset_config(config_path=path)
+
+    def test_load_two_unique_configs(self):
+        path = 'tests/unit/hdmf_config.yaml'
+        path2 = 'tests/unit/hdmf_config2.yaml'
+        tc = TermSetConfigurator(path=path)
+        tc.load_termset_config(config_path=path2)
+        self.assertEqual(tc.path, [path, path2])
```
rly, line 275: Please add a test here to check that old values were replaced and new key-value pairs were added.

tests/unit/test_term_set.py
```diff
@@ -215,3 +217,125 @@ def test_wrapper_extend_error(self):
...
+                               namespace='hdmf-common',
+                               type_map=self.tm,
+                               description='Homo sapiens')
+
+    def test_already_wrapped_warn(self):
+        with self.assertWarns(Warning):
+            terms = TermSet(term_schema_path='tests/unit/example_test_term_set.yaml')
+            VectorData(name='foo',
+                       data=[0],
+                       description=TermSetWrapper(value='Homo sapiens', termset=terms))
```
```suggestion
        terms = TermSet(term_schema_path='tests/unit/example_test_term_set.yaml')
        with self.assertWarns(Warning):
            VectorData(name='foo',
                       data=[0],
                       description=TermSetWrapper(value='Homo sapiens', termset=terms))
```
Wrap just the line that you expect the warning to come from

tests/unit/test_term_set.py
```diff
@@ -215,3 +217,125 @@ def test_wrapper_extend_error(self):
...
+
+        unload_termset_config()
+
+    def test_warn_field_not_in_spec(self):
+        with self.assertWarns(Warning):
+            col1 = VectorData(name='col1',
+                              description='Homo sapiens',
+                              data=['1a', '1b', '1c', '2a'])
```
```suggestion
        col1 = VectorData(name='col1',
                                  description='Homo sapiens',
                                  data=['1a', '1b', '1c', '2a'])
        with self.assertWarns(Warning):
```

src/hdmf/common/__init__.py
rly: Having production code that exists solely to make testing easier is generally frowned upon, because the tests should execute the code as if a user was, and they will not / should not be using `return_map=True`. Here, I understand that it is tedious to access the type map, but I think there are other ways to do what you are trying to do (see below)

tests/unit/test_term_set.py
```diff
@@ -215,3 +217,125 @@ def test_wrapper_extend_error(self):
...
+        self.type_map = type_map
+        self.description = description
+
+
+class TestTermSetConfigVectorData(TestCase):
+    def setUp(self):
+        if not REQUIREMENTS_INSTALLED:
+            self.skipTest("optional LinkML module is not installed")
+
+        self.tm = load_termset_config(config_path='tests/unit/hdmf_config.yaml', return_map=True)
```
rly: It looks like `self.tm` is used solely to initialize `ExtensionContainer`. You have code above that sets `type_map = __TYPE_MAP` in `register_container_class`. Instead of using the special testing-only `return_map=True` functionality, we could decorate the `ExtensionContainer` class with `register_container_class`, which would then set the type map on the class, and this would also more accurately mimic a user's actions.

src/hdmf/term_set.py
mavaylon1: the yaml from ruamel parses differently, resulting in a different output. We already have this try except for yaml merged prior in TermSet.

src/hdmf/common/__init__.py
mavaylon1: Okay a couple things:
1. I can make get_type_map preserve the config as follows: (within get_type_map)
```
global_config = __TYPE_MAP.ts_config
    if global_config is not None:
        type_map.ts_config = global_config
    return type_map
```
However, this gives is a new problem. I'm assigning the type map attribute in the container, but that value is now the copy. That copy does not get updated when we load a config because we are updating __TYPE_MAP.

src/hdmf/container.py
mavaylon1: Can do, but what case would they be different and why do we do that?

tests/unit/test_term_set.py
mavaylon1: There is no spec for ExtensionContainer. In another PR I can expand our test containers a bit, but I think it is more efficient to build this minimally for this specific test. I did do some changes to clean it up though.

Also what is register_container_class? Is that register_class?

mavaylon1:
I have some scratch work left in there to show. Once we decide what to do with the type map, they will be removed. Everything else has been updated.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 3.14.0 (March 20, 2024)
```
```suggestion
## HDMF 3.14.0 (Upcoming)
```

requirements-opt.txt
```diff
@@ -1,8 +1,6 @@
 # pinned dependencies that are optional. used to reproduce an entire development environment to use HDMF
-tqdm==4.66.2
-zarr==2.17.1
-linkml-runtime==1.7.3; python_version >= "3.9"
+tqdm==4.66.1
```
rly:
```suggestion
tqdm==4.66.2
```

requirements-opt.txt
```diff
@@ -1,8 +1,6 @@
 # pinned dependencies that are optional. used to reproduce an entire development environment to use HDMF
 tqdm==4.66.2
-zarr==2.17.1
-linkml-runtime==1.7.3; python_version >= "3.9"
+zarr==2.16.1
```
rly:
```suggestion
zarr==2.17.1
```

src/hdmf/build/manager.py
```diff
@@ -391,25 +392,32 @@ def data_type(self):
...
+        if type_config is None:
+            type_config = TypeConfigurator()
         self.__ns_catalog = namespaces
         self.__mappers = dict()  # already constructed ObjectMapper classes
         self.__mapper_cls = dict()  # the ObjectMapper class to use for each container type
         self.__container_types = OrderedDict()
         self.__data_types = dict()
         self.__default_mapper_cls = mapper_cls
         self.__class_generator = ClassGenerator()
+        self.ts_config = type_config
```
rly:
```suggestion
        self.type_config = type_config
```
Can we rename this to `type_config` internally?

src/hdmf/common/__init__.py
```diff
@@ -20,6 +20,30 @@
...
+@docval({'name': 'config_path', 'type': str, 'doc': 'Path to the configuration file.'},
+        is_method=False)
+def load_type_config(**kwargs):
+    """
+    This method will either load the default config or the config provided by the path.
+    """
+    config_path = kwargs['config_path']
+    __TYPE_MAP.ts_config.load_type_config(config_path)
+
+def get_loaded_config():
```
rly:
```suggestion
def get_loaded_type_config():
```
For consistency, I think we should either use "type config" or just "config". I think "type config" is fine.

src/hdmf/container.py
```diff
@@ -85,10 +86,82 @@ def setter(self, val):
...
-            self.fields[name] = val
+            self.fields[name] = self._field_config(arg_name=name, val=val)
 
         return setter
 
+    def get_data_type(self):
+        """
+        Method that return data type.
+        """
+        return getattr(self, self._data_type_attr)
```
rly:
```suggestion
    @property
    def data_type(self):
        """
        Return the spec data type associated with this container.
        """
        return getattr(self, self._data_type_attr)
```

src/hdmf/container.py
```diff
@@ -85,10 +86,82 @@ def setter(self, val):
...
+
+                    # In the case of dealing with datasets directly or not defined in the spec.
+                    # (Data/VectorData/DynamicTable/etc)
+                    if spec is None:
+                        msg = "Spec not found for %s." % attr
+                        warn(msg)
+                        return val
+                    else:
+                        # If the val has been manually wrapped then skip checking the config for the attr
+                        if type(val) == TermSetWrapper:
```
rly:
```suggestion
                        if isinstance(val, TermSetWrapper):
```
This is more general and allows for subclassing of `TermSetWrapper`.

src/hdmf/term_set.py
```diff
@@ -304,3 +299,85 @@ def extend(self, arg):
...
+    @docval({'name': 'config_path', 'type': str, 'doc': 'Path to the configuration file.'})
+    def load_type_config(self,config_path):
+        """
+        Load the configuration file for validation on the fields defined for the objects within the file.
+        """
+        try:
+            from ruamel.yaml import YAML
+        except ImportError: # pragma: no cover
+            msg = "Install yaml."
+            raise ValueError(msg)
```
rly: This should never happen because the import would fail way earlier. I think we can cut this out and move the import to the top of the file.

src/hdmf/term_set.py
```diff
@@ -162,19 +162,21 @@ def __schemasheets_convert(self):
         This method returns a path to the new schema to be viewed via SchemaView.
         """
         try:
-            import yaml
+            from ruamel.yaml import YAML
```
rly: This can be moved to the top of the file.

mavaylon1: I think it should stay because YAML is not required for TermSet. It's only needed for schemasheets which is one method.

rly:
@mavaylon1 At one of our recent meetings, we discussed not creating a global type config, but rather have a type config be associated with a particular type map, so that it is associated with whatever schema/namespaces are loaded at the moment. It looks like here, in the type map copy method, you pass the same type config around, so it is global instead. Can you change that?

src/hdmf/term_set.py
mavaylon1: Why would this never happen? If the user uses Termset without schemasheets, then they won't have this installed.

mavaylon1:
> @mavaylon1 At one of our recent meetings, we discussed not creating a global type config, but rather have a type config be associated with a particular type map, so that it is associated with whatever schema/namespaces are loaded at the moment. It looks like here, in the type map copy method, you pass the same type config around, so it is global instead. Can you change that?

I must have misunderstood because I disagree. When we load a namespace, we are merging in the sense that the copy of the global type map can now navigate all namespaces. If a user wants to load a config that's unique to the namespace, they can call load and that too will be merged. That way I'm not searching for which config to use for which namespace. It's all under one.

My view is that the config file is associated with a namespace extension. Not the configurator.

Also if I get a copy of a the global type map. The type map that is used throughout hdmf, why would I have a config unique to that copy? That would imply I have to use that type map in other places. That's so much more difficult.

src/hdmf/term_set.py
rly: `ruamel.yaml` is a requirement of HDMF: https://github.com/hdmf-dev/hdmf/blob/dev/pyproject.toml#L37
It's used in https://github.com/hdmf-dev/hdmf/blob/000020232cbf6dcfb053c4d57d984445c2eaa0e7/src/hdmf/spec/namespace.py#L2
and should be automatically imported when one runs `import hdmf`.

mavaylon1: Got it.

src/hdmf/common/__init__.py
```diff
@@ -20,6 +20,30 @@
 # a global type map
 global __TYPE_MAP
 
+@docval({'name': 'config_path', 'type': str, 'doc': 'Path to the configuration file.'},
+        is_method=False)
+def load_type_config(**kwargs):
+    """
+    This method will either load the default config or the config provided by the path.
+    """
```
rly, line 29:
```suggestion
    NOTE: This config is global and shared across all type maps.
    """
```

rly (approved):

rly:
Let's discuss whether the TypeConfigurator should be global or specific to an IO object in a separate PR.

PR #1031, opened 2024-01-17
Fix internal links in docs

stephprince:
## Motivation

See related PR in https://github.com/NeurodataWithoutBorders/pynwb/pull/1827

 There are some broken internal links in the hdmf documentation that impact pynwb as well.
- Links to h5py/pandas data types are not currently working - (see [example in parameters section here](https://hdmf.readthedocs.io/en/stable/hdmf.query.html#hdmf.query.HDMFDataset)) 
- Links to return types are not generated if a function has more than one return type, or a nested return type (see [example in the get_container method here](https://hdmf.readthedocs.io/en/stable/hdmf.common.multi.html#hdmf.common.multi.SimpleMultiContainer.get_container))

## How to test the behavior?
To see linkcheck warnings:
```
sphinx-build -n -b linkcheck ./docs/source ./docs/_build
```

To see updated docs:
```
cd docs
make html
```

## Checklist

- [x] Did you update `CHANGELOG.md` with your changes?
- [X] Does the PR clearly describe the problem and the solution?
- [X] Have you reviewed our [Contributing Guide](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst)?
- [X] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged?

PR #1036, opened 2024-01-23
Update docval to accept fully qualified class names

rly:
## Motivation

Follow-up to https://github.com/hdmf-dev/hdmf/pull/1031 to allow a "type" argument in docval that is the name of a class to be linked to its documentation.

## How to test the behavior?

Use this syntax for `'type'` if you want to create a proper doc link to the `LinkSpec` class:
```python
@docval({'name': 'spec', 'type': 'hdmf.spec.spec.LinkSpec', 'doc': 'the specification for the object to link to'})
    def set_link(self, **kwargs):
        pass
```

## Checklist

### DIFF
--- a/tests/unit/test_io_hdf5_streaming.py
+++ b/tests/unit/test_io_hdf5_streaming.py
@@ -19,6 +19,7 @@ class TestRos3(TestCase):
     def setUp(self):
         # Skip ROS3 tests if internet is not available or the ROS3 driver is not installed
         try:
+            # this is a 174 KB file
             urllib.request.urlopen("https://dandiarchive.s3.amazonaws.com/ros3test.nwb", timeout=1)
         except urllib.request.URLError:
             self.skipTest("Internet access to DANDI failed. Skipping all Ros3 streaming tests.")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # HDMF Changelog
 
+## HDMF 3.12.1 (Upcoming)
+
+### Bug fixes
+- Fixed internal links in docstrings and tutorials. @stephprince [#1031](https://github.com/hdmf-dev/hdmf/pull/1031)
+- Fixed issue with creating documentation links to classes in docval arguments. @rly [#1036](https://github.com/hdmf-dev/hdmf/pull/1036)
+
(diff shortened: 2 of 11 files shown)


tests/unit/test_io_hdf5_streaming.py
```diff
@@ -19,6 +19,7 @@ class TestRos3(TestCase):
     def setUp(self):
         # Skip ROS3 tests if internet is not available or the ROS3 driver is not installed
         try:
+            # this is a 174 KB file
```
stephprince, line 22: not sure if this was meant to be added here?

stephprince (commented):
I think one more type that could be changed with this new docval behavior is `'type': 'HERD'` to `'type':'hdmf.common.resources.HERD'`? It would apply [here](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/hdf5/h5tools.py#L366), [here](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/io.py#L78), and [here](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/io.py#L78)

otherwise looks good to me!

rly:
> I think one more type that could be changed with this new docval behavior is `'type': 'HERD'` to `'type':'hdmf.common.resources.HERD'`? It would apply [here](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/hdf5/h5tools.py?rgh-link-date=2024-01-24T01%3A03%3A01Z#L366), [here](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/io.py?rgh-link-date=2024-01-24T01%3A03%3A01Z#L78), and [here](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/io.py?rgh-link-date=2024-01-24T01%3A03%3A01Z#L78)

I am not sure how I missed that. Good catch!

tests/unit/test_io_hdf5_streaming.py
rly: oops. i did not mean to add that. it is technically unrelated but I think it is a useful comment and it's kind of silly to make it its own PR, so i would prefer ot keep it

stephprince (approved):

PR #1040, opened 2024-01-25
Expose AWS Region to HDF5IO

### LINKED ISSUE #1039: [Bug]: Expose aws_region to HDF5IO
### What happened?

While establishing benchmarks for various streaming methods, I ran into an error I hadn't observed in the past, Point of termination is https://github.com/h5py/h5py/blob/6b512e5edf80f6660e0f07c704ab59faa733008b/h5py/_hl/files.py#L507

Since older versions of h5py don't necessarily run into this, one argument is that `ros3` is only supported under those older versions

But it's also only one extra kwarg so easy to pass along

### Steps to Reproduce

```python
import pynwb

s3_url = "s3://dandiarchive/blobs/8c5/65f/8c565f28-e5fc-43fe-8fb7-318ad2081319"
io = pynwb.NWBHDF5IO(path=s3_url, mode="r", load_namespaces=True, driver="ros3")
nwbfile = io.read()
acquisition_path = "ElectricalSeriesAp"
slice_range = (slice(0, 30_000), slice(0, 384))  # ~23 MB
nwbfile.acquisition[acquisition_path].data[slice_range]
```

### Traceback

_No response_

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.11

### Package Versions

_No response_

CodyCBakerPhD:
fix #1039 

## Motivation

Shows the two points of exposure required to fix this. Would be great if someone could take this PR over and properly expose it as an argument (not sure how tests should handle beyond just making sure it can pass it down...)

Also not sure why the `h5py.File` is instantiated twice

## Checklist


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -63,15 +63,16 @@ def can_read(path):
             {'name': 'file', 'type': [File, "S3File", "RemFile"],
              'doc': 'a pre-existing h5py.File, S3File, or RemFile object', 'default': None},
             {'name': 'driver', 'type': str, 'doc': 'driver for h5py to use when opening HDF5 file', 'default': None},
+            {'name': 'aws_region', 'type': str, 'doc': 'If driver is ros3, then specify the aws region of the url.'},
```
CodyCBakerPhD:
```suggestion
            {'name': 'aws_region', 'type': str, 'doc': 'If driver is ros3, then specify the aws region of the url.', 'default': None},
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 3.12.3 (Upcoming)
+
+### Bug fixes
+- Exposed `aws_region` to `NWBHDF5IO` and downstream passes to `h5py.File`. @codycbakerphd [#1040](https://github.com/hdmf-dev/hdmf/pull/1040)
```
CodyCBakerPhD:
```suggestion
- Exposed `aws_region` to `HDF5IO` and downstream passes to `h5py.File`. @codycbakerphd [#1040](https://github.com/hdmf-dev/hdmf/pull/1040)
```

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -134,13 +145,13 @@ def __check_path_file_obj(cls, path, file_obj):
         return path
 
     @classmethod
-    def __resolve_file_obj(cls, path, file_obj, driver):
+    def __resolve_file_obj(cls, path, file_obj, driver, aws_region):
```
CodyCBakerPhD:
```suggestion
    def __resolve_file_obj(cls, path, file_obj, driver, aws_region=None):
```

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -134,13 +145,13 @@ def __check_path_file_obj(cls, path, file_obj):
...
     @classmethod
-    def __resolve_file_obj(cls, path, file_obj, driver):
+    def __resolve_file_obj(cls, path, file_obj, driver, aws_region=None):
         path = cls.__check_path_file_obj(path, file_obj)
 
         if file_obj is None:
             file_kwargs = dict()
             if driver is not None:
-                file_kwargs.update(driver=driver)
+                file_kwargs.update(driver=driver, aws_region=bytes(aws_region, "ascii"))
```
CodyCBakerPhD:
```suggestion
                file_kwargs.update(driver=driver)
                
                if aws_region is not None:
                    file_kwargs.update(aws_region=bytes(aws_region, "ascii"))
```

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -134,13 +145,16 @@ def __check_path_file_obj(cls, path, file_obj):
         return path
 
     @classmethod
-    def __resolve_file_obj(cls, path, file_obj, driver):
+    def __resolve_file_obj(cls, path, file_obj, driver, aws_region=None):
```
CodyCBakerPhD, line 147: Codecov complaining of lack of hit on this line (and 157 by extension) - unsure how to add a test for that

So I'll let @rly, @oruebel , or @mavaylon1 add that or accept with reduced coverage

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -757,6 +771,9 @@ def open(self):
             if self.driver is not None:
                 kwargs.update(driver=self.driver)
 
+                if self.driver == "ros3" and self.aws_region is not None:
+                    kwargs.update(aws_region=bytes(self.aws_region, "ascii"))
```
CodyCBakerPhD, line 779: Hmmm, again not sure how to test this `open` method, I don't think I ever use that with respect to this class

Deferring to y'all to add a test

CodyCBakerPhD:
@oruebel Needed this PR (and probably a follow-up on PyNWB) for NWB Benchmarks

@rly @mavaylon1 I added a basic test but remaining lines of coverage I have no clue how to evoke, leaving it to you to finish that coverage if you care about it

mavaylon1:
I can take a look after I fly back this weekend. Ideally we want coverage. With @rly on vacation, I'll defer to @oruebel on that. But again, I'll take a look next week.

mavaylon1:
@oruebel I think you are better suited for this PR. Can you take this one and then we talk to get me up to speed.

src/hdmf/backends/hdf5/h5tools.py
oruebel: The ``__resolve_file_obj`` function seems to be used only by the `load_namespaces` and `get_namespaces` function. So to test the function for `aws` you could add a test for `ros3` that calls one of those functions. However, I think the simpler approach may be to just have a test that calls the method on its own and asserts that the file is opened correctly; something like"

```
fileobj = HDF5IO._HDF5IO__resolve_file_obj(path=<s3path>, file_obj=None, driver='ros3', aws_region=<you aws region>)

assertIsInstance(fileobj, h5py.File)
assertEquals(fileobj.path, <s3path>)
```

src/hdmf/backends/hdf5/h5tools.py
oruebel: `open` is called in here

https://github.com/hdmf-dev/hdmf/blob/f163a768dc33409df5014f4c4e487f1713c7d61c/src/hdmf/backends/io.py#L41

in the parent `HDMFIO.__init__`. However, it looks like the `awsregion` etc. is not being passed through:

https://github.com/hdmf-dev/hdmf/blob/f163a768dc33409df5014f4c4e487f1713c7d61c/src/hdmf/backends/hdf5/h5tools.py#L104

So, normally, your `test_basic_read_with_aws_region` should excercise  your new code in `open`  method. The fact that it does not get hit seems suspicious. I.e., I would add a breackpoint in line 772 to check what `self.dirver` and `self.aws_region` are set to when you run that test.

src/hdmf/backends/hdf5/h5tools.py
rly: I just updated `load_namespaces` and `get_namespaces` to take `aws_region` and pass it to `__resolve_file_obj`. Otherwise an h5py File with an s3 path will not open.

rly: I think all of our tests use http instead of S3 paths, so this is new. I'll add some tests today/this weekend.

oruebel: @rly while you are at it, could you also add a docstring for `__resolve_file_obj`. It is currently not clear that this function is specific for the loading of namespaces and it would be helpful to have that documented.

oruebel:
@rly can you help finish up this PR?

src/hdmf/backends/hdf5/h5tools.py
rly: I think coverage was not showing a hit for these lines because previously, coverage was not run when running the ROS3 tests (a separate ros3 environment has to be set up). I just changed that so that we run coverage for the ROS3 tests as well.

rly:
Review notes: 
- HDMF-zarr tests are failing for unrelated issues.

rly (approved):

PR #1059, opened 2024-02-22
Update release.md

mavaylon1:
## Motivation

Remove a duplicate bullet point.

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release.md
@@ -9,7 +9,6 @@ Prepare for release of HDMF [version]
 - [ ] Check legal file dates and information in `Legal.txt`, `license.txt`, `README.rst`, `docs/source/conf.py`,
   and any other locations as needed
 - [ ] Update `pyproject.toml` as needed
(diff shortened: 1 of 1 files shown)


stephprince:
There's nothing in the `README.rst` other than the legal file dates that gets updated usually, right? I was wondering if that's why they were separate steps originally.

mavaylon1:
My view is that "checking information" is the same as "Update". I think this is just an interpretation. I'll leave this as is.

oruebel:
Originally the README contained more detailed information, e.g,. about installing HDMF, so there was more to check then just the legal dates. In general, checking the README is there to make sure it is up-to-date with any changes that have happened in the repo, since this is the first thing developers see when they come to the repo. 

Case in point, currently there is a broken badge in the README that points to a non-existing GitHub workflow:

Screen Shot 2024-02-22 at 9 26 48 AM

It looks like the broken badge is pointing to `check_external_links.yml`, which I assume has been renamed to `check_sphinx_links.yml`

I think taking a look at the README before a release is still a good idea, even if no actions are needed in most cases.

PR #1061, opened 2024-03-03
Compound Dataset Support for TermSetWrapper

### LINKED ISSUE #938: [Feature]: TermSet Restructure (Stage 5): Compound Data

Currently we support compound data functionality; however, TermSet does not yet support this format. 

```
data = Data(
    name='data_name',
    data=np.array(
        [('Mus musculus', 9, 81.0), ('Homo sapiens', 3, 27.0)],
        dtype=[('species', 'U14'), ('age', 'i4'), ('weight', 'f4')]
    )
)
```
In this example, we could have a `TermSet` for species and for units. The wrapper method from #928 will need to be adapted to support multiple TermSets on a dataset. 

_No response_

In discussion on how to approach this.

Yes.

### Code of Conduct

mavaylon1:
## Motivation

Add the ability to wrap compound data. This only supports one field in the compound type. (first step)

Fix #938 

TODO:

- [x] Update documentation/tutorials
- [x] Update tests

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/docs/gallery/plot_term_set.py
+++ b/docs/gallery/plot_term_set.py
@@ -141,7 +141,7 @@ data = VectorData(
     name='species',
     description='...',
     data=TermSetWrapper(value=c_data, termset=terms, field='species')
-    )
+)
 
 
 ######################################################
--- a/docs/gallery/plot_term_set.py
+++ b/docs/gallery/plot_term_set.py
@@ -67,6 +67,7 @@ for this tutorial, which provides a concise example of how a term set schema loo
 """
 from hdmf.common import DynamicTable, VectorData
 import os
+import numpy as np
 
 try:
     import linkml_runtime  # noqa: F401
--- a/docs/gallery/plot_term_set.py
+++ b/docs/gallery/plot_term_set.py
@@ -143,7 +143,6 @@ data = VectorData(
     data=TermSetWrapper(value=c_data, termset=terms, field='species')
 )
 
-
 ######################################################
 # Validate Attributes with TermSetWrapper
 # ----------------------------------------------------
--- a/docs/gallery/plot_term_set.py
+++ b/docs/gallery/plot_term_set.py
@@ -130,6 +130,20 @@ data = VectorData(
     )
 
 ######################################################
+# Validate Compound Data with TermSetWrapper
+# ----------------------------------------------------
+# :py:class:`~hdmf.term_set.TermSetWrapper` can be wrapped around compound data.
+# The user will set the field within the compound data type that is to be validated
+# with the termset.
+c_data = np.array([('Homo sapiens', 24)], dtype=[('species', 'U50'), ('age', 'i4')])
+data = VectorData(
+    name='species',
+    description='...',
+    data=TermSetWrapper(value=c_data, termset=terms, field='species')
+    )
+
+
+######################################################
 # Validate Attributes with TermSetWrapper
 # ----------------------------------------------------
 # Similar to wrapping datasets, :py:class:`~hdmf.term_set.TermSetWrapper` can be wrapped around any attribute.
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -21,12 +21,14 @@ def append_data(data, arg):
         return data
     elif isinstance(data, np.ndarray):
         if isinstance(arg, np.ndarray):
-            if data.ndim != arg.ndim:
-                return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
-            else:
+            if data.ndim == arg.ndim:
                 # arg is a structured array or an array with matching data
                 # dimensions
                 return np.append(data, arg)
+            else: # arg is a row vector
+                return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
+        else: # arg is a scalar
+            return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
     elif isinstance(data, h5py.Dataset):
         shape = list(data.shape)
         shape[0] += 1
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -20,7 +20,9 @@ def append_data(data, arg):
         data.append(arg)
         return data
     elif isinstance(data, np.ndarray):
-        if len(data.dtype)>0: # arg is a structured array
+        if len(data.dtype)>0 or isinstance(arg, np.ndarray):
+            # arg is a structured array or an array with matching data
+            # dimensions
             return np.append(data, arg)
         else: # arg is a scalar or arg is a row being appended to a matrix
             return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -20,14 +20,10 @@ def append_data(data, arg):
         data.append(arg)
         return data
     elif isinstance(data, np.ndarray):
-        if isinstance(arg, np.ndarray):
-            if data.ndim == arg.ndim:
-                # arg is a structured array or an array with matching data
-                # dimensions
-                return np.append(data, arg)
-            else: # arg is a row vector
-                return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
-        else: # arg is a scalar
+        if len(data.dtype)>0:
+            # data is a structured array
+            return np.append(data, arg)
+        else: # arg is a scalar or row vector
             return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
     elif isinstance(data, h5py.Dataset):
         shape = list(data.shape)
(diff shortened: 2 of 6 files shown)


mavaylon1:
Notes:
1. Do we test for append/extend for compound data? Can't find it.

docs/gallery/plot_term_set.py
```diff
@@ -129,6 +130,20 @@
...
+# ----------------------------------------------------
+# :py:class:`~hdmf.term_set.TermSetWrapper` can be wrapped around compound data.
+# The user will set the field within the compound data type that is to be validated
+# with the termset.
+c_data = np.array([('Homo sapiens', 24)], dtype=[('species', 'U50'), ('age', 'i4')])
+data = VectorData(
+    name='species',
+    description='...',
+    data=TermSetWrapper(value=c_data, termset=terms, field='species')
+    )
```
```suggestion
)
```

src/hdmf/term_set.py
```diff
@@ -214,19 +214,25 @@ class TermSetWrapper:
...
-            values = self.__value
-        # create list if none of those -> mostly for attributes
+        if self.__field is not None:
+            values = self.__value[self.__field]
         else:
-            values = [self.__value]
+            # check if list, tuple, array
+            if isinstance(self.__value, (list, np.ndarray, tuple)):
+                values = self.__value
+            # create list if none of those -> mostly for attributes
```
```suggestion
            # create list if none of those - mostly for scalar attributes
```

rly:
@mavaylon1 can you add a test for appending to a VectorData with a compound type array?

tests/unit/common/test_table.py
```diff
@@ -221,6 +221,58 @@ def test_add_row_validate_bad_data_all_col(self):
         with self.assertRaises(ValueError):
             species.add_row(Species_1='bad data', Species_2='bad data')
 
+    def test_add_ref_compound_data_append(self):
```
rly:
```suggestion
    def test_compound_data_append(self):
```

tests/unit/common/test_table.py
```diff
@@ -221,6 +221,58 @@ def test_add_row_validate_bad_data_all_col(self):
...
+        compound_vector_data = VectorData(
+            name='Species_1',
+            description='...',
+            data=c_data
+        )
+        compound_vector_data.append(c_data2)
+
+        np.testing.assert_array_equal(compound_vector_data.data, np.append(c_data, c_data2))
+
+    def test_add_ref_compound_data_extend(self):
```
rly:
```suggestion
    def test_compound_data_extend(self):
```

src/hdmf/data_utils.py
```diff
@@ -20,7 +20,7 @@ def append_data(data, arg):
         data.append(arg)
         return data
     elif isinstance(data, np.ndarray):
-        return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
```
rly, line 23: Why make this change?

`np.append` will flatten multidimensional arrays. The result will always be one-dimensional. `VectorData` columns can have 2-D (or more) data, and with this change, calling `add_row` will flatten those 2-D data arrays into 1-D arrays. I guess since tests pass, we are probably missing a test for this.

```python
>>> a = [[1, 2, 3], [4, 5, 6]]
>>> b = [7, 8, 9]
>>> np.append(a, np.expand_dims(b, axis=0), axis=0)
array([[1, 2, 3],
       [4, 5, 6],
       [7, 8, 9]])
>>> np.append(a, b)
array([1, 2, 3, 4, 5, 6, 7, 8, 9])
```

mavaylon1: I have two issues with the original form:
1. The user does not know what magic is going under the hood. If the user wants to keep the dimensionality, it seems they should use np.vstack, which is extend. 
2. This came up because it does not work with compound arrays:
```
>>> c_data = np.array([('Homo sapiens', 24)], dtype=[('species', 'U50'), ('age', 'i4')])
>>> np.append(c_data, np.expand_dims(c_data, axis=0), axis=0)
  File "<stdin>", line 1, in <module>
  File "/Users/mavaylon/anaconda3/envs/nwb/lib/python3.11/site-packages/numpy/lib/function_base.py", line 5617, in append
    return concatenate((arr, values), axis=axis)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 1 dimension(s) and the array at index 1 has 2 dimension(s)
```
vs
```
>>> np.append(c_data, c_data)
array([('Homo sapiens', 24), ('Homo sapiens', 24)],
      dtype=[('species', '<U50'), ('age', '<i4')])
```
The fact this wasn't caught was due to missing tests.

mavaylon1:
Note: Coverage is not updated on the repo, but the check is passing for both the patch and project.

src/hdmf/data_utils.py
```diff
@@ -20,7 +20,10 @@ def append_data(data, arg):
...
+        if isinstance(arg, np.ndarray):
+            return np.append(data, arg)
```
rly, line 24: I tried this:
```python
a = np.array([[1, 2, 3], [4, 5, 6]])
b = np.array([7, 8, 9])
np.append(a, b)
```
The result is that the array is flattened: `array([1, 2, 3, 4, 5, 6, 7, 8, 9])`

That's not good and could break things.

I get that you want to use `append` for structured arrays. Can you change the `if` to check that it is a structured array instead of any numpy array?

rly (approved):

rly (changes requested):
Can you please check the failing coverage test where opt reqs=true? Why is it failing?

src/hdmf/data_utils.py
rly: I think this issue returns with your latest commit... please review and add a test for this case.

mavaylon1 (commented):
Using the dim of the array cover both compound data and arrays that have the same dimension.

mavaylon1:
> Can you please check the failing coverage test where opt reqs=true? Why is it failing?

Done @rly

src/hdmf/data_utils.py
```diff
@@ -20,7 +20,15 @@ def append_data(data, arg):
         data.append(arg)
         return data
     elif isinstance(data, np.ndarray):
-        return np.append(data,  np.expand_dims(arg, axis=0), axis=0)
+        if isinstance(arg, np.ndarray):
+            if data.ndim == arg.ndim:
```
rly: Note that with this change, when `data = np.array([[1, 2, 3], [4, 5, 6]])` and `arg = np.array([[7, 8, 9]])`, this code used to raise an error. Now it flattens the list to `array([1, 2, 3, 4, 5, 6, 7, 8, 9])`. 

`np.append` works for structured arrays here because we happen to be dealing only with 1D structured arrays, so flattening the array is OK - it is still 1D.

It looks like with these changes, you are trying to capture the case where you call `append_data` on:
```python
c_data = np.array([('Homo sapiens', 24)], dtype=[('species', 'U50'), ('age', 'i4')])
c_data2 = np.array([('Mus musculus', 24)], dtype=[('species', 'U50'), ('age', 'i4')])
```

But if we want to be consistent with past behavior, that _should_ raise an error, because `append_data` is designed to add one new row, and `extend_data` is designed to add multiple rows (this mimics python list append/extend)

To test adding one new compound dtype row, the test should be
```python
c_data = np.array([('Homo sapiens', 24)], dtype=[('species', 'U50'), ('age', 'i4')])
c_data2 = np.array(('Mus musculus', 24), dtype=[('species', 'U50'), ('age', 'i4')])
```
The unmodified `append_data` handles this case just fine. 

With these changes, we are adding new behavior that is inconsistent between structured arrays and non-structured arrays, and can result in silent unlikely-to-be-intended flattening for non-structured arrays. That all seems unnecessary.

I suggest, either
1. revert all changes to this function and change the test
2. use a more specific test for structured arrays like `len(dtype)>1` to add new behavior that allows a user to append a structured array of multiple elements to a structured array of multiple elements. i'm not a big fan of this though because of the inconsistency between structured and non-structured arrays.

mavaylon1: I'm removing my idea of supporting appending arrays.

tests/unit/common/test_table.py
```diff
@@ -220,6 +220,101 @@ def test_add_row_validate_bad_data_all_col(self):
...
+
+        terms = TermSet(term_schema_path='tests/unit/example_test_term_set.yaml')
+        compound_vector_data = VectorData(
+            name='Species_1',
+            description='...',
+            data=TermSetWrapper(value=c_data, termset=terms)
+        )
+
+        with self.assertRaises(ValueError):
+            compound_vector_data.append(c_data2)
```
rly:
```suggestion
        vectordata_termset = VectorData(
            name='Species_1',
            description='...',
            data=TermSetWrapper(value=c_data, termset=terms)
        )

        with self.assertRaises(ValueError):
            vectordata_termset.append(c_data2)
```

rly (approved):

PR #1064, opened 2024-03-04
Have VectorData expandable by default

mavaylon1:
## Motivation

Have VectorData expandable by default:

- [ ] Check that reading existing non-expandable data into the now expandable VectorData is not a problem (due to shape).
- [ ] Make to test that the data is wrapped with H5DataIO on read
- [ ] Update VectorData
- [ ] Add tests
- [ ] Investigate VectorIndex. When indexing vector index that has a wrapped target vector data, an error arises.
- [ ] TBD

Questions to look into:
- Do we have tests where we look at add_row when the data is wrapped with H5DataIO? ---> Yes 
- If a user wanted to make a EnumData column extendable, would they wrap both data and elements with H5DataIO?

Ideas:
- We support Zarr, so wrapping with H5DataIO for every instance does not work. Do we instead want a field that is default to True that will wrap data if true. In hdmf-zarr, we will override this to False (similar to how we import HERD in pynwb into a pynwb version in order to reset the typemap). We will also have logic that says if the user provides their own instance of DataIO, it will not wrap automatically either. We could also just skip the field and within HDMF-zarr just unwrap and reset data, but I understand if that is a little messy. 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -42,6 +42,7 @@ class VectorData(Data):
              'doc': 'a dataset where the first dimension is a concatenation of multiple vectors', 'default': list()},
             allow_positional=AllowPositional.WARNING)
     def __init__(self, **kwargs):
+        kwargs['data'] = H5DataIO(data=kwargs['data'], maxshape=(None,))
         description = popargs('description', kwargs)
         super().__init__(**kwargs)
         self.description = description
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -104,16 +104,16 @@ class VectorIndex(VectorData):
             allow_positional=AllowPositional.WARNING)
     def __init__(self, **kwargs):
         target = popargs('target', kwargs)
+        self.__uint = np.uint8
+        self.__maxval = 255
+        if isinstance(kwargs['data'], (list, np.ndarray)):
+            if len(kwargs['data']) > 0:
+                self.__check_precision(len(target))
+            # adjust precision for types that we can adjust precision for
+            kwargs['data'] = self.__adjust_precision(self.__uint, kwargs['data'])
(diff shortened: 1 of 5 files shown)


mavaylon1:
Note: This could be done by next release if backwards compatibility does not raise an issue. Otherwise it will be Future (the current set milestone).

src/hdmf/common/table.py
```diff
@@ -42,6 +42,7 @@ class VectorData(Data):
              'doc': 'a dataset where the first dimension is a concatenation of multiple vectors', 'default': list()},
             allow_positional=AllowPositional.WARNING)
     def __init__(self, **kwargs):
+        kwargs['data'] = H5DataIO(data=kwargs['data'], maxshape=(None,))
```
oruebel: Hardcoding `H5DataIO` will probably create issues with `ZarrIO`. In the best case, `ZarrIO` ignores the wrapper and worst case it would fail.... I'm not sure which. For `ZarrIO` we should not need to wrap for datasets to be expandable, but the problem is, we won't know which backend will be used until we call write to know which wrapper to use.

Not sure what the best solution is for this. I could imagine a few approaches:

1.  Have "expandable" as an option on the base `DataIO` class and wrap with `DataIO` as a generic wrapper that the specific backend (HD5FIO, ZarrIO etc.) would then need to know how to translate. ---> I think this should be feasible, but would require changes in both `HDF5IO` and `ZarrIO`
2. Maybe we would need to wrap in the build process  --->  I think this could work, but may be tricky 
3. Wrap with `H5DataIO` and require that the other backends know how to translate it  ---> I don't like this one, because it hard-codes backend-specific wrappers in the Container

mavaylon1: This is what was going through my mind as well. I made a note documenting that. I had a expandable field idea.

PR #1066, opened 2024-03-06
Warn when adding ragged arrays to DynamicTable without index argument

### LINKED ISSUE #1065: [Feature]: Warn to use index argument on call to add_row or add_column

See https://github.com/NeurodataWithoutBorders/pynwb/issues/1856. An error occurs during `io.write` when giving a column a ragged input across rows without providing an index.

Warn or error on trying to add the value while still in-memory rather than waiting until writing to disk

Yes.

stephprince:
## Motivation

Fix #1065. When a user tries to add a ragged array to a DynamicTable using `add_row` or `add_column`, a warning will occur prompting the user to use the `index` argument.  

I had some questions about these potential changes.
1. I found a related issue discussion that suggested that doing a full inspection of the data to detect whether it was ragged might be a costly operation - https://github.com/NeurodataWithoutBorders/pynwb/issues/1060. I was wondering if that's still a concern here or not.
2. Numpy's warning refers to the requested array having an "inhomogeneous shape". I wasn't sure if that type of phrasing is preferred over "ragged."

## How to test the behavior?
```python
from pynwb import NWBHDF5IO

# test adding individual trials
nwbfile = mock_NWBFile()
nwbfile.add_trial_column(name="test", description="")
nwbfile.add_trial(start_time=0.0, stop_time=2.0, test=[1, 2])
nwbfile.add_trial(start_time=0.0, stop_time=2.0, test=[3, 4])
nwbfile.add_trial(start_time=0.0, stop_time=2.0, test=[5, 6, 7])  # Warning here

### test adding full column
nwbfile = mock_NWBFile()
nwbfile.add_trial(start_time=0.0, stop_time=2.0)
nwbfile.add_trial(start_time=0.0, stop_time=2.0)
nwbfile.add_trial_column(name="test1", description="", data=[[1, 2], [1, 2]])
nwbfile.add_trial_column(name="test2", description="", data=[[1, 2], [1, 2, 3]])  # Warning here

with NWBHDF5IO('test.nwb', 'w') as f:
    f.write(nwbfile)  # Error here
```

## Checklist

### DIFF
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -958,9 +958,8 @@ def is_ragged(data):
     """
     Test whether a list of lists or array is ragged / jagged
     """
-    if hasattr(data, '__len__') and not isinstance(data, str):
-        lengths = [len(sub_data) if hasattr(sub_data, '__len__') and not isinstance(sub_data, str) else 1
-                for sub_data in data]
+    if isinstance(data, (list, tuple)):
+        lengths = [len(sub_data) if isinstance(sub_data, (list, tuple)) else 1 for sub_data in data]
         if len(set(lengths)) > 1:
             return True  # ragged at this level
 
--- a/src/hdmf/utils.py
+++ b/src/hdmf/utils.py
@@ -954,6 +954,21 @@ def to_uint_array(arr):
     raise ValueError('Cannot convert array of dtype %s to uint.' % arr.dtype)
 
 
+def is_ragged(data):
+    """
+    Test whether a list of lists or array is ragged / jagged
+    """
+    if hasattr(data, '__len__') and not isinstance(data, str):
+        lengths = [len(sub_data) if hasattr(sub_data, '__len__') and not isinstance(sub_data, str) else 1
+                for sub_data in data]
+        if len(set(lengths)) > 1:
+            return True  # ragged at this level
+
+        return any(is_ragged(sub_data) for sub_data in data)  # check next level
+
+    return False
+
+
 class LabelledDict(dict):
     """A dict wrapper that allows querying by an attribute of the values and running a callable on removed items.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 3.13.0 (Upcoming)
+
+### Enhancements
+- Added warning when using `add_row` or `add_column` to add a ragged array to `DynamicTable` without an index parameter. @stephprince [#1066](https://github.com/hdmf-dev/hdmf/pull/1066)
+
 ## HDMF 3.12.2 (February 9, 2024)
 
 ### Bug fixes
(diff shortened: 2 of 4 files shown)


src/hdmf/utils.py
```diff
@@ -954,6 +954,21 @@ def to_uint_array(arr):
     raise ValueError('Cannot convert array of dtype %s to uint.' % arr.dtype)
 
 
+def is_ragged(data):
+    """
+    Test whether a list of lists or array is ragged / jagged
+    """
+    if hasattr(data, '__len__') and not isinstance(data, str):
```
oruebel: I think this check may also need a ``and not instance(data, AbstractDataChunkInterator)``

rly:
> * I found a related issue discussion that suggested that doing a full inspection of the data to detect whether it was ragged might be a costly operation - [Added unit_columns items are saved as strings NeurodataWithoutBorders/pynwb#1060](https://github.com/NeurodataWithoutBorders/pynwb/issues/1060). I was wondering if that's still a concern here or not.

Yes, please correct me if I am wrong - it looks like we are iterating over every element in the array to check for raggedness. That could incur performance issues. I think we should add a flag `check_ragged` (default True) so that a user can disable the check if it is causing issues.

I don't know how we would get around checking every element...

Note that this issue should only come up if 1) the `data` is a list or tuple (arrays and hdf5 datasets cannot be ragged unless you give it an object type) or 2) the user adds rows one by one to the table and a cell in that row has a different shape than the shapes of the cell in the previous rows. (the [original post](https://github.com/NeurodataWithoutBorders/helpdesk/discussions/71#discussioncomment-8638454) mentioned using this method)

Related to 1), I suggest limiting the `is_ragged` check in `add_column` to just list and tuple.

Related to 2), it looks like adding a check for `add_row` is not implemented. I think that would be good to do.

> * Numpy's warning refers to the requested array having an "inhomogeneous shape". I wasn't sure if that type of phrasing is preferred over "ragged."

It seems like [numpy uses "ragged" internally](https://github.com/numpy/numpy/blob/69feeb343eda76b18fb80b33907ea9ade9426142/numpy/_core/src/multiarray/array_coercion.c#L1285). I wonder why they chose to use "inhomogeneous shape" publicly. Perhaps "inhomogeneous shape" is more clear to the non-programmer, but I feel like both terms are poor. Maybe we could say "data has elements that have different lengths and therefore data cannot be coerced into an N-dimensional array"?

stephprince:
> I think we should add a flag check_ragged (default True) so that a user can disable the check if it is causing issues.

Sounds like a good option!

>Note that this issue should only come up if 1) the data is a list or tuple (arrays and hdf5 datasets cannot be ragged unless you give it an object type)

Yes I wasn't sure how often people added arrays that were an object type but since it would still error when writing I think limiting this check make sense.

> Related to 2), it looks like adding a check for add_row is not implemented. I think that would be good to do.

I had a check in `add_row` that inspects the whole column for raggedness every time a row was added, or do you mean additional tests for adding a row?

>  Perhaps "inhomogeneous shape" is more clear to the non-programmer, but I feel like both terms are poor. 

I agree, I think the phrasing you suggested is good.

src/hdmf/utils.py
stephprince: If we limit the check to only lists/tuples I think that should cover this

rly (approved):
Looks great!

rly:
> I had a check in add_row that inspects the whole column for raggedness every time a row was added

Ah, I missed that.

PR #1070, opened 2024-03-15
Unwrap TermSetWrapper in ObjectMapper and not the IO

mavaylon1:
## Motivation

Goal: Remove the unwrapping of TermSetWrapper from HDF5IO to base HDMF. Why? The current setup only allows TermSetWrapper to be unwrapped for HDF5 and not Zarr. The desire is to not have two IO classes dealing with the unwrapping, but one shared tool.

Unwrap in the mapper. 
## How to test the behavior?
```
```

## Checklist


mavaylon1:
Review Notes:

I am going to investigate how this would work on references. We don't have tests for those situations (pointing to a container that has a wrapped data/attribute field. This PR is a fix to continue the status quo, but now outside the backend specific IO. I will make a new PR for the reference case. That PR will not be included in the release on March 19.

tests/unit/build_tests/mapper_tests/test_build.py
```diff
@@ -406,7 +411,6 @@ def set_up_specs(self):
         self.bar_data_spec = DatasetSpec(
             doc='A test dataset specification with a data type',
             data_type_def='BarData',
-            dtype='int',
```
rly: Why modify existing tests that test specific conditions instead of making a separate test for unwrapping wrapped data?

By changing `BuildDatasetExtAttrsMixin`, you changed the original design of `TestBuildDatasetRefinedDtype` - now instead of that class testing the replacement of dtype from int to int64, it tests the addition of dtype (from None to int64). Those are similar, but in general, please consider how modifying existing code affects other code that depends on it.

By modifying the `TestBuildDatasetAddedAttrs`, the test now tests two very unrelated things -- adding attributes to a spec and unwrapping data with a `TermSetWrapper` -- when it was designed to test a specific case before. This is not always bad, but it is generally a good practice when writing unit tests to test a specific feature or set of related features so that if the test fails, you have a better idea of what is not working. If I wanted to add/modify tests related to unwrapping during the build process, I would not know to look here. Also, this test will now only run if LinkML is installed instead of always running...

mavaylon1:
Review notes:
1. The link ml requirement is needed. If not installed, this test will skip like all the other termset tests.
2. I moved the test to test_io_h5_tools as this seems logical to me, though being in test_build.py seemed fine as well given it is a test on build.

rly:
> I moved the test to test_io_h5_tools as this seems logical to me, though being in test_build.py seemed fine as well given it is a test on build.

Could you move it to `test_build.py`? Unless I remember wrong, I don't think this test is related to `h5_tools.py`

mavaylon1:
> > I moved the test to test_io_h5_tools as this seems logical to me, though being in test_build.py seemed fine as well given it is a test on build.
> 
> Could you move it to `test_build.py`? Unless I remember wrong, I don't think this test is related to `h5_tools.py`

I can! I just thought you wanted it moved from your prior comment.

rly:
Sorry for the confusion. I meant that the existing tests should not be modified unnecessarily, but you can add classes or methods to `test_build.py`

rly (approved):

PR #1081, opened 2024-03-25
Write dimension labels to DatasetBuilder on build

### LINKED ISSUE #1077: [Feature]: Store dimension labels in DatasetBuilder for backends to write

See https://github.com/hdmf-dev/hdmf-zarr/issues/176#issuecomment-2013195105

On write/build, if a data type has dimension labels specified in the spec, then based on the dimensions of the data, we can infer the dimension labels for the data and store those on the DatasetBuilder. The backend (HDF5IO, ZarrIO) can then write the labels however it chooses.

On read/construct, inferring the dimension labels based on the dimensions of the data and the possible options in the spec is a bit trickier because the IO object creates the builders directly and does not have easy access to the spec for the data type. But maybe this is not needed. In the use case described above, it seems like the NWB Zarr files are being read outside of PyNWB. 

Adding full support for dimension labels that link across datasets automatically is a bigger feature that requires changes to the schema language and various schema: https://github.com/hdmf-dev/hdmf/pull/192. The effort could be revived but would take a lot longer than the original need.

- Add `dimension_labels` as an attribute on the `DatasetBuilder` (which may be None of the labels are unknown) [hdmf-dev/hdmf@5c85062/src/hdmf/build/builders.py#L321](https://github.com/hdmf-dev/hdmf/blob/5c8506216995f995b891da1e6b596ee42b7dd948/src/hdmf/build/builders.py#L321)
- Enhance `BuildManger.build` to set the `dimension_labels` for `DatasetBuilders` [hdmf-dev/hdmf@5c85062/src/hdmf/build/manager.py#L148](https://github.com/hdmf-dev/hdmf/blob/5c8506216995f995b891da1e6b596ee42b7dd948/src/hdmf/build/manager.py#L148)

Yes.

rly:
## Motivation

Fix #1077 

## How to test the behavior?
```
pytest
```

## Checklist


src/hdmf/build/warnings.py
```diff
@@ -15,6 +15,13 @@ class IncorrectQuantityBuildWarning(BuildWarning):
     pass
 
 
+class IncorrectDatasetShapeBuildWarning(BuildWarning):
```
rly, line 18: All of these warnings should really be errors because they result in an invalid file. @oruebel what do you think? Should we upgrade all of these warnings to errors?

rly: All except for `DtypeConversionWarning` and the generic `BuildWarning`

rly: And maybe also the `OrphanContainerWarning`.

mavaylon1:
@rly can you move `BuildDatasetShapeMixin` to a more general location? When importing this mixin class, it would be cleaner to have it a `test_utils.py` or `test_mixin.py`. 

Otherwise, looks good.

rly:
@mavaylon1 The place to move it would be https://github.com/hdmf-dev/hdmf/blob/dev/tests/unit/helpers/utils.py. However, moving that means also moving the container classes and mappers defined in this file `tests/unit/build_tests/mapper_tests/test_build.py`. We might then want to resolve any redundancies between the Bar, Foo, and Baz test classes. I can do that, but I think this task fits well under the test harness refactoring that you wanted to do. By keeping it here for now, it's clear that this mixin is used only in this file. Should I move it, or let you take care of this as part of the larger refactoring?

rly:
TODO: I will look into the coverage failure

mavaylon1 (approved):

PR #1089, opened 2024-04-01
Post init option for class generator

### LINKED ISSUE #1072: [Feature]: Add a `post_init` method to `get_class`

Extensions creates can build an API for their work using the following:

```python
AClass = get_class("ObjectOnTheSpec", "ndx-extension")
```

It would be nice to have a mechanism to do checks with their arugments when using the `get_class` convenience method. 

Something that looks like this:

```python
def post_init(**init_kwargs):
    
    # Do the check here

ACheckedClass = get_class("ObjectOnTheSpec", "ndx-extension", post_init=post_init)
```

No.

mavaylon1:
## Motivation

Fix #1072

Add optional post init method for get_class (class generation for extensions)

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 from datetime import datetime, date
+import types
 
 import numpy as np
 
@@ -35,6 +36,8 @@ class ClassGenerator:
             {'name': 'spec', 'type': BaseStorageSpec, 'doc': ''},
             {'name': 'parent_cls', 'type': type, 'doc': ''},
             {'name': 'attr_names', 'type': dict, 'doc': ''},
+            {'name': 'post_init_method', 'type': types.FunctionType, 'default': None,
+             'doc': 'The function used as a post_init method to validate the class generation.'},
             {'name': 'type_map', 'type': 'hdmf.build.manager.TypeMap', 'doc': ''},
             returns='the class for the given namespace and data_type', rtype=type)
     def generate_class(self, **kwargs):
@@ -42,8 +45,8 @@ class ClassGenerator:
         If no class has been associated with the ``data_type`` from ``namespace``, a class will be dynamically
         created and returned.
         """
-        data_type, spec, parent_cls, attr_names, type_map = getargs('data_type', 'spec', 'parent_cls', 'attr_names',
-                                                                    'type_map', kwargs)
+        data_type, spec, parent_cls, attr_names, type_map, post_init_method = getargs('data_type', 'spec', 'parent_cls', 'attr_names',
+                                                                    'type_map', 'post_init_method', kwargs)
 
         not_inherited_fields = dict()
         for k, field_spec in attr_names.items():
@@ -82,6 +85,8 @@ class ClassGenerator:
                              + str(e)
                              + " Please define that type before defining '%s'." % name)
         cls = ExtenderMeta(data_type, tuple(bases), classdict)
+        if post_init_method is not None:
+            setattr(post_init_method, cls.__postinit, True)
         return cls
 
 
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 from datetime import datetime, date
 import types
+import typing
 
 import numpy as np
 
@@ -36,7 +37,7 @@ class ClassGenerator:
             {'name': 'spec', 'type': BaseStorageSpec, 'doc': ''},
             {'name': 'parent_cls', 'type': type, 'doc': ''},
             {'name': 'attr_names', 'type': dict, 'doc': ''},
-            {'name': 'post_init_method', 'type': types.Callable, 'default': None,
+            {'name': 'post_init_method', 'type': typing.Callable, 'default': None,
              'doc': 'The function used as a post_init method to validate the class generation.'},
             {'name': 'type_map', 'type': 'hdmf.build.manager.TypeMap', 'doc': ''},
             returns='the class for the given namespace and data_type', rtype=type)
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -85,9 +85,15 @@ class ClassGenerator:
                              + str(e)
                              + " Please define that type before defining '%s'." % name)
         cls = ExtenderMeta(data_type, tuple(bases), classdict)
+        breakpoint()
         if post_init_method is not None:
             setattr(post_init_method, cls.__postinit, True)
         return cls
+
+        cls = ExtenderMeta(data_type, tuple(bases), classdict)
+        if post_init_method is not None:
+            setattr(A, 'post_init_method', post_init_method)
+
 
 
 class TypeDoesNotExistError(Exception):  # pragma: no cover
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -36,7 +36,7 @@ class ClassGenerator:
             {'name': 'spec', 'type': BaseStorageSpec, 'doc': ''},
             {'name': 'parent_cls', 'type': type, 'doc': ''},
             {'name': 'attr_names', 'type': dict, 'doc': ''},
-            {'name': 'post_init_method', 'type': types.FunctionType, 'default': None,
+            {'name': 'post_init_method', 'type': types.Callable, 'default': None,
              'doc': 'The function used as a post_init method to validate the class generation.'},
             {'name': 'type_map', 'type': 'hdmf.build.manager.TypeMap', 'doc': ''},
             returns='the class for the given namespace and data_type', rtype=type)
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -322,7 +322,7 @@ class CustomClassGenerator:
                 fixed_value_attrs_to_set.append(attr_name)
             elif attr_name not in attrs_not_to_set:
                 attrs_to_set.append(attr_name)
-        
+
         @docval(*docval_args, allow_positional=AllowPositional.WARNING)
         def __init__(self, **kwargs):
             original_kwargs = dict(kwargs)
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -1,7 +1,6 @@
 from copy import deepcopy
 from datetime import datetime, date
 import types
-import typing
 
 import numpy as np
 
@@ -37,7 +36,7 @@ class ClassGenerator:
             {'name': 'spec', 'type': BaseStorageSpec, 'doc': ''},
             {'name': 'parent_cls', 'type': type, 'doc': ''},
             {'name': 'attr_names', 'type': dict, 'doc': ''},
-            {'name': 'post_init_method', 'type': typing.Callable, 'default': None,
+            {'name': 'post_init_method', 'type': types.FunctionType, 'default': None,
              'doc': 'The function used as a post_init method to validate the class generation.'},
             {'name': 'type_map', 'type': 'hdmf.build.manager.TypeMap', 'doc': ''},
             returns='the class for the given namespace and data_type', rtype=type)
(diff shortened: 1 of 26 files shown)


mavaylon1:
Note: This is not ready for review. I need to make tests. However, I think the concept is there to talk about. @rly @oruebel

oruebel:
> I think the concept is there to talk about.

The concept looks good to me

src/hdmf/build/classgenerator.py
```diff
@@ -35,15 +36,17 @@ def register_generator(self, **kwargs):
             {'name': 'spec', 'type': BaseStorageSpec, 'doc': ''},
             {'name': 'parent_cls', 'type': type, 'doc': ''},
             {'name': 'attr_names', 'type': dict, 'doc': ''},
+            {'name': 'post_init_method', 'type': types.FunctionType, 'default': None,
```
rly:
```suggestion
            {'name': 'post_init_method', 'type': types.Callable, 'default': None,
```

`typing.Callable` is generally recommended over `types.FunctionType` because it is more general and can be parameterized with the input argument types and return type

rly:
The approach looks good to me too

src/hdmf/build/classgenerator.py
mavaylon1: Is there a callable case that's not a function that we would want included by using callable type?

rly:
Note that you will have to be careful with methods vs functions here. It seems like the `post_init` used in `ExtenderMeta` acts on the class and not the instance, and most likely users will want to run code on the instance. You may need to create a separate `post_init` system for generated classes that allows you to do that.

src/hdmf/build/classgenerator.py
rly: In rare cases, a user may want to make a class callable by implementing `__call__`. There are some examples here: https://realpython.com/python-callable-instances/#putting-pythons-__call__-into-action . No reason to exclude that, in my opinion.

mavaylon1: I tried this, but it says Callable is not recognized from the types package. Are you sure that is the proper syntax?

rly: Sorry. It's in the `typing` package

mavaylon1:
@rly I still need to add the post init to the MCIClassGenerator, but I have a new proof of concept and a test that works.

tests/unit/build_tests/test_classgenerator.py
```diff
@@ -82,6 +82,58 @@ def test_no_generators(self):
...
+
+
+    def test_post_init(self):
+        def post_init_method(self, **kwargs):
+            attr1 = kwargs['attr1']
+            if attr1<10:
+                msg = "attr1 should be >=10"
+                raise ValueError(msg)
+
+        cls = self.type_map.get_dt_container_cls('Baz', CORE_NAMESPACE, post_init_method)
```
rly: I think this approach binds the method to the class rather than the instance.
```python
cls.post_init_method = tp.MethodType(post_init_method, cls) # set as bounded method
```
acts on the class and creates a classmethod, so the first argument is set to the class instead of the instance.

If your method was instead:
```python
    def post_init_method(self, **kwargs):
            self.attr1 = "hi"  # do anything that modifies `self` like change a dtype or set an attribute
            # this will actually set a class variable "attr1" that is shared across all classes and is distinct from the instance variable "attr1"
```

I believe if you call 
```python
obj = cls(name='instance', attr1=9)
print(obj.attr1)  # should still be 9
```

I don't think you need this line:
```python
cls.post_init_method = tp.MethodType(post_init_method, cls) # set as bounded method
```

mavaylon1: It does bind it to the class. My thought process was to encapsulate the post init method with the class. 

I'll look into the concern that it's a class method setting a class variable. I don't think that's the case.

rly: > encapsulate the post init method with the class

I think `cls.post_init_method = post_init_method` (monkey patching) might be sufficient and create a method that is not a class method

mavaylon1: I think we need to establish what we want.

- Add a post init (when provided) such that the kwargs dict of an instance of the class is validated us that post init.

That means we need the class to always have that post init method. This means we need to bind the function as a method to the class. This is that:

```
cls.post_init_method = tp.MethodType(post_init_method, cls) # bounded method
```
We can go one of two ways:
1 . Keep that line above. What this will do is make the function a bounded method. This means when called, it will provide self as a positional arg. This requires that the user defined the post init as 
```
def foo(self, **kwargs):
```
If they instead do `def foo(**kwargs)`, then an error will return. It will say foo does not take positional arguments but 1 was provided. This is because it is being given "self", because we assigned it as a bounded method.

2. We can do `cls.post_init_method = post_init_method`. This assigns the method to the class. It does not take self. It does not take in cls. It is just a method that takes in kwargs. This is also fine. I think this edge out the first idea because users won't have to define "self" in the class definition parameters.  

Thoughts? @rly

mavaylon1: I actually don't think we need to make it a bounded method. There's no benefit to using self since it is only operating on kwargs. However, if we want to expand the scope so that the post_init does something with the instance variables then that is a different conversation.

oruebel: > However, if we want to expand the scope so that the post_init does something with the instance variables then that is a different conversation.

Generally, I think the expectation would be that a `post_init` method would be able to act on the object since the purpose of `__init__` is to initialize objects of a class. I would assume a main use-case for `post_init` in an extension is to add validation checks to make sure instance variables are set to valid values and set default values for instance variables if they are not set, or even add new instance variables.

mavaylon1: Makes sense to me. I believe the request was phrased to operate on the kwargs dict, which is why I raised the question of scope.

oruebel: > I believe the request was phrased to operate on the kwargs dict

Acting on the `kwargs` I don't think is useful in a `post_init` setting, since at that point the `kwargs` have already been used in `__init__` to set instance variables.

rly: I agree - it's useful to act on `self` in `post_init`. I believe that is doable through monkeypatching with `cls.post_init_method = post_init_method` and that that binds the method to the class without making the class a classmethod, though I am not totally familiar with the use of bound vs unbound(?) methods.

```python
>>> class A:
...     def __init__(self, arg1):
...         self.arg1 = arg1
... 
>>> def modify_arg1(self):
...     self.arg1 = self.arg1 + 1
... 
>>> A.modify_arg1 = modify_arg1
>>> dir(A)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'modify_arg1']
>>> a = A(10)
>>> a.modify_arg1()
>>> a.arg1
11
>>> a.modify_arg1
<bound method modify_arg1 of <__main__.A object at 0x1007b76d0>>
>>> type(a.modify_arg1)
<class 'method'>
>>> import types
>>> type(a.modify_arg1) == types.MethodType
True
```

mavaylon1: I believe we are all on the same idea. I'll finalize this PR tomorrow.

mavaylon1: Bound method just means it is tied to an instance and the first argument will be self. Both of our ways do the same thing.

mavaylon1:
@rly I was wrapping this up with test on the MCIClassGenerator, adding post_init to the init method defined in set_init. What's weird is my test `test_multi_container_post_init` uses MCIClassGenerator (I verified by adding a breakpoint in the set_init for MCIClassGenerator), but the init that is called is the CustomClassGenerator (Verified by both the lack of coverage and adding a breakpoint in the init).

I see that MCIClassGenerator inherits from CustomClassGenerator, but the set_init function in MCIClassGenerator should override. It does (see above) but not in a way I expect. Thoughts?

rly:
In https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/build/classgenerator.py old line 334, the `__init__` generated by CustomClassGenerator is called on a subset of kwargs, which calls `post_init` . After that is done, the outer `MCIClassGenerator` calls `post_init` again. Does that answer your question?

mavaylon1:
> https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/build/classgenerator.py

I actually think its from old line 440. With MCIGenerator caling the init within CustomGenerator, do we only want to run the post init for MCIGenerator after the outer init (the init in MCI) and not the inner init?

rly:
You're right - I was looking at the wrong class there. I would say after the outer init.

mavaylon1:
@rly take a look when you can. I need figure out a way to solve the duplication problem in docval. I believe what is happening is that if we generate class A and class B (which inherits from A), it uses the same docval in the set_init method. This duplicates the post_init_bool because `*docval_args` is the docval from the parent class. Thoughts on the overall approach?

src/hdmf/build/classgenerator.py
```diff
@@ -316,8 +323,14 @@ def set_init(cls, classdict, bases, docval_args, not_inherited_fields, name):
             elif attr_name not in attrs_not_to_set:
                 attrs_to_set.append(attr_name)
 
-        @docval(*docval_args, allow_positional=AllowPositional.WARNING)
+        @docval(*docval_args,
+                {'name': 'post_init_bool', 'type': bool, 'default': True,
+                 'doc': 'bool to skip post_init for MCIClassGenerator'},
```
rly: I would rename this `skip_post_init` with default False.

rly:
I see. The approach feels a bit hacky but I cannot think of a better approach. One idea is that when getting the parent class' docval args, you can remove `post_init_bool` from the result.

mavaylon1:
> I see. The approach feels a bit hacky but I cannot think of a better approach. One idea is that when getting the parent class' docval args, you can remove `post_init_bool` from the result.

Yeah that makes sense. I'm going to push this idea forward and make update tests.

mavaylon1:
I will shoot to wrap this up tomorrow since it is very close and we have an agreed idea@rly

mavaylon1:
@rly This should be good to go. I am going to do another pass before I request review. However, I wanted to note that the tutorial on this functionality makes more sense to me to be in the extension template, rather than here. What do you think?

rly:
Let's document this in the NWB extensions tutorial rather than here https://nwb-overview.readthedocs.io/en/latest/extensions_tutorial/extensions_tutorial_home.html. We could do it in the template, but I would consider this more of an advanced feature.

mavaylon1:
1. The user would call `get_class`. This method now has a new input parameter called post_init_method. It is a Callable and it is a function that will be used as a post_init method to validate something during the class generation. It could also be more broad in its functionality.
2. `get_class` calls `get_dt_container_cls` from the the TypeMap. This calls `generate_class` within the class_generator. We add the post_init_method as an input parameter here as well. 
3. Within `generate_class` we do two things of interest. The first is we set the init method in the class. This is done from a class generator. This will be in either `CustomClassGenerator` or `MCIClassGenerator`. We simply call the post_init method at the end of the init method being set using the original kwargs. The second thing we do is set the post_init_method to the class via `cls.post_init_method = post_init_method`.

Let's talk about some details:
1. We can have classes that are generated where the parent class is also generated with a docval field that already has `skip_post_init` as an arg. We remove it before setting the docval of the new init. (I can't think of another way. It's not a bug, but just a product of generating classes)
2. Within `MCIClassGenerator`, we call `previous_init`. This means the __init__ generated by `CustomClassGenerator` is called on a subset of kwargs. We want to make sure that the post_init is not called here, but rather at the end of the init set within `MCIClassGenerator`. That's why we have a `skip_post_init`.

rly (approved):

rly:
Looks good to me. Thanks!

mavaylon1:
@h-mayorquin 
Hey Heberto this is the post_init method you were requesting. To use it, simply add a method into the `post_init_method` parameter in `get_class` (without the the parenthesis).

h-mayorquin:
Thanks, yes, I saw that you added, I will give it a test in the following weeks  : )

PR #1091, opened 2024-04-04
Don't override col_cls in DynamicTable.add_column

rly:
## Motivation
Issue: If a user creates a schema that extends `DynamicTableRegion` or `EnumData` and tries to create a column with that subclass as the column class, the column is instead created with the base `DynamicTableRegion` or `EnumData` class.

For example, if a user creates a new type `CustomDTR` that extends the `DynamicTableRegion` type, and then they create a `DynamicTable` and calls `add_column` with the column class being set to the new type (`col_cls=CustomDTR`), then the new column for this table currently has type `DynamicTableRegion`. 

```python
    test_table.add_column(
            name='custom_dtr_column',
            description='this is a custom DynamicTableRegion column',
            col_cls=self.CustomDTR,
            table=True,
        )
```

See new test: `test_add_custom_dtr_column`

Another use case: if a user creates a new type `CustomDTR` that extends the `DynamicTableRegion` type, and they create a new type `CustomTable` that extends `DynamicTable` and contains a column with type `CustomDTR`, then the auto-generated column for this table currently has type `DynamicTableRegion`. 

See test: `test_custom_dtr_class`

To make this work correctly in this PR, I updated `add_column` to use the passed `col_cls` instead of assuming `table=True` means `col_cls=DynamicTableRegion`. In order to do that, I also had to change the default value for `col_cls` away from `VectorData`. That default was set in more places than was needed anyway. I also had to move when the check that the passed `col_cls` equals the `class` included in a pre-defined column in `__columns__` occurs, so that it is checked after `col_cls` is appropriately set, not before. 

## Checklist


mavaylon1:
Will this be in by next milestone?

VisLab:
Please let us know when it gets in.  We are relying on it for implementation of ndx-hed.

mavaylon1 (approved):

PR #1093, opened 2024-04-07
VectorData Expand by Default via write_dataset

mavaylon1:
## Motivation

This change allows the new default behavior for writing VectorData data as expandandable datasets. We do this by providing maxshape to dataset settings that do not already have a defined maxshape set by the user. 

## How to test the behavior?
```
```
Testts
## Checklist


mavaylon1:
Notes and Questions:
1. Does scalar_fill imply that the dataset has only 1 value and should only have 1? There is no doc string. If so I can move the logic for max_shape into list_fill.
2. There are three cases with references where the shape is defined within required_dataset. I assume this is because get_data_shape returns a shape (#,1) where "#" is the number of references. This is something I found as an edge case. The quickest solution is to set maxshape at each of the three locations. The precedent being each 3 locations repeats shape being set. 
@oruebel

oruebel:
> 1. Does scalar_fill imply that the dataset has only 1 value and should only have 1?

Are you referring to 

https://github.com/hdmf-dev/hdmf/blob/d85d0cbc36d2e0fdb25e8fbea14d58ba7bf24a40/src/hdmf/backends/hdf5/h5tools.py#L1319

If so, this function is used to write scalar datasets, i.e., dataset with a single value. 

> 2\. There are three cases with references where the shape is defined within required_dataset.

Could you point to the case you are referring to? `require_dataset` is usually used to create a dataset it if it doesn’t exist and open the dataset if it does.

> 2\. The quickest solution is to set maxshape at each of the three locations.

This would mean to make all datasets expandable by enabling chunking for all datasets. That is a bit broader approach then to make this the default just for VectorData, but it would make it the default behavior for all (non-scalar) datasets. If that is the approach we'd want to take, then I would suggest adding a parameter `enable_chunking=True` on `HDFIO.write` and `HDFIO.export` so that we can configure the default behavior for write. @rly thoughts?

mavaylon1:
> > 1. Does scalar_fill imply that the dataset has only 1 value and should only have 1?
> 
> 
> 
> Are you referring to 
> 
> 
> 
> https://github.com/hdmf-dev/hdmf/blob/d85d0cbc36d2e0fdb25e8fbea14d58ba7bf24a40/src/hdmf/backends/hdf5/h5tools.py#L1319
> 
> 
> 
> If so, this function is used to write scalar datasets, i.e., dataset with a single value. 
> 
> 
> 
> > 2\. There are three cases with references where the shape is defined within required_dataset.
> 
> 
> 
> Could you point to the case you are referring to? `require_dataset` is usually used to create a dataset it if it doesn’t exist and open the dataset if it does.
> 
> 
> 
> > 2\. The quickest solution is to set maxshape at each of the three locations.
> 
> 
> 
> This would mean to make all datasets expandable by enabling chunking for all datasets. That is a bit broader approach then to make this the default just for VectorData, but it would make it the default behavior for all (non-scalar) datasets. If that is the approach we'd want to take, then I would suggest adding a parameter `enable_chunking=True` on `HDFIO.write` and `HDFIO.export` so that we can configure the default behavior for write. @rly thoughts?
> 
> 
> 
> 

The enable chunking parameter to give the user the option to turn off the expandable default? If so, would there be a reason they would want to?

oruebel:
> The enable chunking parameter to give the user the option to turn off the expandable default? If so, would there be a reason they would want to?

In my experience it is best to make choices explicit and provide useful defaults rather than hiding configurations.  A user may not want to use chunking if they want to use numpy memory mapping to read contiguous datasets.

mavaylon1:
@rly I will shoot to have this done by next week (formerly Friday May 3).

mavaylon1:
Dev Notes:
When writing datasets, we have a few options:
1. __setup_chunked_dset__: This is is for datachunkiterator so we do not need to mess with this
2. __scalar_fill__: scalar datasets do not need to be extended by default by nature of being scalar
3. __setup_empty_dset__: This is is for datachunkiterator so we do not need to mess with this
4. __list_fill__: Write a regular in memory array (e.g., numpy array, list etc.)

From my understanding we only need modify the input parameter options for only list_fill.

Now Oliver mentioned being more explicit with a switch `enable_chunking=True` (default will be true) on HDFIO.write and HDFIO.export so that we can configure the default behavior for write. this will need to be passed through the chain of methods from `write` and `export` to `write_dataset`.

oruebel:
> From my understanding we only need modify the input parameter options for only list_fill.
> 
> Now Oliver mentioned being more explicit with a switch `enable_chunking=True` (default will be true) on HDFIO.write and HDFIO.export so that we can configure the default behavior for write. this will need to be passed through the chain of methods from `write` and `export` to `write_dataset`.

Yes, I believe that is correct. I think only logic in `list_fill` should need to be modified and then the `enable_chunking` setting will need to be passed through. Note, `list_fill` already is being passed the argument `options` which contains `io_settings` so I think you may just need to set `chunks=True` in the `io_settings` (if `chunks` is set to None) to enable chunking. I'm not sure if it will be easiest to do this change if `io_settings` in `list_fill` or to update the `io_settings` outside of `list_fill` so that `list_fill` would not need to change at all.

https://github.com/hdmf-dev/hdmf/blob/126bdb100c6d5ce3e2dadd375de9d32524219404/src/hdmf/backends/hdf5/h5tools.py#L1432C5-L1438C53

mavaylon1:
Tests:
1. compound cases: added a check on existing test
2. regular array data: added a check on existing test
3. existing tests: updated three tests (refer to changed files)
4. Check test for when maxshape is already set, that my update does not interfere

mavaylon1:
@oruebel This is mostly done. I need to check/update or write a test that my changes does not do any interference of existing maxshape settings. (And do another pass through to make sure the logic is efficient) However, the main point I want to bring up is your idea of having a parameter for turning on and off the expandability. This would mean `HDMFIO` has a parameter that is not used in ZarrIO. In fact there are tests failing due to the parameter not being recognized. I see we have two options:
1. Update hdmf-zarr to accept and do nothing with the arg
2. Don't have the arg
I do like the explicit nature of having it, but I also think having this trickle into hdmf-zarr is not clean.

oruebel:
> This would mean HDMFIO has a parameter that is not used in ZarrIO. 

I don't think the parameter needs to be in HDMFIO. I think it's ok to just add as a parameter in HDF5IO

mavaylon1:
> > This would mean HDMFIO has a parameter that is not used in ZarrIO. 
> 
> I don't think the parameter needs to be in HDMFIO. I think it's ok to just add as a parameter in HDF5IO

HDF5IO write needs to call write_builder. It does that by calling super().write(**kwargs). This then gets us to HDMFIO write, which calls write_builder.

oruebel:
> HDF5IO write needs to call write_builder. It does that by calling super().write(**kwargs). This then gets us to HDMFIO write, which calls write_builder.

Yes, but `HDMFIO.write` allows extra keyword arguments:

https://github.com/hdmf-dev/hdmf/blob/126bdb100c6d5ce3e2dadd375de9d32524219404/src/hdmf/backends/io.py#L80-L81

and those are being passed through to `write_builder`

https://github.com/hdmf-dev/hdmf/blob/126bdb100c6d5ce3e2dadd375de9d32524219404/src/hdmf/backends/io.py#L99

So you can add custom keyword arguments without having to add them in `HDMFIO`. `HDF5IO` already has several additional arguments on `write` and `write_builder` that are not in `HDMFIO`, such as the `exhaust_dci` parameter.

mavaylon1:
> > HDF5IO write needs to call write_builder. It does that by calling super().write(**kwargs). This then gets us to HDMFIO write, which calls write_builder.
> 
> Yes, but `HDMFIO.write` allows extra keyword arguments:
> 
> https://github.com/hdmf-dev/hdmf/blob/126bdb100c6d5ce3e2dadd375de9d32524219404/src/hdmf/backends/io.py#L80-L81
> 
> and those are being passed through to `write_builder`
> 
> https://github.com/hdmf-dev/hdmf/blob/126bdb100c6d5ce3e2dadd375de9d32524219404/src/hdmf/backends/io.py#L99
> 
> So you can add custom keyword arguments without having to add them in `HDMFIO`. `HDF5IO` already has several additional arguments on `write` and `write_builder` that are not in `HDMFIO`, such as the `exhaust_dci` parameter.

Well isn't that just right in front of my face.

mavaylon1:
Notes:
My approach to the tests:
1. Make sure the default is now expandable. This is accomplished by adding an additional assert on existing test. This is done for both regular and compound data.
2. Make sure the default does not override user chunking. This is accomplished by existing tests passing.
3. Make sure when told "false" that it turns off said chunking.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Added `TypeConfigurator` to automatically wrap fields with `TermSetWrapper` according to a configuration file. @mavaylon1 [#1016](https://github.com/hdmf-dev/hdmf/pull/1016)
 - Updated `TermSetWrapper` to support validating a single field within a compound array. @mavaylon1 [#1061](https://github.com/hdmf-dev/hdmf/pull/1061)
 - Updated testing to not install in editable mode and not run `coverage` by default. @rly [#1107](https://github.com/hdmf-dev/hdmf/pull/1107)
+- Updated the new default behavior for writing VectorData as expandandable datasets. This does not override user set chunking parameters. @mavaylon1 [#1093](https://github.com/hdmf-dev/hdmf/pull/1093)
```
```suggestion
- Updated the default behavior for writing HDF5 datasets to be expandandable datasets with chunking enabled by default. This does not override user set chunking parameters. @mavaylon1 [#1093](https://github.com/hdmf-dev/hdmf/pull/1093)
```

oruebel: The implementation changes this for all datasets, not just VectorData (although VectorData was the motivation for this change)

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -364,7 +364,9 @@ def copy_file(self, **kwargs):
              'default': True},
             {'name': 'herd', 'type': 'hdmf.common.resources.HERD',
              'doc': 'A HERD object to populate with references.',
-             'default': None})
+             'default': None},
+            {'name': 'expandable', 'type': bool, 'default': True,
+             'doc': 'Bool to set whether datasets are expandable through chunking by default.'})
```
oruebel: Would be nice to make it more explicit in the docstring what is happening under the hood, maybe say     `'doc': 'Bool to set whether datasets are expandable by setting the max shape for all dimensions of a dataset to None and enabling auto-chunking by default.'})`

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -804,10 +806,15 @@ def close_linked_files(self):
              'doc': 'exhaust DataChunkIterators one at a time. If False, exhaust them concurrently',
              'default': True},
             {'name': 'export_source', 'type': str,
-             'doc': 'The source of the builders when exporting', 'default': None})
+             'doc': 'The source of the builders when exporting', 'default': None},
+            {'name': 'expandable', 'type': bool, 'default': True,
+             'doc': 'Bool to set whether datasets are expandable through chunking by default.'})
```
oruebel: Same comment here. Use the same doc for `expandable` as in `def write(...)`

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1451,7 +1463,15 @@ def __list_fill__(cls, parent, name, data, options=None):
...
-
+        if expandable:
+            if 'maxshape' not in io_settings:
+                io_settings['maxshape'] = tuple([None]*len(data_shape))
+            else:
+                # Don't override existing settings
+                pass
+        else:
+            msg = "Datasets written using user defined parameters. Default expandable via chunking: False"
+            warnings.warn(msg)
```
oruebel:
```suggestion
```

I don't think the `warning` is necessary, because it is only being raised when the user explicitly sets `expandable=False`, so the user explicitly made that choice. The `else: pass` part I don't think is needed either.

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -763,6 +766,29 @@ def test_read_str(self):
...
+
+    def test_expand_false(self):
+        # Setup all the data we need
+        foo1 = Foo('foo1', [1, 2, 3, 4, 5], "I am foo1", 17, 3.14)
+        foobucket = FooBucket('bucket1', [foo1])
+        foofile = FooFile(buckets=[foobucket])
+
+        with self.assertWarns(Warning):
+            with HDF5IO(self.path, manager=self.manager, mode='w') as io:
+                io.write(foofile, expandable=False)
```
oruebel:
```suggestion
```

If we remove the warning, then I think this check in the unit tests also needs to be removed.

oruebel (commented):
I added some minor suggestions, but otherwise this looks good to me.

src/hdmf/backends/hdf5/h5tools.py
mavaylon1: So the pass, was just for me to be explicit in my conditionals. I can remove it, but I think the note provides clarity for the future. 

The warning I think is a good reminder for now since this is a change to a default. Maybe we can remove it after this feature has been here for a while, but I think having some "flashing lights" is helpful.

I can remove these if you feel strongly about it.

mavaylon1:
> I added some minor suggestions, but otherwise this looks good to me.

Thanks for the quick review. I will make the doc string more detailed, but take a look at my comments for the other changes. The pass was a deliberate thing (vs a left over from a draft) and I like the warning.

src/hdmf/backends/hdf5/h5tools.py
mavaylon1: I'll remove the pass, but will keep the comment

mavaylon1: I do like the warning still

oruebel: > The warning I think is a good reminder for now since this is a change to a default.

But the warning is issued when the user is switching back to the previous default behavior. Also, this warning will be issued for every dataset in a file so this will result in a large number of warnings for a single file.  My vote this to remove the warning or at least move it to the `write` function so that it is issued only once. 

> I can remove it, but I think the note provides clarity for the future.

I would just leave a comment without having the extraneous `else` statement that doesn't do anything

mavaylon1: The large number of warnings for a single file is a good point. I'll make the changes and ping for review to merge.

oruebel (approved):
Looks good to me. Thanks!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Added `TypeConfigurator` to automatically wrap fields with `TermSetWrapper` according to a configuration file. @mavaylon1 [#1016](https://github.com/hdmf-dev/hdmf/pull/1016)
 - Updated `TermSetWrapper` to support validating a single field within a compound array. @mavaylon1 [#1061](https://github.com/hdmf-dev/hdmf/pull/1061)
 - Updated testing to not install in editable mode and not run `coverage` by default. @rly [#1107](https://github.com/hdmf-dev/hdmf/pull/1107)
+- Updated the default behavior for writing HDF5 datasets to be expandandable datasets with chunking enabled by default. This does not override user set chunking parameters. @mavaylon1 [#1093](https://github.com/hdmf-dev/hdmf/pull/1093)
```
rly, line 10: expandandable -> expandable

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1301,7 +1317,7 @@ def _filler():
                 self.__dci_queue.append(dataset=dset, data=data)
             # Write a regular in memory array (e.g., numpy array, list etc.)
             elif hasattr(data, '__len__'):
-                dset = self.__list_fill__(parent, name, data, options)
+                dset = self.__list_fill__(parent, name, data, expandable, options)
```
rly, line 1320: I think `expandable` should be passed to `__setup_empty_dset__` and `__setup_chunked_dset__` as well so that if a user is iteratively writing a `VectorData` or any other dataset, it can also be expanded.

rly: Unless these are already expandable. Is that the case?

mavaylon1: When I read the tutorial, DataChunkIterator already sets the None for maxshape since it has no way of knowing the minimum size of the array.

rly:
Could you add documentation on how to expand a VectorData?

It looks like creation of a dataset of references is not modified here. Some tables in NWB contain columns that are all references, e.g., the electrode table has a column with references to the ElectrodeGroup. I think such datasets should be expandable as well.

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -739,12 +742,12 @@ def test_copy_h5py_dataset_h5dataio_input(self):
                              self.f['test_copy'][:].tolist())
 
     def test_list_fill_empty(self):
-        dset = self.io.__list_fill__(self.f, 'empty_dataset', [], options={'dtype': int, 'io_settings': {}})
+        dset = self.io.__list_fill__(self.f, 'empty_dataset', [], True, options={'dtype': int, 'io_settings': {}})
```
rly, line 745: Can we add:
```
self.assertEqual(get_data_shape(dset), (None,))
```
to confirm that this dataset is expandable?

src/hdmf/backends/hdf5/h5tools.py
rly: Got it. It should be fine then. Thanks.

tests/unit/test_io_hdf5_h5tools.py
mavaylon1: We already test `list_fill` expandability with the other tests. We can add it here to, but I think it is redundant. (which is why I didn't add it).

rly: That's fair. Redundancy in the tests can be nice to catch unexpected edge cases, but in this case, after looking at the code again, I agree, it's not needed.

mavaylon1:
> Could you add documentation on how to expand a VectorData?
> 
> It looks like creation of a dataset of references is not modified here. Some tables in NWB contain columns that are all references, e.g., the electrode table has a column with references to the ElectrodeGroup. I think such datasets should be expandable as well.

Yeah the lack of dataset of references support was just a smaller scope for this idea. I agree this makes a lot of sense to have. I will make this an issue ticket.

As for the expansion of VectorData documentation, I thought we had that. Maybe I am thinking of the HDF5 documentation, but I will look. If it does not exist, I will loop that into the ticket for dataset of references.

PR #1094, opened 2024-04-08
Field_config refactored 

mavaylon1:
## Motivation

1. We want to use the PyNWB typemap when using the TypeConfigurator in pynwb. As a result we need to update the field setter to take in typemap. 
2. I cleaned up the field_config method.

We can override the get type map function on the pynwb side. 

## How to test the behavior?
```
```

## Checklist


stephprince (approved):
looks good to me!

PR #1100, opened 2024-04-19
improve html representation of datasets

h-mayorquin:
## Motivation

Improve the display of the data in the html representation of containers. Note that this PR is focused on datasets that were already written. For in memory representation the issue on what to do with things that are wrapped in an iterator or an `DataIO` subtype can be addressed in another PR I think. 

### HDF5
I have been using this script 
```python
from pynwb.testing.mock.ecephys import mock_ElectricalSeries
from hdmf.backends.hdf5.h5_utils import H5DataIO
from pynwb.testing.mock.ophys import mock_ImagingPlane, mock_TwoPhotonSeries

import numpy as np

data=np.random.rand(500_000, 384)
timestamps = np.arange(500_000)
data = data=H5DataIO(data=data, compression=True, chunks=True)

nwbfile = mock_NWBFile()
electrical_series = mock_ElectricalSeries(data=data, nwbfile=nwbfile, rate=None, timestamps=timestamps)

imaging_plane = mock_ImagingPlane(grid_spacing=[1.0, 1.0], nwbfile=nwbfile)


data = H5DataIO(data=np.random.rand(2, 2, 2), compression=True, chunks=True)
two_photon_series = mock_TwoPhotonSeries(name="TwoPhotonSeries", imaging_plane=imaging_plane, data=data, nwbfile=nwbfile)


# Write to file
from pynwb import NWBHDF5IO
with NWBHDF5IO('ecephys_tutorial.nwb', 'w') as io:
    io.write(nwbfile)



from pynwb import NWBHDF5IO

io = NWBHDF5IO('ecephys_tutorial.nwb', 'r')
nwbfile = io.read()
nwbfile
```

image

### Zarr

```python
from numcodecs import Blosc
from hdmf_zarr import ZarrDataIO
import numpy as np
from hdmf_zarr.nwb import NWBZarrIO
import os
import zarr
from numcodecs import Blosc, Delta

from pynwb.testing.mock.ecephys import mock_ElectricalSeries
filters = [Delta(dtype="i4")]

data_with_zarr_data_io = ZarrDataIO(
    data=np.arange(100000000, dtype='i4').reshape(10000, 10000),
    chunks=(1000, 1000),
    compressor=Blosc(cname='zstd', clevel=3, shuffle=Blosc.SHUFFLE),
    # filters=filters,
)

timestamps = np.arange(10000)

data = data_with_zarr_data_io

nwbfile = mock_NWBFile()
electrical_series_name = "ElectricalSeries"
rate = None
electrical_series = mock_ElectricalSeries(name=electrical_series_name, data=data, nwbfile=nwbfile, timestamps=timestamps, rate=None)


path = "zarr_test.nwb.zarr"
absolute_path = os.path.abspath(path)
with NWBZarrIO(path=path, mode="w") as io:
    io.write(nwbfile)
    
from hdmf_zarr.nwb import NWBZarrIO

path = "zarr_test.nwb.zarr"

io = NWBZarrIO(path=path, mode="r")
nwbfile = io.read()
nwbfile


```

image

## Checklist


src/hdmf/container.py
```diff
@@ -728,18 +730,22 @@ def _generate_field_html(self, key, value, level, access_code):
...
+            html_content = self._generate_array_html(value, level + 1)
+        elif hasattr(value, "store") and hasattr(value, "shape"):  # Duck typing for zarr array
+            html_content = self._generate_array_html(value, level + 1)
+        elif hasattr(value, "generate_html_repr"):
             html_content = value.generate_html_repr(level + 1, access_code)
-
         elif hasattr(value, '__repr_html__'):
             html_content = value.__repr_html__()
-
         elif hasattr(value, "fields"):
```
h-mayorquin: I think we maybe should add a comment here. I think this branching was intended for Containers that have fields but it also captures h5py.Datasets because they have a method called fields. I am solving this by just having the branching to catch h5py.Datasets before this but it seems brittle to rely on order of execution.

src/hdmf/container.py
```diff
@@ -728,18 +730,22 @@ def _generate_field_html(self, key, value, level, access_code):
...
 
-        if hasattr(value, "generate_html_repr"):
+        if isinstance(value, (np.ndarray, h5py.Dataset)):
+            html_content = self._generate_array_html(value, level + 1)
+        elif hasattr(value, "store") and hasattr(value, "shape"):  # Duck typing for zarr array
+            html_content = self._generate_array_html(value, level + 1)
+        elif hasattr(value, "generate_html_repr"):
             html_content = value.generate_html_repr(level + 1, access_code)
-
         elif hasattr(value, '__repr_html__'):
```
h-mayorquin, line 733: I also think that maybe that was intended to be `_repr_html_` (single underscore) instead of the double underscore version that is written here?

Single underscore is what I found in the ipython documentation:

https://ipython.readthedocs.io/en/stable/config/integrating.html

src/hdmf/container.py
stephprince: I agree a comment there would be helpful to note that h5py.Datasets also have the fields method.

src/hdmf/container.py
stephprince: I couldn't find any examples in the code that have the `__repr_html__` with the double underscores so I'm also not sure what the original intention was or where this might be called.

However if we replace this with single underscores `_repr_html_`, other Container objects in the file will display like this with a new header and without the correct margin-left spacing, so I would lean towards leaving for now.
Screenshot 2024-04-24 at 10 36 53 AM

src/hdmf/container.py
```diff
@@ -749,10 +755,58 @@ def _generate_field_html(self, key, value, level, access_code):
 
         return html_repr
 
+
     def _generate_array_html(self, array, level):
         """Generates HTML for a NumPy array."""
-        str_ = str(array).replace("\n", "</br>")
-        return f'<div style="margin-left: {level * 20}px;" class="container-fields">{str_}</div>'
+
+        # This is a placeholder function, you should define your own conversion
```
stephprince: Could you clarify this part, when/where should you define your own conversion?

stephprince (commented):
This looks great! Thanks for the PR.

Could you add tests for the data html representation with hdf5 and zarr? I think we mainly have string equivalence tests for this kind of thing.

I'm also wondering if it would be nice to have the hdf5 dataset info displayed in a similar table format as the zarr arrays to make it more consistent across backends. I think we should be able to replicate this using the hdf5 dataset info as an input to a method like this: https://github.com/zarr-developers/zarr-python/blob/9d046ea0d2878af7d15b3de3ec3036fe31661340/zarr/util.py#L402

src/hdmf/container.py
h-mayorquin: Myserious. OK, we can figure this out in another PR.

src/hdmf/container.py
h-mayorquin: Outdated comments I wll remove.

src/hdmf/container.py
h-mayorquin: Done.

h-mayorquin:
OK, I added table formating for hdf5:

image

h-mayorquin:
@stephprince 
Concerning the test, yes, I can do it, but, can you helmp to create a container that contains array data? I just don't have experienced with the bare bones object. This is my attempt:

```python
from hdmf.container import Container

container = Container(name="Container")
container.__fields__ = {
    "name": "data",
    "description": "test data",
}

test_data = np.array([1, 2, 3, 4, 5])
setattr(container, "data", test_data)
container.fields
```

But the data is not added as a field. How can I move forward?

h-mayorquin:
Related:

https://github.com/hdmf-dev/hdmf-zarr/issues/186

src/hdmf/container.py
h-mayorquin: Done .

tests/unit/test_container.py
```diff
@@ -456,6 +475,74 @@ def test_repr_html_(self):
         )
 
 
+    def test_repr_html_hdf5_dataset(self):
```
h-mayorquin, line 496: @stephprince I added a test.

stephprince: I see what you meant about the missing data field. I'll take a look at how to set up the tests

h-mayorquin: The current example sort of works, is there something missing?

h-mayorquin:
I added the handling division by zero, check this out what happens with external files (like Video):

image


From this example:

```python
import remfile
import h5py

asset_path = "sub-CSHL049/sub-CSHL049_ses-c99d53e6-c317-4c53-99ba-070b26673ac4_behavior+ecephys+image.nwb"
recording_asset = dandiset.get_asset_by_path(path=asset_path)
url = recording_asset.get_content_url(follow_redirects=True, strip_query=True)
file_path = url

rfile = remfile.File(file_path)
file = h5py.File(rfile, 'r')

from pynwb import NWBHDF5IO

io = NWBHDF5IO(file=file, mode='r')

nwbfile = io.read()
nwbfile
```

tests/unit/test_container.py
stephprince: ah you're right it was working fine, I was just getting a test failure from the docval data type in the `ContainerWithData` initialization so I changed it to be the more general "array_data" type.

I also added a test for numpy arrays. Let me know if those look ok to you.

stephprince:
There are still some failing tests for different python versions, it looks like one of the reasons is because h5py added the `.nbytes` attribute in version 3.0 and we still have `h5py==2.10` as a minimum version.

```
>       array_size_in_bytes = array.nbytes
E       AttributeError: 'Dataset' object has no attribute 'nbytes'
```

I'm not sure if there's another way to access that information or if we would just want to optionally display it if available.

oruebel:
> I'm not sure if there's another way to access that information or if we would just want to optionally display it if available.

Checking `if hasattr(data, "nbytes")` to optionally seems reasonable to me. In this way you can also avoid custom behavior depending on library versions.

src/hdmf/container.py
```diff
@@ -749,10 +755,71 @@ def _generate_field_html(self, key, value, level, access_code):
...
+
+            head = "HDF5 Dataset"
+            hdf5_info_dict = {"chunks": chunks, "compression": compression, "compression_opts": compression_opts,
+                              "compression_ratio": compression_ratio}
+            backend_info_dict = {**basic_array_info_dict, **hdf5_info_dict}
+
+        if hasattr(array, "store") and hasattr(array, "shape"):  # Duck typing for zarr array
+            head = "Zarr Array"
+            zarr_info_dict = {k:v for k, v in array.info_items()}
+            backend_info_dict = zarr_info_dict
```
oruebel: It would be nice to avoid having logic that is specific for a particular io backend in the `Container`. The reason is that this inhibts implementing backends in a self-contained manner and stand-alone packages and requires updating many places throughout HDMF. 

The checks for `h5py.Dataset` and `Zarr.array` are really only relevant when a file was read from file. To help disentangle the dependencies, I'm wondering whether we could do the following:

1. Add a static method ``generate_dataset_html`` to ``HDMFIO`` that would then need to implemented by ``HDF5IO`` and ``ZarrIO``
2. In the ``Container`` you could then do something like:

```
read_io = self.get_read_io()  # if the Container was read from file, this will give you the IO object that read it
if read_io is not None:
    html_repr = read_io.generate_dataset_html(my_dataset)
else:
    # The file was not read from disk so the dataset should be numpy array or a list
```



`

h-mayorquin:
@stephprince 
> I'm not sure if there's another way to access that information or if we would just want to optionally display it if available.

It can be estimated from the dtype and the number of elements. I will do that when the attribute does not exists.

src/hdmf/container.py
h-mayorquin: > It would be nice to avoid having logic that is specific for a particular io backend in the Container. The reason is that this inhibts implementing backends in a self-contained manner and stand-alone packages and requires updating many places throughout HDMF.

I see, yes, it would be nice if we could decouple this.  On the other hand, right now, if they do implement their own backend they will just lose the representation for datasets which is not critical. 

> The checks for h5py.Dataset and Zarr.array are really only relevant when a file was read from file.

Is it? Right now in the test we are passsing an hdf5 dataset as data without writing the nwbfile for testin the display. Is this not intended?

This proposal seems very good:
```
read_io = self.get_read_io()  # if the Container was read from file, this will give you the IO object that read it
if read_io is not None:
    html_repr = read_io.generate_dataset_html(my_dataset)
else:
    # The file was not read from disk so the dataset should be numpy array or a list
```

I see two downsides:
1) Missing extensive representation for in-memory files (it is nice to know what you will write!).
2) Fragmenting the code base for html representations.

Is there any other backend in the works right now? If not, maybe we can do this simpler way and add the complexity once we are closer to need it?

src/hdmf/container.py
```diff
@@ -794,7 +794,10 @@ def html_table(item_dicts) -> str:
             hdf5_dataset = array
             chunks = hdf5_dataset.chunks
             compression = hdf5_dataset.compression
-            uncompressed_size = hdf5_dataset.nbytes
+            if hasattr(hdf5_dataset, "nbytes"):  # TODO, remove if statement when and if h5py 3.0 is minimal
```
h-mayorquin: @stephprince 
Something like this should work I think.

tests/unit/test_container.py
h-mayorquin: The current tests look godo to me, thanks.

src/hdmf/container.py
h-mayorquin: But needs to be higher up. I need to check how to run those tests locally : )

stephprince: Looks good! Thanks!

src/hdmf/container.py
oruebel: I think it would worth giving this a try here. If it works for HDF5, we can then we can easily move the logic for Zarr to hdmf-zarr. I don't think it should be too hard to make this work right now, but these things tend to get hard to change later on.

h-mayorquin: What about handling in-memory objects as well?

oruebel: In-memory-only objects (i.e., numpy arrays and lists) can be handled here in the Container class.

h-mayorquin: But you can have in-memory h5py datasets and zarr arrays, right? Am I wrong about this? I pointed out above that that is how we are testing this.

Ok, I think that the easiest thing is to restrict this PR to just displaying information that is encoded in the array standard:

https://data-apis.org/array-api/latest/API_specification/array_object.html#attributes

So no backend specific information like compression options, chunks, etc here. We can then follow your approach in another PR to add backend related information extracted through the `DataIO` objects.

What do you think?

oruebel: > But you can have in-memory h5py datasets and zarr arrays, right?

As far as I know, h5py datasets are always tied to a file on disk (and I believe in Zarr as well). There are ways (at least in Zarr) to have the whole file then live in memory only, but the datasets are still tied to a file that is being read from. 

> So no backend specific information like compression options, chunks, etc here. We can then follow your approach in another PR to add backend related information extracted through the `DataIO` objects.

Sure, that sounds good to me.

h-mayorquin: > As far as I know, h5py datasets are always tied to a file on disk (and I believe in Zarr as well). There are ways (at least in Zarr) to have the whole file then live in memory only, but the datasets are still tied to a file that is being read from.

What I was asking is that if for those examples (like the one in the test here) the trick of using the `DataIO` would work. I think not but maybe I am missing something.

oruebel: > We can then follow your approach in another PR to add backend related information extracted through the `DataIO` objects.

Could you point me to the PR you are referring to. I'm not sure what role `DataIO` plays for this PR.

h-mayorquin: I mean that we can implement the following strategy in another PR to add backend specific information:

```python
read_io = self.get_read_io()  # if the Container was read from file, this will give you the IO object that read it
if read_io is not None:
    html_repr = read_io.generate_dataset_html(my_dataset)
else:
    # The file was not read from disk so the dataset should be numpy array or a list
```

oruebel: Yes that looks good 👍 . Just to avoid confusion, `read_io` is an instance of `HDMFIO` (i.e., `HDF5IO` or `ZarrIO`) and not `DataIO`. To implement the logic we would then need to:

1. Add `HDMFIO.generate_dataset_html(dataset)` which would implement just a minimalist representation
2. Implement `HDF5IO.generate_dataset_html(h5py.Dataset)` to represent `h5py.Dataset`
3. In a separate PR on `hdmf_zarr` implement `ZarrIO.generate_dataset_html(Zarr.array)`

To simplify this implementation and generate consistent representations, we could make a utility function that would take information about a dataset (e.g,. name, shape, data type, etc.) as input and generate the html representation such that the individual `generate_data_html` on the I/O backends would just collect the information from the dataset and use the utility function to generate the actual html.

h-mayorquin: > Yes that looks good 👍 . Just to avoid confusion, read_io is an instance of HDMFIO (i.e., HDF5IO or ZarrIO) and not DataIO. To implement the logic we would then need to.

Yes, I realized after that I was confusing these two objects.

oruebel: Sounds good. Thanks for your hard work on this PR and the fruitful discussion.

rly:
@stephprince when you have time, can you review this?

stephprince:
Rereading through this discussion, I believe where we left off is that the we want to remove the backend-specific logic from the `Container` class. To do so, it was proposed that:

In this PR we:

- Add `HDMFIO.generate_dataset_html(dataset)` which would implement a minimalist representation
- Implement `HDF5IO.generate_dataset_html(h5py.Dataset)` to represent an `h5py.Dataset`

In a separate PR on `hdmf_zarr` we would:
- implement `ZarrIO.generate_dataset_html(Zarr.array)`

In the `Container` class, it would look like this:
```python
read_io = self.get_read_io()  # if the Container was read from file, this will give you the IO object that read it
if read_io is not None:
    html_repr = read_io.generate_dataset_html(my_dataset)
else:
    # The file was not read from disk so the dataset should be numpy array or a list
```

@h-mayorquin did you want to do this? Otherwise I can go ahead and make the proposed changes to finish up this PR.

h-mayorquin:
Hi, @stephprince

I think this is a good summary.  

I am not sure how to decouple HDF5IO.generate_dataset_html(h5py.Dataset) here as hdmf seems super coupled with hdf5. Or is it the idea that we only want to exclude zarr? 

This has been on the back of my mind for a while and everytime but I had other priorities. It would be great if you have time to finish it.

stephprince:
@h-mayorquin yes I can take a look at it and finish it up

stephprince:
I have pushed the updates we discussed:
- added utility functions `generate_array_html_repr` and `get_basic_array_info` to the utils module to get basic array info and generate an array html table
- added a static `HDMFIO.generate_dataset_html()` method, the HDF5/Zarr implementations collect information from the dataset and then generate the actual html representation
- updated  `Container._generate_array_html()`  to use these methods

I tested a Zarr implementation that looks like this and can submit a PR in hdmf_zarr for that:
```python
def generate_dataset_html(dataset):
    """Generates an html representation for a dataset for the ZarrIO class"""

    # get info from zarr array and generate html repr
    zarr_info_dict = {k:v for k, v in dataset.info_items()}
    repr_html = generate_array_html_repr(zarr_info_dict, dataset, "Zarr Array")

    return repr_html
```

@oruebel @h-mayorquin if you could please review and let me know if there are any remaining concerns

h-mayorquin:
Looks good to me, thanks for taking on this.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1603,3 +1604,29 @@ def set_dataio(cls, **kwargs):
             data = H5DataIO(data)
         """
         return H5DataIO.__init__(**kwargs)
+
+    @staticmethod
+    def generate_dataset_html(dataset):
+        """Generates an html representation for a dataset for the HDF5IO class"""
```
oruebel, line 1610: Would be useful to raise a corresponding issue on HDMF_ZARR to have `generate_dataset_html` be implemented on `ZarrIO` as well (if we have not done this yet). @stephprince can you make and issue

src/hdmf/backends/io.py
```diff
@@ -188,6 +188,14 @@ def close(self):
         ''' Close this HDMFIO object to further reading/writing'''
         pass
 
+    @staticmethod
+    def generate_dataset_html(dataset):
```
oruebel, line 192: I believe this function would be triggered when using `ZarrIO`. Did we test that this indeed works with `ZarrIO`?

oruebel (dismissed):

src/hdmf/backends/io.py
stephprince: @oruebel as we discussed in person, I tested locally that this function was triggered with `ZarrIO` and added a test for the `HDMFIO` method since `ZarrIO` will (hopefully) soon have its own implementation.

oruebel (approved):

PR #1110, opened 2024-05-06
Expose progress bar class control

CodyCBakerPhD:
## Motivation

For @garretmflynn and the GUIDE project

## How to test the behavior?

I will add a minimal test with an in-place extension of `tqdm`

## Checklist


src/hdmf/data_utils.py
```diff
@@ -179,6 +179,12 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             doc="Display a progress bar with iteration rate and estimated completion time.",
             default=False,
         ),
+        dict(
+            name="progress_bar_class",
+            type=callable,
```
CodyCBakerPhD: I have no idea if docval is OK with this?

garrettmflynn: Nah it's giving an error.
```
[2024-05-13 14:06:09,926] ERROR in app: Exception on /startup/preload-imports [GET]
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/utils.py", line 605, in dec
    a['type'] = __resolve_type(a['type'])
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/utils.py", line 527, in __resolve_type
    raise ValueError(msg)
ValueError: argtype must be a type, a str, a list, a tuple, or None - got <class 'builtin_function_or_method'>

  File "/Users/garrettflynn/Documents/GitHub/nwb-guide/pyflask/apis/startup.py", line 40, in get
    import neuroconv
  File "/Users/garrettflynn/Documents/GitHub/neuroconv/src/neuroconv/__init__.py", line 1, in <module>
    from .basedatainterface import BaseDataInterface
  File "/Users/garrettflynn/Documents/GitHub/neuroconv/src/neuroconv/basedatainterface.py", line 9, in <module>
    from pynwb import NWBFile
  File "/Users/garrettflynn/miniconda3/envs/nwb-guide/lib/python3.9/site-packages/pynwb/__init__.py", line 9, in <module>
    from hdmf.spec import NamespaceCatalog
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/__init__.py", line 2, in <module>
    from .backends.hdf5.h5_utils import H5Dataset, H5RegionSlicer
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/backends/__init__.py", line 1, in <module>
    from . import hdf5
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/backends/hdf5/__init__.py", line 1, in <module>
    from . import h5_utils, h5tools
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/backends/hdf5/h5_utils.py", line 20, in <module>
    from ...data_utils import DataIO, AbstractDataChunkIterator
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/data_utils.py", line 140, in <module>
    class GenericDataChunkIterator(AbstractDataChunkIterator):
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/data_utils.py", line 197, in GenericDataChunkIterator
    def __init__(self, **kwargs):
  File "/Users/garrettflynn/Documents/GitHub/hdmf/src/hdmf/utils.py", line 608, in dec
    raise Exception(msg)
Exception: docval for progress_bar_class: error parsing argument type: argtype must be a type, a str, a list, a tuple, or None - got <class 'builtin_function_or_method'>

```

Is there a simple way to fix this?

CodyCBakerPhD: @rly @oruebel I want to specify a `callable` in a function/class wrapped by docval

Is there any way to do that or otherwise circumvent it?

oruebel: The problem is that `callable` is not a type in Python but rather a built-in function. If you want to restrict to progress bar classes from tqdm then you could set the `type` to `tqdm.std`, which I believe is the common base class for progress bars in `tqdm`. Or you could make a list of all the progress bar types you want to allow.  I believe `docval` allows this to be specified as a string `tqdm.std` so that you should not need to import `tqdm` in order to specify this for docval. Alternatively, if you want any class to pass validation, then I believe you could set `type` to None.

src/hdmf/data_utils.py
```diff
@@ -179,6 +179,12 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
...
+            doc="The progress bar class to use. Defaults to tqdm.tqdm if the TQDM package is installed.",
+            default=None,
+        ),
         dict(
             name="progress_bar_options",
             type=None,
```
oruebel: The comment for `progress_bar_options` says this should be dict but the `type` says `None`. Should this be `type=dict` instead?

src/hdmf/data_utils.py
CodyCBakerPhD: `tqdm.std` is the module, not a parent class; `tqdm.tqdm` is the thing they recommend using as a base class when extending, but in general you could use any generic iterable wrapper if you wanted to 

I'll just use the override for now

src/hdmf/data_utils.py
CodyCBakerPhD: Tried to update this, will see how it goes

CodyCBakerPhD:
@garrettmflynn Should be good to try again

src/hdmf/data_utils.py
```diff
@@ -179,9 +179,15 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             doc="Display a progress bar with iteration rate and estimated completion time.",
             default=False,
         ),
+        dict(
+            name="progress_bar_class",
+            type=None,  # Would prefer callable, but docval can't handle
```
rly: I think this should work:
```suggestion
            type=Callable,
```

with
```python
from collections.abc import Callable
```

Test:
```python
from collections.abc import Callable
from hdmf.utils import docval
@docval(dict(name="a", type=Callable, doc="a"), is_method=False)
def func(**kwargs):
    print(kwargs)
def func2():
    print("hi")
func(func2)  # returns {'a': <function func2 at 0x103716e80>}
from tqdm import tqdm
func(tqdm)  # returns {'a': <class 'tqdm.std.tqdm'>}
```

CodyCBakerPhD: Thanks @rly

src/hdmf/data_utils.py
```diff
@@ -179,9 +179,15 @@ class GenericDataChunkIterator(AbstractDataChunkIterator):
             doc="Display a progress bar with iteration rate and estimated completion time.",
             default=False,
         ),
+        dict(
+            name="progress_bar_class",
+            type=Callable,
```
rly, line 184: For whatever reason, `typing.Callable` is a deprecated alias to `collections.abc.Callable`:
https://docs.python.org/3/library/typing.html#typing.Callable

CodyCBakerPhD: Probably because the more modern type is, simply, `callable`

garrettmflynn: Just FYI this did not seem to work in practice when triggered with the GUIDE. Just opened https://github.com/catalystneuro/neuroconv/pull/861 so that all the pieces are in place to see the progress of a full conversion.

garrettmflynn: Nevermind this is working fine, as shown on https://github.com/NeurodataWithoutBorders/nwb-guide/pull/778#issuecomment-2113243480

CodyCBakerPhD:
@oruebel @rly This is ready for review

rly (approved):

PR #1115, opened 2024-05-15
Update pyproject.toml to use only Numpy <2.0

mavaylon1:
## Motivation

Numpy has released numpy 2.0, which came with deprecations. Even though python 3.9 and above is supported, there are deprecations that need to be updated in HDMF for numpy 2.0.
example: np.string_ is called in the objectmapper but that is now deprecated.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
Review Notes:

This is what we did for hdmf-zarr

stephprince (approved):

Issue #1079, opened 2024-03-25
[Bug]: Tests with optional requirements failing


rly:
### What happened?

See https://github.com/hdmf-dev/hdmf/actions/runs/8415368500

@mavaylon1 can you look into this? Seems to be an issue with pydantic versioning from updating pydantic in the last release

### Steps to Reproduce

```python
n/a
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.11

mavaylon1:
Can do

rly:
@mavaylon1 Several daily tests are now failing: https://github.com/hdmf-dev/hdmf/actions/runs/8811335853

Could you take a look into this? Some failures are due to the fact that GitHub recently updated the macos-latest environment to `macos-14-arm64`, and Python 3.9 and earlier aren't available there. You can set the `os` variable for those tests to `macos-13`

mavaylon1:
> @mavaylon1 Several daily tests are now failing: https://github.com/hdmf-dev/hdmf/actions/runs/8811335853
> 
> 
> 
> Could you take a look into this? Some failures are due to the fact that GitHub recently updated the macos-latest environment to `macos-14-arm64`, and Python 3.9 and earlier aren't available there. You can set the `os` variable for those tests to `macos-13` 

I'll take a look on Monday. This week is dedicated to other projects.

PR #1116, opened 2024-05-16
Fix warning

mavaylon1:
## Motivation

Fixed the warning regarding the TermSetWrapper being already wrapped. This shouldn't be raised when a configuration is not loaded. 

## How to test the behavior?
```
```

## Checklist


mavaylon1:
Review Notes: 
This is a possible fix to https://github.com/hdmf-dev/hdmf/issues/1079. However, the pydantic error is not showing up anymore. I am going to wait to see the output of manually running run_all_tests

Update: It seems to work.

stephprince (commented):
The container changes look good to me. What was the reason for removing the `.yaml` files?

mavaylon1:
> The container changes look good to me. What was the reason for removing the `.yaml` files?

They don't really need to be there since they can be generated when running the documentation.

stephprince:
Got it, would it make sense to add them to the `.gitignore` then if they are generated by the docs?

mavaylon1:
> Got it, would it make sense to add them to the `.gitignore` then if they are generated by the docs?

Done. The failing hdmf-zarr test is also on schedule and seems to be unrelated to this PR.

stephprince (approved):

PR #1135, opened 2024-06-27
Append a Dataset of References

mavaylon1:
## Motivation

## How to test the behavior?
```
```

## Checklist


docs/source/install_users.rst
```diff
@@ -29,4 +29,4 @@ You can also install HDMF using ``conda`` by running the following command in a
 
    conda install -c conda-forge hdmf
 
-.. _Anaconda Distribution: https://www.anaconda.com/products/distribution
+.. _Anaconda Distribution: https://www.anaconda.com/products
```
rly:
```suggestion
.. _Anaconda Distribution: https://www.anaconda.com/download
```

docs/source/install_developers.rst
```diff
@@ -73,7 +73,7 @@ environment by using the ``conda remove --name hdmf-venv --all`` command.
     For advanced users, we recommend using Mambaforge_, a faster version of the conda package manager
     that includes conda-forge as a default channel.
 
-.. _Anaconda: https://www.anaconda.com/products/distribution
+.. _Anaconda: https://www.anaconda.com
```
rly:
```suggestion
.. _Anaconda: https://www.anaconda.com/download
```

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -2958,6 +2958,27 @@ def test_append_data(self):
...
+        with HDF5IO(self.paths[0], manager=get_baz_buildmanager(), mode='w') as write_io:
+            write_io.write(bucket)
+
+        with HDF5IO(self.paths[0], manager=get_baz_buildmanager(), mode='a') as read_io:
+            read_bucket1 = read_io.read()
+            DoR = read_bucket1.baz_data.data
+            ref = read_bucket1.baz_data.data.dataset[0]
+            DoR.append(ref)
+            self.assertEqual(len(read_bucket1.baz_data.data), 11)
+
```
rly: Just in case, could you test closing this file, opening it, and making sure the size is still 11?

rly:
Minor suggestion to a test. Looks good otherwise.

tests/unit/test_io_hdf5_h5tools.py
mavaylon1: So call write again then read it?

rly: The write *should not* be needed. Append should have acted directly on the file, just like changing the values of the dataset. But if it doesn't work, then that's odd and something we would need to look into or document.

rly:
I added a test that raises an unexpected error:

```
self = <Closed HDF5 file>, name = <HDF5 object reference (null)>

    @with_phil
    def __getitem__(self, name):
        """ Open an object in the file """
    
        if isinstance(name, h5r.Reference):
            oid = h5r.dereference(name, self.id)
            if oid is None:
>               raise ValueError("Invalid HDF5 object reference")
E               ValueError: Invalid HDF5 object reference
```
We just chatted in person, but just to note that you were going to take a look at it.

mavaylon1:
@rly There may be a work around, but I think the problem below might be enough to just start the proxy idea.

```
def append(self, arg):
        child = arg
        while True:
            if child.parent is not None:
                parent = child.parent
                child = parent
            else:
                parent = child
                break
        self.io.manager.build(parent)
        builder = self.io.manager.build(arg)

        # Create Reference
        ref = self.io._create_ref(builder)
        append_data(self.dataset, ref)
```

When a user calls append on a reference, we build the root builder first. We then call `_create_ref`, which will try to create a reference `return self.__file[path].ref`. This fails with `KeyError: "Unable to open object "new"`. This fails because it is trying to create a reference to an object, i.e., the new baz, within the file; however, it is not in the file. It is in the root builder. 

Why isn't in the file? I am in append mode right? Let's ignore the reference for now, and just add the new baz. It works (sort of). When you read it back, the new baz is not there. We need to call write again. Once you do that it is there. That means when we try to create a reference, and it is looking for the new baz in the file only to find nothing because it is not added till write (which we never call during append). 

Earlier I said in conversation that you do not need to call write. That is half true. in my method prior (seen below), you do not need to call write to append to a dataset of references, but you do need to call write to add a new baz because it itself is a new group.  

Earlier I had 
```
def append(self, arg):
        # Get Builder
        builder = self.io.manager.build(arg)

        # Get HDF5 Reference
        ref = self.io._create_ref(builder)
        append_data(self.dataset, ref)
```

This leads to a reference being created, but not found with the test `self.assertIs(read_bucket1.baz_data.data[10], read_bucket1.bazs["new"])`. This is because the reference path is just `\`. This is wrong. It needs to be `'/bazs/new'`.

Note: yes this is the same code from hdmf-zarr. I started to wonder if this could just be in hdmf because the append calls _create_ref which means all we need to do is have unique `create_ref` methods per backend. AKA we wouldn't need a zarr PR that has this logic, just some name changes probably.

rly:
I see. Tricky indeed. You can't create an HDF5 reference to an object that isn't in the file yet, and rebuilding the whole hierarchy on each append is not ideal. A proxy makes sense. I can't think of another workaround without severely limiting and documenting the ways in which you cannot append. 

> Note: yes this is the same code from hdmf-zarr. I started to wonder if this could just be in hdmf because the append calls _create_ref which means all we need to do is have unique create_ref methods per backend. AKA we wouldn't need a zarr PR that has this logic, just some name changes probably.

That sounds useful to look into. You may be able to refactor it and some fields into the base `HDMFDataset` class.

mavaylon1:
Add documentation here: https://github.com/NeurodataWithoutBorders/pynwb/issues/1951

src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -108,6 +108,14 @@ def ref(self):
     def shape(self):
         return self.dataset.shape
 
+    def append(self, arg):
+        # Get Builder
+        builder = self.io.manager.get_builder(arg)
```
rly, line 113:
```suggestion
        builder = self.io.manager.get_builder(arg)
        if builder is None:
            raise ValueError(
                "The container being appended to the dataset has not yet been built. "
                "Please write the container to the file, then open the modified file, and "
                "append the read container to the dataset."
            )
```

rly: adding a helpful error message for the user. feel free to modify

src/hdmf/query.py
```diff
@@ -163,6 +163,12 @@ def __next__(self):
     def next(self):
         return self.dataset.next()
 
+    def append(self, arg):
+        """
+        Override this method for new backends, i.e., not HDF5 and Zarr.
```
rly:
```suggestion
        Override this method to support appending to backend-specific datasets
```

rly (approved):

PR #1136, opened 2024-06-27
Zarr append for datasets (non-reference)

mavaylon1:
## Motivation

This supports appending to zarr arrays. This does not cover the edge case for dataset of references. This has the base tests for append_data. There are more tests (TBD) in hdmf-zarr that incorporate the full IO process. 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/validate/validator.py
+++ b/src/hdmf/validate/validator.py
@@ -164,7 +164,7 @@ def get_type(data, builder_dtype=None):
                 # Empty array
                 else:
                     # Empty string array
-                    if isinstance(data.dtype.metadata["vlen"], str):
+                    if data.dtype.metadata["vlen"] is str:
                         return "utf", None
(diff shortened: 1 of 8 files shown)


src/hdmf/validate/validator.py
```diff
@@ -164,7 +164,7 @@ def get_type(data, builder_dtype=None):
                 # Empty array
                 else:
                     # Empty string array
-                    if data.dtype.metadata["vlen"] == str:
+                    if isinstance(data.dtype.metadata["vlen"], str):
```
stephprince:
```suggestion
                    if data.dtype.metadata["vlen"] is str:
```

I think the original code here was checking if it is the str class itself

stephprince (changes requested):
small suggestion otherwise looks good to me!

stephprince (approved):

PR #1140, opened 2024-06-29
Move scipy to optional requirements

rly:
## Motivation

`scipy` is used only by the `CSRMatrix` data type, which is currently rarely used. To reduce the chance of dependency conflicts and reduce the install time, I propose we make `scipy` an optional requirement.

This is a breaking change since users using the `CSRMatrix` type will need to install `scipy` separately.

Merging into new `hdmf_4.0` branch.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 4.0.0 (Upcoming)
+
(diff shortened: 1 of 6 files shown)


mavaylon1 (approved):

mavaylon1 (changes requested):
Code Coverage

PR #1141, opened 2024-07-01
Make Zarr optional for testing

mavaylon1:
## Motivation

After discussions with the team, we decided to have the zarr be an optional dependency. 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -7,7 +7,7 @@ from typing import Tuple
 from itertools import product, chain
 
 try:
-    from zarr import ZarrArray
+    from zarr import Array as ZarrArray
     ZARR_INSTALLED = True
 except ImportError:
     ZARR_INSTALLED = False
@@ -45,7 +45,6 @@ def append_data(data, arg):
     else:
         msg = "Data cannot append to object of type '%s'" % type(data)
         raise ValueError(msg)
-
 
 
 def extend_data(data, arg):
--- a/src/hdmf/data_utils.py
+++ b/src/hdmf/data_utils.py
@@ -35,13 +35,14 @@ def append_data(data, arg):
         data.resize(shape)
         data[-1] = arg
         return data
+    elif ZARR_INSTALLED:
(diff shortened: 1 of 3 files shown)


mavaylon1:
@rly We don't currently cover the else statement in append data. The gap in this PR is the same else statment, but under the zarr conditional. Are you okay with this gap in coverage for now?

mavaylon1:
@rly made a simple example. It has full coverage and covers the missing spot I mentioned from prior to this PR. It's not a actual iterable class since it does not have iter or next, but I think it is fine.

src/hdmf/data_utils.py
```diff
@@ -33,6 +35,13 @@ def append_data(data, arg):
         data.resize(shape)
         data[-1] = arg
         return data
+    elif ZARR_INSTALLED:
+        if isinstance(data, ZarrArray):
+            data.append([arg], axis=0)
+            return data
+        else:
+            msg = "Data cannot append to object of type '%s'" % type(data)
+            raise ValueError(msg)
```
stephprince:
```suggestion
    elif ZARR_INSTALLED and isinstance(data, ZarrArray):
        data.append([arg], axis=0)
        return data
```

stephprince: I think this can be consolidated since it is the same error handling?

tests/unit/utils_test/test_data_utils.py
```diff
@@ -3,9 +3,36 @@
...
+class TestZarrAppendData(TestCase):
+
+    def setUp(self):
+        if not ZARR_INSTALLED:
+            self.skipTest("optional Zarr package is not installed")
+
+    def test_append_exception(self):
+        data = MyIterable([1, 2, 3, 4, 5])
+        with self.assertRaises(ValueError):
+            append_data(data, 4)
```
stephprince:
```suggestion

```

If you consolidate the error handling I think you would not need the same append exception test twice

mavaylon1:
@stephprince done

stephprince (approved):

PR #1149, opened 2024-07-10
Fix #1148 : Add passthrough on non-DCI H5DataIO to support its use in pynwb TimeSeries.

### LINKED ISSUE #1148: [Bug]: Non-DCI H5DataIO does not passthrough `maxshape` property which crashes pynwb Timeseries
### What happened?

The pynwb documentation provides an example of passing a `H5DataIO` object as the `data` argument to the `TimeSeries` constructor. [docs here](https://pynwb.readthedocs.io/en/latest/tutorials/advanced_io/plot_iterative_write.html#alternative-approach-user-defined-dataset-write). However, this only works in the case where `timestamps=None` because, if timestamps is not None, then the `TimeSeries` initializer will check for the presence of `maxshape` and this fails because `H5DataIO` does not have `.maxshape`, it only has `.io_settings["maxshape"]`.

I raised this bug in pynwb here: https://github.com/NeurodataWithoutBorders/pynwb/issues/1929

Are there any unintended side-effects of adding a `maxshape` getter to `H5DataIO`?
Should this be fixed in pynwb? Maybe its `TimeSeries` `_check_time_series_dimension` can have a special case if the data object is H5DataIO.

For now I will monkey-patch H5DataIO to add a `maxshape` property.

### Steps to Reproduce

```python
import datetime
import numpy as np
from hdmf.backends.hdf5.h5_utils import H5DataIO
from pynwb import NWBHDF5IO, NWBFile, TimeSeries


def test_nwb_stream_timestamps():
    srate = 10.0
    data_shape = 20, 16, 2
    test_data = np.arange(np.prod(data_shape)).reshape(data_shape)

    dataio = H5DataIO(
        shape=(0,) + data_shape[1:],
        dtype=test_data.dtype,
        maxshape=(None,) + data_shape[1:],
        fillvalue=np.nan,
    )
    # Next line fails
    data_series = TimeSeries(
        name="some name",
        data=dataio,
        timestamps=np.arange(test_data.shape[0]),
        # rate=srate,  # Comment previous line and uncomment this to bypass error
        unit="n/a",
    )

    nwbfile = NWBFile(
        session_description="demonstrate streaming timestamps",
        identifier="abcd1234",
        session_start_time=datetime.datetime.now(datetime.timezone.utc),
    )
    nwbfile.add_acquisition(data_series)

    # Write the data to file
    file_path = Path(tempfile.gettempdir()) / "test_nwb_stream_timestamps.nwb"
    io = NWBHDF5IO(file_path, "w")
    io.write(nwbfile)

    dataio.dataset.resize(len(dataio.dataset) + len(data), axis=0)
    dataio.dataset[-len(data) :] = data
    timestampio.dataset.resize(len(timestampio.dataset) + len(timestamps), axis=0)
    timestampio.dataset[-len(timestamps) :] = timestamps

    io.close()
```

### Traceback

```pytb
>       data_series = TimeSeries(
            name="TODO: name from params",
            data=dataio,
            timestamps=timestampio,
            # rate=srate,
            unit="n/a",
        )

test_nwb.py:57: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.venv/lib/python3.9/site-packages/hdmf/utils.py:667: in func_call
    pargs = _check_args(args, kwargs)
../../../.venv/lib/python3.9/site-packages/hdmf/utils.py:639: in _check_args
    parsed = __parse_args(
../../../.venv/lib/python3.9/site-packages/hdmf/utils.py:351: in __parse_args
    valshape = get_data_shape(argval)
../../../.venv/lib/python3.9/site-packages/hdmf/utils.py:916: in get_data_shape
    if hasattr(data, 'maxshape'):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <hdmf.backends.hdf5.h5_utils.H5DataIO object at 0x14755d190>
attr = 'maxshape'

    def __getattr__(self, attr):
        """Delegate attribute lookup to data object"""
        if attr == '__array_struct__' and not self.valid:
            # np.array() checks __array__ or __array_struct__ attribute dep. on numpy version
            raise InvalidDataIOError("Cannot convert data to array. Data is not valid.")
        if not self.valid:
>           raise InvalidDataIOError("Cannot get attribute '%s' of data. Data is not valid." % attr)
E           hdmf.data_utils.InvalidDataIOError: Cannot get attribute 'maxshape' of data. Data is not valid.

../../../.venv/lib/python3.9/site-packages/hdmf/data_utils.py:1104: InvalidDataIOError
```

### Operating System

macOS

### Python Executable

Python

### Python Version

3.9

### Package Versions

h5py==3.11.0
hdmf==3.14.1
hdmf_zarr==0.8.0
pynwb==2.8.0

cboulay:
Fix #1148 

## Motivation

As explained in #1148 and [pynwb #1929](https://github.com/NeurodataWithoutBorders/pynwb/issues/1929), when a `pynwb.TimeSeries` object is created where (1) its `data` is a `H5DataIO` object that DOES NOT wrap a DataChunkIterator, AND (2) its `timestamps` are provided explicitly (vs `rate` and `starting_time`), then the `TimeSeries` object crashes because it cannot access the `maxshape` attribute on its `data` value.

With this change, the `H5DataIO` has a new `maxshape` property which passes through the correct maxshape if known (either set during `__init__` or inferred from the dataset) else `None`, and `pynwb.TimeSeries` no longer crashes.

## How to test the behavior?

I added a unit test in unit/test_io_hdf5_h5tools.py

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -637,3 +637,13 @@ class H5DataIO(DataIO):
         if isinstance(self.data, Dataset) and not self.data.id.valid:
             return False
         return super().valid
+
+    @property
+    def maxshape(self):
+        if 'maxshape' in self.io_settings:
+            return self.io_settings['maxshape']
+        elif hasattr(self.data, 'maxshape'):
+            return self.data.maxshape
+        elif 'shape' in self.io_settings:
+            return self.io_settings['shape']
+        return None
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -21,7 +21,7 @@ from ...data_utils import DataIO, AbstractDataChunkIterator, append_data
 from ...query import HDMFDataset, ReferenceResolver, ContainerResolver, BuilderResolver
 from ...region import RegionSlicer
 from ...spec import SpecWriter, SpecReader
-from ...utils import docval, getargs, popargs, get_docval
+from ...utils import docval, getargs, popargs, get_docval, get_data_shape
 
 
 class HDF5IODataChunkIteratorQueue(deque):
@@ -681,6 +681,5 @@ class H5DataIO(DataIO):
             return self.data.maxshape
         elif hasattr(self, "shape"):
             return self.shape
-        elif 'shape' in self.io_settings:
-            return self.io_settings['shape']
-        return None
+        else:
+            return get_data_shape(self.data)
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -644,6 +644,8 @@ class H5DataIO(DataIO):
             return self.io_settings['maxshape']
         elif hasattr(self.data, 'maxshape'):
             return self.data.maxshape
+        elif hasattr(self, "shape"):
+            return self.shape
         elif 'shape' in self.io_settings:
             return self.io_settings['shape']
         return None
--- a/tests/unit/test_io_hdf5_h5tools.py
+++ b/tests/unit/test_io_hdf5_h5tools.py
@@ -569,6 +569,12 @@ class H5IOTest(TestCase):
         dset = self.f["test_dataset"]
         self.assertEqual(dset.chunks, chunk_shape)
 
+    def test_pass_through_of_maxshape_on_h5dataset(self):
+        k = 10
+        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', np.arange(k), attributes={}))
+        dset = H5DataIO(self.f['test_dataset'])
+        self.assertEqual(dset.maxshape, (k,))
+
     #############################################
     #  H5DataIO general
     #############################################
--- a/tests/unit/test_io_hdf5_h5tools.py
+++ b/tests/unit/test_io_hdf5_h5tools.py
@@ -604,15 +604,15 @@ class H5IOTest(TestCase):
         dset = self.f["test_dataset"]
         self.assertEqual(dset.chunks, chunk_shape)
 
+    #############################################
+    #  H5DataIO general
+    #############################################
     def test_pass_through_of_maxshape_on_h5dataset(self):
         k = 10
         self.io.write_dataset(self.f, DatasetBuilder('test_dataset', np.arange(k), attributes={}))
         dset = H5DataIO(self.f['test_dataset'])
         self.assertEqual(dset.maxshape, (k,))
 
-    #############################################
-    #  H5DataIO general
-    #############################################
     def test_warning_on_non_gzip_compression(self):
         # Make sure no warning is issued when using gzip
         with warnings.catch_warnings(record=True) as w:
@@ -3768,6 +3768,14 @@ class H5DataIOTests(TestCase):
         dataio = H5DataIO(shape=(10, 10), dtype=int)
         with self.assertRaisesRegex(ValueError, "Setting data when dtype and shape are not None is not supported"):
             dataio.data = list()
+
+    def test_dataio_maxshape(self):
+        dataio = H5DataIO(data=np.arange(10), maxshape=(None,))
+        self.assertEqual(dataio.maxshape, (None,))
+
+    def test_dataio_maxshape_from_data(self):
+        dataio = H5DataIO(data=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+        self.assertEqual(dataio.maxshape, (10,))
 
 
 def test_hdf5io_can_read():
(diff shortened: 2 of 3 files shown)


src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -658,3 +658,15 @@ def valid(self):
...
+    def maxshape(self):
+        if 'maxshape' in self.io_settings:
+            return self.io_settings['maxshape']
+        elif hasattr(self.data, 'maxshape'):
+            return self.data.maxshape
+        elif hasattr(self, "shape"):
+            return self.shape
+        elif 'shape' in self.io_settings:
+            return self.io_settings['shape']
+        return None
```
stephprince:
```suggestion
        else:
            return get_data_shape(self.data)
```

Did you have an example in mind for the `elif 'shape' in self.io_settings` condition? I think `shape` in `self.io_settings` will only be true if the `hasattr(self, "shape")` condition above is also true, but I might be missing a case.

I think there's also one more case where `maxshape` is not specified and the wrapped data object does not have a shape attribute (e.g. if `self.data` is a list/tuple), we could use `get_data_shape` in this case and it should return `None` if the shape cannot be inferred.

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -569,6 +569,12 @@ def test_pass_through_of_chunk_shape_generic_data_chunk_iterator(self):
...
 
+    def test_pass_through_of_maxshape_on_h5dataset(self):
+        k = 10
+        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', np.arange(k), attributes={}))
+        dset = H5DataIO(self.f['test_dataset'])
+        self.assertEqual(dset.maxshape, (k,))
+
     #############################################
     #  H5DataIO general
     #############################################
```
stephprince: Small nitpick comment - could you move this test to the "H5DataIO general" section below?

```suggestion
    #############################################
    #  H5DataIO general
    #############################################
    
    def test_pass_through_of_maxshape_on_h5dataset(self):
        k = 10
        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', np.arange(k), attributes={}))
        dset = H5DataIO(self.f['test_dataset'])
        self.assertEqual(dset.maxshape, (k,))
```

stephprince: Could you also add a test that fails without the maxshape property. Something like these could go in the [H5DataIOTests section](https://github.com/hdmf-dev/hdmf/blob/49a60df21016fb4da83d64e84e7bc99dac03e94e/tests/unit/test_io_hdf5_h5tools.py#L3664C1-L3667C33):

```python
    def test_dataio_maxshape(self):
        dataio = H5DataIO(data=np.arange(10), maxshape=(None,))
        self.assertEqual(dataio.maxshape, (None,))

    def test_dataio_maxshape_from_data(self):
        dataio = H5DataIO(data=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
        self.assertEqual(dataio.maxshape, (10,))
```

stephprince (commented):
Hi @cboulay, sorry for the delayed review. I added a couple of small suggestions but otherwise looks great. Thanks again for the PR!

cboulay:
Sorry, this fell off the radar. I'll look at it this evening.

src/hdmf/backends/hdf5/h5_utils.py
cboulay: Is there already an implementation of `get_data_shape` somewhere that I should use? The one in `from hdmf.utils import get_data_shape`?

stephprince: Yes that's the one I was thinking of!

cboulay: Re `elif 'shape' in self.io_settings`... instantiating `H5DataIO(data=None, shape=(1,2))` will cause the base `DataIO` class to set `self.__data = None` and `self.__shape = shape`. I see now that if `self.__data is None`, then the `H5DataIO` constructor will assign `self.__iosettings['shape'] = self.shape`.

It's a little bit weird to have this redundant information stored in 2 locations in the same class. I guess I was hedging against the possibility of them getting out of sync somehow. I don't mind removing this condition.

cboulay:
@stephprince , thank you for the suggestions. I incporporated all your suggested changes.

stephprince (approved):
Thanks for the updates! Looks good to me

PR #1154, opened 2024-07-26
Add doc to H5DataIO.dataset setter

rly:
## Motivation

While reading through the code for `H5DataIO`, I was confused about the purpose of the `dataset` property and how it differs from `data` since it is not used locally: https://github.com/hdmf-dev/hdmf/blob/dfb1df79bce3e34c52af7996429c3f561d96390a/src/hdmf/backends/hdf5/h5_utils.py#L545-L555

It turns out `dataset` was added for a very specific use case where a user wants to create an empty hdf5 dataset, write it to disk, and have access to the handle to that dataset without having to close and reopen the file -- https://github.com/hdmf-dev/hdmf/pull/747

To prevent myself and others from retracing those steps to understand this code, I added a docstring.

## Checklist

### DIFF
--- a/src/hdmf/backends/hdf5/h5_utils.py
+++ b/src/hdmf/backends/hdf5/h5_utils.py
@@ -551,8 +551,9 @@ class H5DataIO(DataIO):
 
     @dataset.setter
     def dataset(self, val):
-        """Cache an h5py.Dataset, which can be used to cache a written, empty dataset and fill it in later.
+        """Cache the h5py.Dataset written with the stored IO settings.
 
+        This attribute can be used to cache a written, empty dataset and fill it in later.
         This allows users to access the handle to the dataset *without* having to close and repoen a file.
 
         For example::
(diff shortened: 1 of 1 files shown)


src/hdmf/backends/hdf5/h5_utils.py
```diff
@@ -546,10 +546,30 @@ def __init__(self, **kwargs):
...
     def dataset(self):
+        """Get the cached h5py.Dataset."""
         return self.__dataset
 
     @dataset.setter
     def dataset(self, val):
+        """Cache the h5py.Dataset written with the stored IO settings.
+
+        This attribute can be used to cache a written, empty dataset and fill it in later.
+        This allows users to access the handle to the dataset *without* having to close and repoen a file.
```
oruebel:
```suggestion
        This attribute can be used to cache a written, empty dataset and fill it in later.
        This allows users to access the handle to the dataset *without* having to close 
        and reopen a file.
```

oruebel (approved):

PR #1158, opened 2024-07-31
VectorData Refactor Expandable

### LINKED ISSUE #1153: [Feature]: Expandable datasets (HDF5)

We previously implemented VectorData expanable by default, but had to pull it from production. https://github.com/hdmf-dev/hdmf/pull/1093/files#diff-508770382f78baf6681e42837b28809b522de14e7fa401c32daffa28538396c4

We are doing round 2:
1. Implement the same setup as the ticket above; however, pull shape information from the schema. There are examples where you have multiple shape options in the schema; we need to ensure that the shape option used matches the shape of the data. If there is no shape, determine via dim of the data.
2. We want to support dataset of references also being expandable. Look to append ticket: https://github.com/hdmf-dev/hdmf/pull/1135/files#diff-2f0fd76dfef06a7d2164a052a72a224ea80cafbfde6dd753aca67d2b1eb02145

above

Yes.

mavaylon1:
## Motivation

Fix #1153 

## How to test the behavior?
```
```

## Checklist


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1473,7 +1489,13 @@ def __list_fill__(cls, parent, name, data, options=None):
             data_shape = (len(data),)
         else:
             data_shape = get_data_shape(data)
-
+        if expandable:
+            # Don't override existing settings
+            if 'maxshape' not in io_settings:
+                if matched_spec_shape is not None:
+                    io_settings['maxshape'] = tuple([None]*len(matched_spec_shape))
```
mavaylon1: This is not correct. You need to keep the dimensions that are explicit.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1473,7 +1489,13 @@ def __list_fill__(cls, parent, name, data, options=None):
             data_shape = (len(data),)
         else:
             data_shape = get_data_shape(data)
-
+        if expandable:
+            # Don't override existing settings
+            if 'maxshape' not in io_settings:
+                if matched_spec_shape is not None:
+                    io_settings['maxshape'] = matched_spec_shape
```
mavaylon1, line 1492: This needs an additional test where the spec has a non-none shape. Even if this line is covered, that test case is not

mavaylon1:
Note: Dataset of references will be handled with HDMF proxy append (PR tbd)

mavaylon1:
@rly I plan on wrapping this up today; however, I wanted to point out #1163. I saw that my implementation and also dimension labels omit the case of untyped, named datasets. They are both related enough in terms of the code required that we can tackle them in a single Issue/PR. 

That being said, to prevent scope creep and because I believe this PR is solid enough to stand as a merge, let's aim to get this merged this week.

mavaylon1:
Review Notes: 
We already synced about the tests within PyNWB that failing in regards to DecompositionSeries. There is a ticket: https://github.com/NeurodataWithoutBorders/pynwb/issues/1947

**What is the logic of this PR:**
The goal is to be phase 1 of having datasets be expandable by default for the HDF5 backend. This can be toggled off within the user facing `write` method. 

1. Within h5tools.py

- Feed the new `expandable` parameter from write into write_builder, write_group, write_dataset, and __list_fill__. The reason we adjust list_fill is because that is what we use to write array data.
- Within list_fill the update is having a boolean conditional for `expandable`. I did not want to override user defined maxshape from H5DataIO. If maxshape is not provided, then maxshape is set to matched_spec_shape. The logic of matched_spec_shape is handled in the objectmapper, but stored in the builder.

2. Within Builders.py
The idea relies on getting the shape from the schema which is in the spec. The builder needs have the shape from the spec for us to be able to use it for maxshape. Why? Let's say in the schema is its [None, 2]. That means in the second dimension I want the data to be 2 no matter what. The first dimension can be whatever size. If we do not use the shape from the spec, it will be [None, None] which will fail validation in the validator and just be not what the schema/or user defines/expects.

The update is to have the shape in the DatasetBuilder.

3. Within objectmapper.py
Here lies I would say is the most important part of the PR. I will go into that, but I will also make a few comments about what is NOT in here.

Recall what the objectmapper is for: map a container to a builder (visa versa). Within the `build` method, we need to make sure that the DatasetBuilders we create have the shape from the spec. This is done by `__get_spec_info`, which is a modified method for getting the dimension labels. I then added spec_shapes into the DatasetBuilder (see note).

The `__get_spec_info` is modifies `__get_matched_dimension` in a pretty small way. The logic of the function uses the shape anyways. I know just return the corresponding shape with the dimension labels. The other note is that it now supports the case where the shape is defined but the dimension labels are not. 

Note: There is a spot where I do not update the DatasetBuilder. That is described in #1163. It will be phase 2.
4. Updated and New tests
I created a new test class Qux to easily have a tool with a variable shapes.

mavaylon1:
Todo items:
- [x]  Code Coverage
- [x] Your assumption on the spec in the object mapper is incorrect.
- [x] Create a staging branch and have this PR merge into that.

PR #1162, opened 2024-08-07
Allow Codec for compression for use with LINDI

magland:
## Motivation

While hdf5 does not allow numcodecs.Codec for compression, [lindi](https://github.com/neurodatawithoutborders/lindi) does. Right now hdmf does type checking for the compression parameter and raises an exception if the type is not in (str, bool, int). This PR simply allows the Codec type to be passed through so that it can be used with lindi.

## How to test the behavior?

Here's an example that applies a custom compressor to filtered ephys data

https://github.com/NeurodataWithoutBorders/lindi/blob/lindi-tar/examples/DANDI/preprocess_ephys.py

You need to install the lindi-tar branch of lindi and `pip install qfc`. See https://github.com/magland/qfc

@oruebel @rly 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 - Added support to append to a dataset of references for HDMF-Zarr. @mavaylon1 [#1157](https://github.com/hdmf-dev/hdmf/pull/1157)
+- Allow Codec for compression for use with LINDI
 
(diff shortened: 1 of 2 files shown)


magland:
Oops I accidentally deleted my fork and that closed this PR. I created a new fork and I am going to open another PR. Sorry about that.

PR #1167, opened 2024-08-10
Add ObjectMapper.NO_OVERRIDE sentinel and deprecate override functions returning None

### LINKED ISSUE #178: Allow overwrite of constructor arguments to return None
When customizing in the ObjectMapper the mapping for a constructor argument, the mapping is ignored if the function returns None. E.g., 

```
@register_map(Sweeps)
class SweepsMap(DynamicTableMap):
    @ObjectMapper.constructor_arg('intracellular_recordings')
    def intracellular_recordings_arg(self, builder, manager):
          return None
```

will always be ignored. This is due to the following:

https://github.com/hdmf-dev/hdmf/blob/44c02ea7f3666d68aa7402407c91e9503f0fd99b/src/hdmf/build/map.py#L1204-L1210

Here the check ``if override is not None:`` is used to check if a constructor argument is overwritten or not. Instead the check should look for whether a custom mapping is defined, rather than using the None return value as an indicator that there is no overwrite. 

**Proposed solution** Add function **has_overwrite_carg** to check if a constuctor argument is overwritten and use it instead. 

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/hdmf-dev/hdmf/issues) ?

rly:
## Motivation

Fix #178

`ObjectMapper` `constructor_arg` and `object_attr` override functions cannot distinguish "no override, use the value from the file/container" from "override the value to `None`". Returning `None` is the signal to fall through to the default resolution (see the check `if override is not None:` in `construct`, and the equivalent in `get_attr_value`), so an override can never legitimately set a constructor argument or attribute to `None`.

## Approach

Add `ObjectMapper.NO_OVERRIDE`, a sentinel that an override function returns to fall through to the value built from the file or read from the container. This frees `None` to become a real override value.

Making `None` mean the value `None` is a behavior change for any existing override (in HDMF, PyNWB, or extensions) that returns `None` to signal "no override", so the change is deferred rather than applied immediately:

- Now: returning `None` from an override still falls through to the built/read value, but emits a `DeprecationWarning` when doing so actually changes the result. Override functions should return `NO_OVERRIDE` to fall through silently.
- HDMF 8.0: a `None` return will set the constructor argument or attribute to `None`.

`MeaningsTableMap.target_carg`, which relied on the old fall-through, is updated to return `NO_OVERRIDE`.

## Checklist


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Enhancements
 - Added support to append to a dataset of references for HDMF-Zarr. @mavaylon1 [#1157](https://github.com/hdmf-dev/hdmf/pull/1157)
+- Allow override of constructor args and object attrs to return None. @rly [#1167](https://github.com/hdmf-dev/hdmf/pull/1167)
```
```suggestion
- Allow override of constructor args and object attrs in `hdmf.build.ObjectMapper` to return None. @rly [#1167](https://github.com/hdmf-dev/hdmf/pull/1167)
```

rly:
This causes a few errors in PyNWB that I will investigate.

mavaylon1:
> This causes a few errors in PyNWB that I will investigate.

Is this PR okay to review or should I wait for the PyNWB errors.

rly:
> > This causes a few errors in PyNWB that I will investigate.
> 
> Is this PR okay to review or should I wait for the PyNWB errors.

Please wait to merge. I haven't had time to look through these yet. If it does not make it in this upcoming release, that is OK.

rly:
## Downstream impact: NWB extension mapper scan

To gauge the blast radius of deprecating "override returns `None` → fall through", I scanned the custom `ObjectMapper`s across the extension ecosystem for the affected pattern (an override that returns `None` while the mapped value is still present, which would warn now and be set to `None` in HDMF 8.0).

**Scope:** 75 repositories — the full [nwb-extensions catalog](https://nwb-extensions.github.io) (all 30 `-record` repos → their `src` repos, including 2 on GitLab and the ones CI marks inactive) plus every `ndx-*` repo in the `catalystneuro` org, deduplicated. Each was shallow-cloned and grepped for `@*.constructor_arg` / `@*.object_attr` / `register_map`, and every hit was read and classified.

**Finding: no extension needs a migration.** Only 6 repos define any custom mapper. Three have override functions that can return `None`, but all in the safe case (the value is absent, so `const_args`/`getattr` is also `None` → no warning, no change in 8.0):

- **`ndx-pose`** (`nodes`, `edges`): returns `None` only for ≥0.2.0 files, where those datasets moved to `Skeleton` and aren't in the builder. The spec no longer declares them, so `const_args` never holds them, and both args default to `None`.
- **`ndx-icephys-meta`** (`electrodes`/`stimuli`/`responses`): returns `None` only when the category table is absent (same pattern as pynwb core icephys). Also deprecated (folded into core).
- **`ndx-acquisition-module`** (`name`): returns `builder.name`, never `None`.

The other three (`ndx-events`, `ndx-ontology-table`, `ndx-point-cloud-table`) only remap specs or are empty subclasses — no override functions. The remaining ~67 repos have no custom Python mappers.

**Takeaway:** the affected pattern — returning `None` to suppress an override *while the value still exists in the file/container* — is specific to PyNWB core (`columns_carg`, `experimenter_obj_attr`, `publication_obj_attr`, `device.model_carg`, and several legacy mappers), handled in NeurodataWithoutBorders/pynwb#2224. Extension authors that return `None` do so only when the target is genuinely absent, which is safe.

**Caveats:** covers the catalog + `catalystneuro` only; scanned each repo's default branch; matched the decorator patterns (a mapper overriding `get_attr_value`/`construct` directly rather than via the decorators wouldn't be flagged, but that isn't affected by `NO_OVERRIDE`).

oruebel (approved):

src/hdmf/build/objectmapper.py
```diff
@@ -669,27 +669,38 @@ def get_attr_value(self, **kwargs):
...
-                               % (container.__class__.__name__, container.name, attr_name))
-                        raise ContainerConfigurationError(msg) from e
-            # else: attr_val is an attribute on the Container and its value is None
+        override = self.__get_override_attr(attr_name, container, manager)
+        if override is not self.NO_OVERRIDE and override is not None:
+            return override
+        # No override function is registered (NO_OVERRIDE), or an override function returned None. In
+        # both cases the value is resolved from the container attribute.
+        # TODO(HDMF 8.0): return None directly when an override function returns None, instead of
+        # falling through to the container attribute.
```
oruebel, line 678: Is there a corresponding issue for this TODO item that is assigned to the 8.0 milestone?

rly: Here we go: https://github.com/hdmf-dev/hdmf/issues/1541

PR #1168, opened 2024-08-10
Check DynamicTableRegion data is in bounds

### LINKED ISSUE #210: Do not allow write with a `DynamicTableRegion` index that is out of bounds
Do not allow writing of a file with a `DynamicTableRegion` index that is out of bounds. The file is unreadable. Attempting to read it results in a cryptic error message (#209).

rly:
## Motivation

Fix #210 

## Checklist


src/hdmf/common/table.py
```diff
@@ -1360,42 +1360,64 @@ class DynamicTableRegion(VectorData):
...
             msg = "can't set attribute 'table' -- already set"
             raise AttributeError(msg)
-        dat = self.data
-        if isinstance(dat, DataIO):
-            dat = dat.data
-        self.fields['table'] = val
+
+        self.fields['table'] = table
+        for val in self.data:
+            self._validate_new_data_element(val)
```
oruebel: You might want to try this as a list comprehension ``[self._validate_new_data_element(val) for val in self.data]`` just because they tend to be faster in Python than regular for loops

src/hdmf/common/table.py
```diff
@@ -1360,42 +1360,64 @@ class DynamicTableRegion(VectorData):
...
+    def _validate_index_in_range(self, data, table):
+        """If the length of data is small, and if data contains an index that is out of bounds, then raise an error.
+        If the object is being constructed from a file, raise a warning instead to ensure invalid data can still be
+        read.
+        """
+        if table and len(data) <= self.MAX_ROWS_TO_VALIDATE_INIT:
+            for val in data:
+                if val >= len(table) or val < 0:
+                    error_msg = f"DynamicTableRegion index {val} is out of bounds for {type(table)} '{table.name}'."
+                    self._error_on_new_warn_on_construct(error_msg, error_cls=IndexError)
```
oruebel: Instead of iterating through all the values, I think checking for `np.max(data) >= len(table)` would be more efficient.

rly: I was skeptical because `np.max` doesn't short circuit, but it seems like `np.max` is highly optimized under the hood. `np.where` is also highly optimized and returns the violating values which is useful, so I am going to use that.

src/hdmf/common/table.py
rly: It turns out we can reuse the `self._validate_index_in_range` above.

tests/unit/build_tests/test_classgenerator.py
```diff
@@ -659,9 +658,6 @@ class TestGetClassSeparateNamespace(TestCase):
 
     def setUp(self):
         self.test_dir = tempfile.mkdtemp()
-        if os.path.exists(self.test_dir):  # start clean
```
rly, line 680: This code raised hidden errors during test setup. `tempfile.mkdtemp()` creates a clean temp directory so the below code is not needed

tests/unit/test_io_hdf5_streaming.py
```diff
@@ -80,9 +80,9 @@ def setUp(self):
         self.manager = BuildManager(type_map)
 
     def tearDown(self):
-        if os.path.exists(self.ns_filename):
+        if hasattr(self, 'ns_filename') and os.path.exists(self.ns_filename):
```
rly, line 83: This code raised hidden errors during test setup when this test was skipped because ROS3 is not installed or there is no internet, because the instance vars are not defined before skipping and tearing down.

rly:
#1293 should be merged before this

src/hdmf/common/table.py
```diff
@@ -1499,42 +1493,73 @@ class DynamicTableRegion(VectorData):
         'table',
     )
 
+    MAX_ROWS_TO_VALIDATE_INIT = int(1e3)
```
oruebel: Why is ` MAX_ROWS_TO_VALIDATE_INIT ` a class variable instead being an optional argument on `__init__` and `def table`?

src/hdmf/common/table.py
```diff
@@ -1499,42 +1493,73 @@ class DynamicTableRegion(VectorData):
...
-        self.table = t
+        if table is not None:  # set the table attribute using fields to avoid another validation in the setter
+            self.fields['table'] = table
+
+    def _validate_index_in_range(self, data, table):
+        """If the length of data is small, and if data contains an index that is out of bounds, then raise an error.
+        If the object is being constructed from a file, raise a warning instead to ensure invalid data can still be
+        read.
+        """
+        if table and len(data) <= self.MAX_ROWS_TO_VALIDATE_INIT:
```
oruebel: `MAX_ROWS_TO_VALIDATE_INIT` implied to me that only that many rows would be validated, but the logic here implies that validation is fully being disabled once we have  more than `MAX_ROWS_TO_VALIDATE_INIT` rows. I.e., I think we should either:
1. Use a boolean flag on `__init__` and `table` to say `disable_index_validation`, or
2. Update the logic here to only validate the first N rows as specified by `MAX_ROWS_TO_VALIDATE_INIT`

I think option 1 is probably cleaner and more useful. I don't think it is very useful to validate the first 1000 rows and then miss the error in row 1001. I think for a user it is probably more intuitive to either validate or not.

rly: I agree and made the changes for option 1.

src/hdmf/common/table.py
```diff
@@ -1505,36 +1499,96 @@ class DynamicTableRegion(VectorData):
...
             {'name': 'table', 'type': DynamicTable,
              'doc': 'the DynamicTable this region applies to', 'default': None},
+            {'name': 'validate_data', 'type': bool,
+             'doc': 'whether to validate the data is in bounds of the linked table', 'default': True},
             allow_positional=AllowPositional.WARNING)
     def __init__(self, **kwargs):
-        t = popargs('table', kwargs)
+        table, validate_data = popargs('table', 'validate_data', kwargs)
+        data = getargs('data', kwargs)
+        self._validate_data = validate_data
```
oruebel, line 1508: Not critical. With this setup it may be nice to add a setter/getter for the validate_date field. But that can be a separate issue and is probably not necessary in most cases. Maybe something that can wait until a user actually has a real need.

oruebel (approved):
Looks good to me.

PR #1173, opened 2024-08-15
Add support to write multidimensional string arrays

### LINKED ISSUE #1096: [Bug]: Support writing multidimensional string datasets and attributes
### What happened?

In trying to resolve https://github.com/NeurodataWithoutBorders/pynwb/pull/1886/ in HDMF, I discovered that on build, a multidimensional list or array that maps to a multidimensional dataset/attribute spec is converted to a 1-D list or array.

`[['a', 'b'], ['c', 'd']]` -> `["['a', 'b']", "['c', 'd']"]`
`np.array([['a', 'b'], ['c', 'd']])` -> `array(["['a' 'b']", "['c' 'd']"], dtype='<U9')`

### Steps to Reproduce

```python
def test_build_2d_lol(self):
        bar_spec = GroupSpec(
            doc='A test group specification with a data type',
            data_type_def='Bar',
            datasets=[
                DatasetSpec(
                    doc='an example dataset',
                    dtype='text',
                    name='data',
                    shape=(None, None),
                    attributes=[AttributeSpec(name='attr2', doc='an example integer attribute', dtype='int')],
                )
            ],
            attributes=[AttributeSpec(name='attr1', doc='an example string attribute', dtype='text')],
        )
        type_map = self.customSetUp(bar_spec)
        type_map.register_map(Bar, BarMapper)
        bar_inst = Bar('my_bar', [['a', 'b'], ['c', 'd']], 'value1', 10)
        builder = type_map.build(bar_inst)
        self.assertEqual(builder.get('data').data, [['a', 'b'], ['c', 'd']])

    def test_build_2d_ndarray(self):
        bar_spec = GroupSpec(
            doc='A test group specification with a data type',
            data_type_def='Bar',
            datasets=[
                DatasetSpec(
                    doc='an example dataset',
                    dtype='text',
                    name='data',
                    shape=(None, None),
                    attributes=[AttributeSpec(name='attr2', doc='an example integer attribute', dtype='int')],
                )
            ],
            attributes=[AttributeSpec(name='attr1', doc='an example string attribute', dtype='text')],
        )
        type_map = self.customSetUp(bar_spec)
        type_map.register_map(Bar, BarMapper)
        bar_inst = Bar('my_bar', np.array([['a', 'b'], ['c', 'd']]), 'value1', 10)
        builder = type_map.build(bar_inst)
        np.testing.assert_array_equal(builder.get('data').data, np.array([['a', 'b'], ['c', 'd']]))
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.12

### Package Versions

_No response_

### LINKED ISSUE #1137: Cannot write multi-dimensional string arrays
This only works for 1-D string arrays. I think the if statement in this code is intended to check for structured arrays / compound dtypes. Else, execution should go to `get_data_shape` which gets the shape correctly. 

https://github.com/hdmf-dev/hdmf/blob/7426275cacc769a10ffca89836765df1355ba9db/src/hdmf/backends/hdf5/h5tools.py#L1472-L1473

stephprince:
## Motivation

Fixes https://github.com/hdmf-dev/hdmf/issues/1137 and fixes https://github.com/hdmf-dev/hdmf/issues/1096. 

On build, multidimensional string lists/arrays were being converted into 1-D lists/arrays:

`[['aa', 'bb'], ['cc', 'dd']] -> ["['aa', 'bb']", "['cc', 'dd']"]`.

On write, the shape of a multidimensional string dataset was being defined incorrectly and so the dataset could not be written. The data shape definition was being caught by this conditional that I believe is intended to check for structured arrays / compound data types:

https://github.com/hdmf-dev/hdmf/blob/7426275cacc769a10ffca89836765df1355ba9db/src/hdmf/backends/hdf5/h5tools.py#L1472-L1473

 I updated the conditional and `get_data_shape` is now used to set the shape of the dataset for multidimensional string lists/arrays.

## How to test the behavior?

This came up in https://github.com/NeurodataWithoutBorders/pynwb/pull/1924. 

## Checklist

### DIFF
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -617,7 +617,9 @@ class ObjectMapper(metaclass=ExtenderMeta):
                     def string_type(x):
                         return x.isoformat()  # method works for both date and datetime
                 if string_type is not None:
-                    if spec.shape is not None or spec.dims is not None:
+                    if spec.shape is not None and len(spec.shape) > 1:
+                        ret = [[string_type(item) for item in sublist] for sublist in value]
+                    elif spec.shape is not None or spec.dims is not None:
                         ret = list(map(string_type, value))
                     else:
                         ret = string_type(value)
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -598,11 +598,17 @@ class ObjectMapper(metaclass=ExtenderMeta):
 
     def __convert_string(self, value, spec):
         """Convert string types to the specified dtype."""
+        def __apply_string_type(value, string_type):
+            if isinstance(value, (list, tuple, np.ndarray, DataIO)):
+                return [__apply_string_type(item, string_type) for item in value]
+            else:
+                return string_type(value)
+
         ret = value
         if isinstance(spec, AttributeSpec):
             if 'text' in spec.dtype:
                 if spec.shape is not None or spec.dims is not None:
-                    ret = list(map(str, value))
+                    ret = __apply_string_type(value, str)
                 else:
                     ret = str(value)
         elif isinstance(spec, DatasetSpec):
@@ -617,10 +623,8 @@ class ObjectMapper(metaclass=ExtenderMeta):
                     def string_type(x):
                         return x.isoformat()  # method works for both date and datetime
                 if string_type is not None:
-                    if spec.shape is not None and len(spec.shape) > 1:
-                        ret = [[string_type(item) for item in sublist] for sublist in value]
-                    elif spec.shape is not None or spec.dims is not None:
-                        ret = list(map(string_type, value))
+                    if spec.shape is not None or spec.dims is not None:
+                        ret = __apply_string_type(value, string_type)
                     else:
                         ret = string_type(value)
                     # copy over any I/O parameters if they were specified
--- a/tests/unit/test_io_hdf5_h5tools.py
+++ b/tests/unit/test_io_hdf5_h5tools.py
@@ -164,6 +164,14 @@ class H5IOTest(TestCase):
         dset = self.f['test_dataset']
         self.assertTrue(np.all(dset[:] == a))
 
+    def test_write_dataset_lol_strings(self):
+        a = [['aa', 'bb'], ['cc', 'dd']]
+        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', a, attributes={}))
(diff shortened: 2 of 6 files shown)


src/hdmf/build/objectmapper.py
```diff
@@ -617,7 +617,9 @@ def __convert_string(self, value, spec):
                     def string_type(x):
                         return x.isoformat()  # method works for both date and datetime
                 if string_type is not None:
-                    if spec.shape is not None or spec.dims is not None:
+                    if spec.shape is not None and len(spec.shape) > 1:
+                        ret = [[string_type(item) for item in sublist] for sublist in value]
```
rly: This is fine for 2D arrays but I think it will fail for higher-dimensional arrays. Can you add a test for a 3D array? You could convert value to a numpy array and change the type. Or do the conversion recursively on the list elements.

stephprince: Good point. Converting it to a numpy array and changing the type didn't work as well with the isodatetimes so I added a recursive conversion.

stephprince:
@rly I realized that I had only updated the string conversion for datasets so I updated `__convert_string` to accommodate multidimensional string arrays as attributes as well. 

For testing, I added an additional `attr_array` argument to the `Bar` class constructor, but I'm not sure if that was the best approach or if there's a better way to test it since the `Bar` class seemed to be used in multiple places.

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -164,6 +164,30 @@ def test_write_dataset_list(self):
...
 
+    def test_write_dataset_lol_strings(self):
+        a = [['aa', 'bb'], ['cc', 'dd']]
+        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', a, attributes={}))
+        dset = self.f['test_dataset']
+        decoded_dset = [[item.decode('utf-8') if isinstance(item, bytes) else item for item in sublist]
+                        for sublist in dset[:]]
+        self.assertTrue(decoded_dset == a)
+
+    def test_write_dataset_list_compound_datatype(self):
```
stephprince, line 175: There was missing code coverage for compound data types in the `__list_fill__` method in `src/hdmf/backends/hdf5/h5tools.py` so I added this test that I think should cover that case.

rly: Looks good!

rly (approved):

PR #1171, opened 2024-08-15
Fix conversion of Zarr string dataset to HDF5

oruebel:
## Motivation

Fix https://github.com/hdmf-dev/hdmf-zarr/issues/211

Zarr does not natively support variable length strings, instead the data is stored as objects with an encoder. When converting from Zarr to HDF5 the `ObjectMapper` tries to change the dtype in some cases for text datasets to unicode, which in turn can cause reading the data from Zarr to fail. This PR changes this behavior to: 1) not change the dtype for Zarr string datasets, and 2) detect the correct dtype to use in HDF5 in the backend. 

## Checklist

- [x] Did you update `CHANGELOG.md` with your changes?
- [X] Does the PR clearly describe the problem and the solution?
- [X] Have you reviewed our [Contributing Guide](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst)?
- [X] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged?

PR #1175, opened 2024-08-20
Modification to Fix conversion of Zarr string dataset to HDF5

rly:
Proposed changes to #1171.

It would be nice if `HDF5IO` and the `ObjectMapper` can accept any Zarr array just like it accepts any Numpy array, instead of accepting only Zarr arrays that have been written with hdmf-zarr.

### DIFF
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -20,6 +20,16 @@ from ..spec import Spec, AttributeSpec, DatasetSpec, GroupSpec, LinkSpec, RefSpe
 from ..spec.spec import BaseStorageSpec
 from ..utils import docval, getargs, ExtenderMeta, get_docval, get_data_shape
 
+try:
+    from zarr import Array as ZarrArray
+    ZARR_INSTALLED = True
+except ImportError:
+    ZARR_INSTALLED = False
+
+
+def _is_zarr_array(value):
+    return ZARR_INSTALLED and isinstance(value, ZarrArray)
+
 _const_arg = '__constructor_arg'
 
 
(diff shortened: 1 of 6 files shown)


src/hdmf/build/objectmapper.py
```diff
@@ -206,17 +216,19 @@ def convert_dtype(cls, spec, value, spec_dtype=None):  # noqa: C901
         spec_dtype_type = cls.__dtypes[spec_dtype]
         warning_msg = None
         # Numpy Array or Zarr array
-        if (isinstance(value, np.ndarray) or
-                (hasattr(value, 'astype') and hasattr(value, 'dtype'))):
+        # NOTE: Numpy < 2.0 has only fixed-length strings.
+        # Numpy 2.0 introduces variable-length strings (dtype=np.dtypes.StringDType()).
+        # HDMF does not yet do any special handling of numpy arrays with variable-length strings.
+        if isinstance(value, np.ndarray) or _is_zarr_array(value):
```
oruebel: The basic approach seems fine. A couple of thoughts:
- I think instead of the `if isinstance(value, np.ndarray) or _is_zarr_array(value)` it would be useful to split this case into `if` and `elif` cases to treat numpy and zarr separately here. I think that would be more readable. 
- I'd move the `_is_zarr_array` method ot the `utils` or `data_utils` 
- Alternatively to this approach we could also use `hasattr(value, filters)` to distinguish been Zarr and numpy arrays, because numpy arrays don't have a `filters` attribute which is used in Zarr to store all the filters, including the encoder for strings. 
-

rly: done

rly:
Tests are failing because the base PR is too out of date. I will merge and we can review in the base PR.

PR #1176, opened 2024-08-20
Write scalar datasets with compound data type

stephprince:
## Motivation

This came up in https://github.com/NeurodataWithoutBorders/pynwb/pull/1770, where `ElectrodeGroup.position` is a scalar dataset with a compound data type. There was an issue with write, so the written dataset was not scalar and/or values would be duplicated. 

Once this PR is incorporated, the related pynwb PR needs additional changes to the ElectrodeGroup position validation so that the position input is converted to a scalar with a compound data type. For example, something like:

```python
if args_to_set['position']:
        if len(np.shape(args_to_set['position'])) > 1 or \
        (len(np.shape(args_to_set['position'])) == 1 and len(args_to_set['position']) != 3):
            raise ValueError(f"ElectrodeGroup position argument must have three elements: x, y, z,"
                            f"but received: {args_to_set['position']}")
        else:
            args_to_set['position'] = np.array(args_to_set['position'], dtype=[('x', float), ('y', float), ('z', float)])

```

## How to test the behavior?
```python
from pynwb import NWBFile, NWBHDF5IO
from uuid import uuid4
from datetime import datetime

nwbfile = NWBFile(
    session_description="no description.",
    session_start_time=datetime.now(),
    identifier=str(uuid4()),
)

device = nwbfile.create_device(name="Device", description="no description.")

electrode_group = nwbfile.create_electrode_group(
    name="ElectrodeGroup",
    description="no description.",
    device=device,
    location="unknown",
    position=[(1., 2., 3.)],  # --> dataset shape = (1,)
    # position=(1, 2, 3)      # --> dataset shape = (3,)
)
)

with NWBHDF5IO('test.nwb', 'w') as io:
    io.write(nwbfile)
    
with NWBHDF5IO('test.nwb', 'r') as io:
    read_nwbfile = io.read()
    
    print(read_nwbfile.electrode_groups['ElectrodeGroup'].position.shape)  # (1,)
```

## Checklist

### DIFF
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -632,7 +632,7 @@ class Container(AbstractContainer):
             # if isinstance(v, DataIO) or not hasattr(v, '__len__') or len(v) > 0:
             if hasattr(v, '__len__'):
                 if isinstance(v, (np.ndarray, list, tuple)):
-                    if len(v) > 0:
+                    if (hasattr(v, 'shape') and len(np.shape(v)) == 0) or len(v) > 0:
                         template += "  {}: {}\n".format(k, self.__smart_str(v, 1))
                 elif v:
                     template += "  {}: {}\n".format(k, self.__smart_str(v, 1))
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -629,12 +629,8 @@ class Container(AbstractContainer):
             template += "\nFields:\n"
         for k in sorted(self.fields):  # sorted to enable tests
             v = self.fields[k]
-            # if isinstance(v, DataIO) or not hasattr(v, '__len__') or len(v) > 0:
             if hasattr(v, '__len__'):
-                if isinstance(v, (np.ndarray, list, tuple)):
-                    if (hasattr(v, 'shape') and len(np.shape(v)) == 0) or len(v) > 0:
-                        template += "  {}: {}\n".format(k, self.__smart_str(v, 1))
-                elif v:
+                if isinstance(v, (np.ndarray, list, tuple)) or v:
                     template += "  {}: {}\n".format(k, self.__smart_str(v, 1))
             else:
                 template += "  {}: {}\n".format(k, v)
--- a/tests/unit/test_io_hdf5_h5tools.py
+++ b/tests/unit/test_io_hdf5_h5tools.py
@@ -806,7 +806,7 @@ class H5IOTest(TestCase):
         self.io.close()
         with HDF5IO(self.path, 'r') as io:
             bldr = io.read_builder()
-            np.array_equal(bldr['test_dataset'].data[()], a)
+            np.testing.assert_array_equal(bldr['test_dataset'].data[()], a)
 
 
 class TestRoundTrip(TestCase):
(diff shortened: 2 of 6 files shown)


tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -787,6 +797,17 @@ def test_read_str(self):
...
+    def test_read_scalar_compound(self):
+        cmpd_dtype = np.dtype([('x', np.int32), ('y', np.float64)])
+        a = np.array((1, 0.1), dtype=cmpd_dtype)
+        self.io.write_dataset(self.f, DatasetBuilder('test_dataset', a,
+                                                     dtype=[DtypeSpec('x', doc='x', dtype='int32'),
+                                                            DtypeSpec('y', doc='y', dtype='float64')]))
+        self.io.close()
+        with HDF5IO(self.path, 'r') as io:
+            bldr = io.read_builder()
+            np.array_equal(bldr['test_dataset'].data[()], a)
```
```suggestion
            np.testing.assert_array_equal(bldr['test_dataset'].data[()], a)
```

src/hdmf/container.py
```diff
@@ -632,7 +632,7 @@ def __repr__(self):
             # if isinstance(v, DataIO) or not hasattr(v, '__len__') or len(v) > 0:
             if hasattr(v, '__len__'):
                 if isinstance(v, (np.ndarray, list, tuple)):
-                    if len(v) > 0:
+                    if (hasattr(v, 'shape') and len(np.shape(v)) == 0) or len(v) > 0:
```
rly: What does this whole block from 633-638 do? The `template += ...` is the same. It looks odd.

It looks like this if statement adds to the template if `v` has more than one element or is a scalar compound. But what if `v` is not a compound dtype and has only one element? Does nothing happen?

stephprince: My understanding of the original code was that it will add to the template as long as it's not an empty list/array/tuple. However for a scalar with a compound data type, `hasattr(v, '__len__')` is True but `len(v)` will error with `TypeError: len() of unsized object`.

I think if `v` is not a compound type and has at least one element it will be added to the template.... I can add a test for that to make sure?

rly: Even if the value is a list/array/tuple that is empty, I think we should still add it to the template. Maybe that will clutter the display, but I think it is still useful to see. What do you think? That's a bit tangential to this PR but I think it's a good improvement. It would remove the modified if statement all together.

stephprince: I think it might clutter things up... removing the if statement will make an nwbfile look like this:

Screenshot 2024-08-21 at 3 39 58 PM

vs. with the if statement:
Screenshot 2024-08-21 at 3 40 40 PM

stephprince: Actually disregard the previous comment, I had condensed the if statements too much. I agree if it's an empty list/array/tuple it could be useful to display.

Simplifying to this is what you're suggesting then?
```python
if isinstance(v, (np.ndarray, list, tuple)) or v:
    template += "  {}: {}\n".format(k, self.__smart_str(v, 1))
```

rly: Yes, exactly. Great! Let's do that.

rly (approved):

rly:
Looks good to me!

PR #1180, opened 2024-08-27
Make HDF5 datasets expandable by default (#1158)

### LINKED ISSUE #1163: [Bug]: Dimension Labels and Expandable Datasets Edge Cases
### What happened?

1. Within `__get_dimension_labels_from_spec` (or `__get_spec_info` depending on if #1158 has been merged) uses get_data_shape to get the shape. But this does not work when we have dtype defined:
```
data = [(1, 2, 3.)]
dtype=[('x', '<u4'), ('y', '<u4'), ('weight', '<f4')])
```
get_data_shape will return (1,3)
However, when writing the data we use this for the shape:
```
elif isinstance(dtype, np.dtype):
            data_shape = (len(data),)
```
which is (1,)

2. Both #1158 and the current implementation of dimension_labels do not support the case of "untyped, named dataset":
```
elif spec.data_type_def is None and spec.data_type_inc is None:  # untyped, named dataset
                if spec.name in builder.datasets:
                    sub_builder = builder.datasets[spec.name]
                    self.logger.debug("        Retrieving existing DatasetBuilder '%s' for spec name %s and adding "
                                      "attributes" % (sub_builder.name, repr(spec.name)))
                else:
                    self.logger.debug("        Converting untyped dataset for spec name %s to spec dtype %s"
                                      % (repr(spec.name), repr(spec.dtype)))
                    try:
                        data, dtype = self.convert_dtype(spec, attr_value)
                    except Exception as ex:
                        msg = 'could not convert \'%s\' for %s \'%s\''
                        msg = msg % (spec.name, type(container).__name__, container.name)
                        raise BuildError(builder, msg) from ex
                    self.logger.debug("        Adding untyped dataset for spec name %s and adding attributes"
                                      % repr(spec.name))
                    sub_builder = DatasetBuilder(spec.name, data, parent=builder, source=source, dtype=dtype)
                    builder.set_dataset(sub_builder)
```

### Steps to Reproduce

```python
Above.
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.12

### Package Versions

_No response_

mavaylon1:
## Motivation

Fix #1158  
Fix #1163

Add expandable default for datasets.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
@rly This PR is essentially done. I need to go through the pynwb side to make sure what is failing we can ignore through future development. I also want to run the neuroconv tests to make sure those pass as well.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1365,6 +1389,11 @@ def __list_fill__(cls, parent, name, data, options=None):
             data_shape = (len(data),)
         else:
             data_shape = get_data_shape(data)
+        if expandable:
+            # Don't override existing settings
+            if 'maxshape' not in io_settings:
+                if matched_spec_shape is not None:
+                    io_settings['maxshape'] = matched_spec_shape
```
oruebel: Do `matched_spec_shape` and `expandable` actually need to be passed to `__list_fill__`? It looks like the code update the `maxshape` is already being injected in `write_dataset` before `__list_fill__` is being called.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1131,7 +1143,13 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
                     msg = 'cannot add %s to %s - could not determine type' % (name, parent.name)
                     raise Exception(msg) from exc
-                dset = parent.require_dataset(name, shape=(len(data),), dtype=_dtype, **options['io_settings'])
+                io_settings = options['io_settings']
+                if expandable:
+                    # Don't override existing settings
+                    if 'maxshape' not in io_settings:
+                        if matched_spec_shape is not None:
+                            io_settings['maxshape'] = matched_spec_shape
+                dset = parent.require_dataset(name, shape=(len(data),), dtype=_dtype, **io_settings)
```
oruebel: This code snippet currently appears 3 times in the code. I'm just wondering whether it may be worth making that a helper function. Your choice, just a thought.

src/hdmf/build/objectmapper.py
```diff
@@ -886,56 +892,131 @@ def __check_dset_spec(self, orig, ext):
...
+        Determine which allowed shapes match the given data shape.
+
+        Parameters
+        ----------
+        data_shape : tuple or list of int
+            The shape of the data being validated.
+        spec_shape : list of list
+            A list of alternative shapes, where each element is itself a shape
+            specification (a list of dimension lengths or ``None``, where ``None``
+            means any length is allowed for that dimension).
```
oruebel, line 906: Is `spec_shape` allowed to be an empty list, and does an empty `spec_shape` imply that no restrictions on the shape are defined in the spec? 

If `spec_shape` is an empty list, then the function currently would return an empty list, i.e., the returned shape would not be compatible with the `data_shape`.  I think instead it probably should return a `[None] * len(data_shape)` to match the dimensionality of the `data_shape` and to make the data expandable in all dimensions.

src/hdmf/backends/hdf5/h5tools.py
rly: It is for one of the cases. You're right, though. Let me pull out this code so that we do it once at the beginning of `write_dataset`.

src/hdmf/build/objectmapper.py
rly: Defining the spec shape as an empty list would raise an error before getting here at `spec_shape[0]`. `shape: []` is also not valid in the schema language. Either a non-empty list should be provided or it should be set to null/None. I'll add a guard in DatasetSpec.__init__ and AttributeSpec.__init__ to check if `shape` is empty, so it doesn't get this far.

oruebel (approved):

ehennestad:
@mavaylon1 @rly @bendichter 

This change causes a deviation in the default write behaviour of datasets in PyNWB and MatNWB (MatNWB default is to write non-extendable datasets).

It is not clear to me from this PR (or linked PRs) why this change was made. It is also not clear to me if MatNWB should be updated to mimic the new default write behaviour of HDMF/PyNWB.

oruebel:
> It is not clear to me from this PR (or linked PRs) why this change was made.

PyNWB is used more heavily in analysis workflows where data is being modified at multiple stages. To support this, datasets often need to be expandable. However, this requires users to manually specify chunking setting for every dataset. One main use case here is the need to add rows to DynamicTables, which requires all VectorData columns to be expandable. Changing the default simplifies this. 

> It is also not clear to me if MatNWB should be updated to mimic the new default write behaviour of HDMF/PyNWB.

Either default behavior is valid. NWB does not enforce details of the storage layout. As such, the specific default behavior is in my a view a question of what makes most sense for the particular API and how it is being used. I think for MatNWB the main uses may be to convert data to NWB and read existing datasets. For conversion of data, expandablity of data is often not necessary since it is usually implemented as a single script where all data is being input at once. As such, users know beforehand which specific datasets they may need to configure with chunking for their conversion. I.e., I think it is Ok in this case for the default behavior of MatNWB to remain because: 1) users can configure the data to use chunking if they want to (i.e., there is an existing process to create expandable dataset with MatNWB), 2) the default behavior is (or should be) documented in the docs, 3) the default behavior should be what makes most sense for the users of the API (which may be different for MatNWB). 

That being said. I think there could be utility in making datasets expandable in MatNWB by default. E.g., to make the data more flexible for use with downstream analyses; to optimize chunking by default to accelerate cloud read; or to enhance consistency in behavior with PyNWB.

I think a decision to change or keep the current default behavior in MatNWB will need to weigh the questions of "Which behavior is most appropriate for MatNWB users?" and if a different default behavior may be preferred "How much effort would be required to change and maintain a different default behaviors?"

ehennestad:
@oruebel Thanks for clarifying.

We have recently added some utility functions in MatNWB to make it easy for users to apply chunking (and thus also create expandable datasets) for all datasets on export, so I think the effort to change the default behaviour is already small.

oruebel:
> We have recently added some utility functions in MatNWB to make it easy for users to apply chunking (and thus also create expandable datasets) for all datasets on export, so I think the effort to change the default behaviour is already small.

That's great @ehennestad . @bendichter do you have thoughts on whether the current default behavior for chunking should be updated or whether the current should remain as.

PR #1194, opened 2024-10-01
Remove mamba key from GH Actions

rly:
## Motivation

All conda tests are failing. It seems to be because the experimental `mamba: *` config is no longer supported / is broken in the setup-miniconda github action. In any case, the latest version of conda uses the libmamba solver by default so we should not need to specify `mamba: *` anymore.

~I do not think this needs a changelog entry, but happy to add one.~ I added one.

## Checklist

### DIFF
--- a/.github/workflows/run_all_tests.yml
+++ b/.github/workflows/run_all_tests.yml
@@ -165,13 +165,12 @@ jobs:
           auto-update-conda: true
           python-version: ${{ matrix.python-ver }}
           channels: conda-forge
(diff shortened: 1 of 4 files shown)


rly:
Total runtime is about the same here vs when the workflow was last succeeding, so I think this change does not affect the conda solver speed.

mavaylon1 (approved):

mavaylon1:
That's nice libmamba is the default now.

PR #1201, opened 2024-11-06
Small patch to html repr in #1100 

h-mayorquin:
## Motivation

There are some edge cases that #1100 did not take into account. I am submitting this patch for discussion, happy to close if a better solution comes.

The problem is that the code here assumes that things that have an IO are hdf5 datasets:

https://github.com/hdmf-dev/hdmf/blob/be602e51804922e01bcd0e6f89b23873f9e13b4b/src/hdmf/backends/hdf5/h5tools.py#L1612-L1623

https://github.com/hdmf-dev/hdmf/blob/be602e51804922e01bcd0e6f89b23873f9e13b4b/src/hdmf/container.py#L754-L763

But sometimes, this assumption is false. For example, the starting frames of an ImageSeries are a numpy object even after they are written. Maybe there are more such cases?

## How to test the behavior?

The following code generates an error when using dev.

```python
from pynwb.image import ImageSeries

nwbfile = mock_NWBFile()

series = ImageSeries(name="ImageSeries", description="", external_file=["test"], rate=0.1)
nwbfile.add_acquisition(series)


from pynwb import NWBHDF5IO

nwbfile_path = "./test_nwb.nwb"
with NWBHDF5IO(nwbfile_path, 'w') as io:
    io.write(nwbfile)
    
io = NWBHDF5IO(nwbfile_path, 'r')
nwbfile_read = io.read()
nwbfile_read

AttributeError: 'numpy.ndarray' object has no attribute 'id'

```

## Checklist


stephprince:
@rly this was the case we discussed in person yesterday, do you know if there any other edge cases we should consider when a `Container` would have a `read_io` object but the array would not be an hdf5 dataset?

stephprince (commented):
@h-mayorquin could you add a comment to this line indicating that even if the object has a `read_io` the array may still be a numpy array, so `generate_dataset_html` will be called even if it is not an HDF5/Zarr dataset.

  https://github.com/hdmf-dev/hdmf/blob/be602e51804922e01bcd0e6f89b23873f9e13b4b/src/hdmf/container.py#L758

At some point we may want to add a `HDMFIO.is_dataset` method, so that `generate_dataset_html` is not unnecessarily called on a numpy array, but I think a comment is sufficient for now. Other than that, Iooks good to me!

h-mayorquin:
@stephprince 
Done
But I think I wrote a more robust solution here #1206.
I branched from this one though, so this can be merged first and then we can discuss the other in more detail.

stephprince:
Sounds good, let's merge this for now and I will take a look at the other solution you proposed later.

Since this is a small patch to #1100 I think it is ok to not include a changelog update.

stephprince (approved):

PR #1206, opened 2024-11-09
Route array representation for HTML

h-mayorquin:
Improved version of #1201


src/hdmf/container.py
```diff
@@ -753,12 +753,23 @@ def _generate_field_html(self, key, value, level, access_code):
...
+        is_data_io = isinstance(array, DataIO)
+
+        if is_numpy_array:
             array_info_dict = get_basic_array_info(array)
             repr_html = generate_array_html_repr(array_info_dict, array, "NumPy array")
+        elif is_data_io:
+            array_info_dict = get_basic_array_info(array.data)
+            repr_html = generate_array_html_repr(array_info_dict, array.data, "DataIO")
+        elif it_was_read_with_io:  # The backend handles the representation
+            read_io = self.get_read_io()  # Note that sometimes numpy array have a read_io attribute
```
oruebel: In which cases do numpy arrays have a read_io attribute?

h-mayorquin: Starting frames in an `ImageSeries` are numpy even if the object was read with NWBHDF5IO. See here for the previous discussion:

https://github.com/hdmf-dev/hdmf/pull/1201

src/hdmf/container.py
```diff
@@ -753,12 +753,23 @@ def _generate_field_html(self, key, value, level, access_code):
     def _generate_array_html(self, array, level):
         """Generates HTML for array data"""
 
-        read_io = self.get_read_io()  # if the Container was read from file, get IO object
-        if read_io is not None: # Note that sometimes numpy array have a read_io attribute
-            repr_html = read_io.generate_dataset_html(array)
-        else:
+        is_numpy_array = isinstance(array, np.ndarray)
+        it_was_read_with_io = self.get_read_io() is not None
```
oruebel: `it_was_read_with_io` should capture most cases to determine if we need to route to the backend.  However, `it_was_read_with_io` applies to the whole container, but does not detect any changes the user may have done after reading.  As such, one case where this will potentially result in routing to the backend when we shouldn't is when the user replaces the dataset after read. However, I think adding a comment here to explain this should be sufficient.

src/hdmf/container.py
oruebel: > Starting frames in an `ImageSeries` are numpy even if the object was read with NWBHDF5IO.

Makes sense. This is because `starting_frame` is an Attribute in HDF5 on `external_file` rather and not at Dataset. Attributes are read in while Datasets are loaded lazily. 

Just to clarify, the numpy array itself does not have read_io, but the Container object that owns it. This is similar to the case I mentioned below. The Container has a read_io but the dataset we are looking at has been modified (instead of being the backend dataset). In the case of `starting_frame` the "modification" is that it was read into memory.

src/hdmf/container.py
h-mayorquin: Ah, very good point about the user modifying the data on the fly.

The nature of the if-elif-else would catch that in most cases I think. Let's enumerate:
* If the user modified the data to be a numpy array it would be caught before in the first if.
* If the user modified the data to be DataIO it would be catch before in the second elif.
*If the user modified the data to be an Iterator then I think this will get here.

The Iterator is the tricky case, I was thinking already that maybe they should have a representation of their own. Do `AbstractDataChunkIterator` children have a shape and dtype attributes? If so, they will get here as these lines send things here for a representation.
```
 is_array_data = hasattr(value, "shape") and hasattr(value, "dtype")
```
(see the diff)

src/hdmf/container.py
h-mayorquin: Oh, yes, I did not realize that the `get_io` method goes to the parent of the object though.

https://github.com/hdmf-dev/hdmf/blob/ea6504f6c49e78d7b3d0effccef2bd1fe19520ab/src/hdmf/container.py#L344-L360

From that perspective the comment is wrong, should be updated.

src/hdmf/container.py
oruebel: > Do `AbstractDataChunkIterator` children have a shape and dtype attributes?

The iterators have a `dtype` but not a `shape`, because the final `shape` may not be known. However, they do have a `maxshape`. One option may be to add a html representation to `AbstractDataChunkIterator`

h-mayorquin: Yes, I think adding a html represeentation to AbstractDataChunkIterator is something that I would like to do at some point if you are OK with it. Discussion for another PR though : )

oruebel: > Yes, I think adding a html represeentation to AbstractDataChunkIterator is something that I would like to do at some point if you are OK with it. Discussion for another PR though : )

Sure, that sounds good

src/hdmf/container.py
h-mayorquin: I made an attempt to improve the comments.

src/hdmf/container.py
h-mayorquin: I added an improvement comment that states this.

src/hdmf/container.py
```diff
@@ -751,14 +754,29 @@ def _generate_field_html(self, key, value, level, access_code):
...
+            array_info_dict = get_basic_array_info(array.data)
+            repr_html = generate_array_html_repr(array_info_dict, array.data, "DataIO")
+        elif it_was_read_with_io:
+            # The backend handles the representation here. Two special cases worth noting:
+            # 1. Array-type attributes (e.g., start_frame in ImageSeries) remain NumPy arrays
+            #    even when their parent container has an IO
+            # 2. Data may have been modified after being read from storage
+            read_io = self.get_read_io()
+            repr_html = read_io.generate_dataset_html(array)
+        else:  # Not sure which object could get here
```
oruebel, line 760: Based on our discussion above, I believe any `AbstractDataChunkIteator` would currently go to the else statement. Also, if the data was specified as a `list` or `tuple` , I think it would end up in `else`

src/hdmf/container.py
oruebel: Looks good.

src/hdmf/container.py
h-mayorquin: Those objects will fail to enter this function:

https://github.com/h-mayorquin/hdmf/blob/4a0dbd9b635add255e49803be5d03d0e6600d89c/src/hdmf/container.py#L730-L733

oruebel: Thanks for the clarification. I did not see that there was additional logic outside of the function for this.

src/hdmf/container.py
```diff
@@ -735,14 +738,29 @@ def _generate_field_html(self, key, value, level, access_code):
...
     def _generate_array_html(self, array, level):
-        """Generates HTML for array data"""
+        """Generates HTML for array data (e.g., NumPy arrays, HDF5 datasets, Zarr datasets and DataIO objects)."""
 
-        read_io = self.get_read_io()  # if the Container was read from file, get IO object
-        if read_io is not None: # Note that sometimes numpy array have a read_io attribute
-            repr_html = read_io.generate_dataset_html(array)
-        else:
+        is_numpy_array = isinstance(array, np.ndarray)
+        it_was_read_with_io = self.get_read_io() is not None
```
```suggestion
        read_io = self.get_read_io()
        it_was_read_with_io = read_io is not None
```

This is just to avoid having to call `self.get_read_io` twice. The `io` object is actually only stored on the root container (i.e., `NWBFile`). As such `get_read_io` has to iterate through the parent hierarchy to find the `io` object. This probably just minor, but I think it could help in some cases with performance to avoid repeating this search.

src/hdmf/container.py
```diff
@@ -735,14 +738,29 @@ def _generate_field_html(self, key, value, level, access_code):
...
             repr_html = generate_array_html_repr(array_info_dict, array, "NumPy array")
+        elif is_data_io:
+            array_info_dict = get_basic_array_info(array.data)
+            repr_html = generate_array_html_repr(array_info_dict, array.data, "DataIO")
+        elif it_was_read_with_io:
+            # The backend handles the representation here. Two special cases worth noting:
+            # 1. Array-type attributes (e.g., start_frame in ImageSeries) remain NumPy arrays
+            #    even when their parent container has an IO
+            # 2. Data may have been modified after being read from storage
+            read_io = self.get_read_io()
```
```suggestion
```

This is a follow-up to the above suggestion, removing the extra call to `self.get_read_io` and using the object retrieved at the beginning of the function.

oruebel (changes requested):
I suggested a minor enhancement to avoid repeat calls to `get_read_io`. Otherwise this looks good to me 👍

h-mayorquin:
Thakns @oruebel . I submited the suggestions

oruebel (approved):

PR #1209, opened 2024-11-14
Remove Python 3.8 Support and Add 3.13 Support

### LINKED ISSUE #1207: [Feature]: Python 3.8

Python 3.8 is deprecated. Update the pyproject.toml

Above

Yes.

mavaylon1:
## Motivation

Remove Python 3.8 Support
Fix #1207 
## Checklist


requirements-min.txt
```diff
@@ -1,15 +1,10 @@
...
-# schemasheets==0.1.23; python_version >= "3.9"
-# oaklib==0.5.12; python_version >= "3.9"
-# pyyaml==6.0.1; python_version >= "3.9"
+h5py==3.11.0
+jsonschema==4.22.0
+numpy==1.26.4
+pandas==2.2.2;
+ruamel.yaml==0.18.2
+scipy==1.14.0; python_version >= "3.10"
+scipy==1.11.3; python_version < "3.10"
```
rly: This file is for minimum requirements. They should be kept as low as possible.

requirements.txt
```diff
@@ -1,10 +1,7 @@
 # pinned dependencies to reproduce an entire development environment to use HDMF
 h5py==3.11.0
-importlib-resources==6.1.0; python_version < "3.9"  # TODO: remove when minimum python version is 3.9
 jsonschema==4.22.0
-numpy==1.26.4  # TODO: numpy 2.0.0 is supported by hdmf but incompatible with pandas and scipy
-pandas==2.2.2; python_version >= "3.9"
```
rly, line 6: I think pandas should not be removed here.

rly:
Can you also add python 3.13 to the tests?

requirements-min.txt
mavaylon1: The ones that were there had issues with 3.9. Let me try a few version changes to get a more forgiving minimum.

mavaylon1: @rly I made a more forgiving set of minimums. I can't seem to get h5py to stay at 2.10 so I have it at 3.1 as the lowest. I did not do 3.0 because they came out with 3.1 a few weeks after and that was steady for months. I will add python 3.13 support.

rly (approved):

PR #1230, opened 2025-01-13
Update dates and how optional deps are specified and tested

rly:
## Motivation
- Updated copyright dates
- Created optional dependency groups in `pyproject.toml` and update GitHub Actions workflows to use those instead of requirements files. 
- Stopped using pinned dependencies in the docs and testing. These are not necessary for library testing, confuse new users and developers, and add maintenance burden. Current dependencies are stable enough that they need not be pinned and users can report the libraries they use.
- Stopped redundant testing using a conda environment.
- Adopted changelog format conventions: https://keepachangelog.com/en/1.1.0/

## Checklist


rly:
`linkml` does not yet support Python 3.13 so we will need to use Python 3.12 to test optional dependencies.

pyproject.toml
```diff
@@ -40,15 +40,39 @@ dependencies = [
 dynamic = ["version"]
 
 [project.optional-dependencies]
-tqdm = ["tqdm>=4.41.0"]
-zarr = ["zarr>=2.12.0,<3"]
-termset = [
+# all optional dependencies
+full = [
```
mavaylon1: Could we change the name from full to something else? "opt" maybe.

scripts/check_py_support.py
```diff
@@ -0,0 +1,204 @@
...
+and optional dependencies, checking their trove classifiers for explicit version support.
+
+The script provides:
+- Grouped output of supported and unsupported packages
+- Latest supported Python version for packages without explicit support
+- Error reporting for packages that cannot be checked
+- Summary statistics of compatibility status
+
+Usage:
+    python check_py_support.py
```
mavaylon1: This won't work because it needs to find pyproject.toml. I would change this to `python scripts/check_py_support.py` and you would need to run it at the root level.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -65,6 +56,7 @@ body:
         - "3.11"
         - "3.10"
         - "3.9"
+        - "other"
```
mavaylon1: What is this for?

pyproject.toml
rly: spikeinterface and neuroconv use "full":
- https://github.com/catalystneuro/neuroconv/blob/main/pyproject.toml#L336
- https://github.com/SpikeInterface/spikeinterface/blob/main/pyproject.toml#L94

We don't need to follow suit, though. 

Neither "opt" and "full" are very descriptive. I was thinking "pip install hdmf[full]" because it is like the full user installation, which feels a little descriptive than "pip install hdmf[opt]" - install hdmf with all optional dependencies. But either is OK.

scripts/check_py_support.py
rly: Good point

rly:
```suggestion
    python scripts/check_py_support.py
```

.github/ISSUE_TEMPLATE/bug_report.yml
rly: We have this in pynwb. It's for when users report an error with an older or newer version of Python for whatever reason. https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/ISSUE_TEMPLATE/bug_report.yml

scripts/check_py_support.py
```diff
@@ -0,0 +1,204 @@
...
+and optional dependencies, checking their trove classifiers for explicit version support.
+
+The script provides:
+- Grouped output of supported and unsupported packages
+- Latest supported Python version for packages without explicit support
+- Error reporting for packages that cannot be checked
+- Summary statistics of compatibility status
+
+Usage:
+    python scripts/check_py_support.py
```
rly, line 16:
```suggestion
    # Run this command from the root of the repo
    python scripts/check_py_support.py
```

pyproject.toml
mavaylon1: I don't like full because it implies including dev and doc. "select" "core" "recommended" "core-opt". Those are some ideas

.github/ISSUE_TEMPLATE/bug_report.yml
mavaylon1: Hmmm what do you think about "newer" because we wouldn't help if they are using an older python. That's the point of dropping support for it.

rly: Sure. I'll change it.

pyproject.toml
rly: "select" is not very informative. "core" suggests a minimal installation. "recommended" could work but is not very informative. "core-opt" is confusing in my opinion. I like "opt" better than those.

fastapi uses "standard" to say install all optional dependencies, and it recommends users install with `pip install "fastapi[standard]"`: https://fastapi.tiangolo.com/#dependencies . I believe `pip install fastapi` is then a quite minimal installation. I don't think we're ready to go that far yet, though it would be useful for handling https://github.com/NeurodataWithoutBorders/pynwb/issues/2010

What if we keep "tqdm", "zarr", and "termset" as the depedency group names, and wherever "full" is currently used in this PR, replace it with "tqdm, zarr, termset"?

Removal of `requirements-opt.txt` should only affect developers. Developers should now run `pip install -e .[dev,doc,tqdm,zarr,termset]` or simply `pip install -e ".[all]"`

mavaylon1: I like "standard". the other option is also good but a little long. I have no issues with either, so I will leave it to you to choose!

rly: OK. I'm not ready to commit to "standard" just yet. Maybe when we bump pandas and scipy from the core requirements. I'll update the PR to use "tqdm", "zarr", and "termset" as the depedency group names for now.

mavaylon1 (approved):

PR #1246, opened 2025-02-08
Minor housekeeping updates

rly:
## Motivation

- Update ruff action to non-deprecated one
- Update github-app-token action to official github one
- Remove `requirements-min.txt` in favor of an optional dependency group in `pyproject.toml` and update docs accordingly
- Update `pyproject.toml` `license` key to correct format
- Add back ruff mccabe complexity check which was accidentally removed in transition to ruff
- Refactor some code to reduce complexity and added complexity check exceptions in other code
- Clean up ruff config

## How to test the behavior?
```
```

## Checklist


mavaylon1:
@rly overall this looks good. There are coverage gaps. If you are okay with that, then I will approve.

rly:
Looking closely at the coverage report, I see that all the lines that now lack coverage had previously lacked coverage. We can add the missing tests in a separate PR, but I think that should not need to block this one.

mavaylon1 (approved):

PR #1248, opened 2025-02-16
Slight refactor to Vector Index get

mavaylon1:
## Motivation

https://github.com/hdmf-dev/hdmf-zarr/issues/237
Using to_dataframe in Zarr is slower than HDF5, most likely due to how we handle references. The change is gathering all the data at once to resolve, rather than multiple I/O reads iteratively.

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -203,7 +203,10 @@ class VectorIndex(VectorData):
                     arg = np.where(arg)[0]
                 indices = arg
             ret = list()
-            if len(indices) > 0: # This is for test_to_hierarchical_dataframe_empty_tables
+            if len(indices) > 0:
+                # Note: len(indices) == 0 for test_to_hierarchical_dataframe_empty_tables.
+                # This is an edge case test for to_hierarchical_dataframe() on empty tables.
+                # When len(indices) == 0, ret is expected to be an empty list, defiend above.
                 try:
                     data = self.target.get(slice(None),  **kwargs)
                     slices = [self.__get_slice(i) for i in indices]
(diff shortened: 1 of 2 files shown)


alejoe91:
@mavaylon1 tested again on my end an this dramatically speeds things up :)

src/hdmf/common/table.py
```diff
@@ -199,8 +203,20 @@ def get(self, arg, **kwargs):
...
-                ret.append(self.__getitem_helper(i, **kwargs))
+            if len(indices) > 0: # This is for test_to_hierarchical_dataframe_empty_tables
+                try:
+                    data = self.target.get(slice(None),  **kwargs)
+                    slices = [self.__get_slice(i) for i in indices]
+                    if isinstance(data, pd.DataFrame):
+                        ret = [data.iloc[s] for s in slices]
+                    else:
+                        ret = [data[s] for s in slices]
+                except IndexError:
```
mavaylon1, line 212: Review Notes: This Index error capture essentially preserves the old method of doing it for that specific edge case. I want to get this rolled out for Alessio since this isn't a bug fix, but an optimization problem. If given the green light on this approach, I will create a ticket for this TODO.

rly: I don't understand the IndexError capture here - when does it occur?

rly:
This looks good to me, but see comment above

src/hdmf/common/table.py
mavaylon1: Sure thing. The index error occurs in `test_to_hierarchical_dataframe_empty_tables`. This test covers a weird edge case of handling a hierarchy of empty DynamicTable objects, specifically that to_dataframe can handle empty tables and preserve said hierarchy. I have no clue why this is even an edge case, but it is being tested.

oruebel: In the NWB schema this is relevant for the ICEphys metadata schema. Empty tables may occur if a user creates the tables but does not fill in any metadata.

rly: OK, I think it is fine to punt cleaning this code up to a separate PR, but I think this should be a high priority to resolve since it is fresh on our minds and I don't think it would take too much more effort to trace through and find a clean solution for. These edge cases are a pain to maintain if we leave it in.

rly (approved):

PR #1258, opened 2025-04-07
Update warning about ignoring cached namespaces

### LINKED ISSUE #1018: [Feature]: Remove or enchance the warning about ignoring cached name spaces

I get this warning to standard error sometimes and I don't know what to make out of it:

```
/home/heberto/miniconda3/envs/neuroconv_env/lib/python3.10/site-packages/hdmf/spec/namespace.py:531: UserWarning: Ignoring cached namespace 'hdmf-common' version 1.5.0 because version 1.8.0 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
/home/heberto/miniconda3/envs/neuroconv_env/lib/python3.10/site-packages/hdmf/spec/namespace.py:531: UserWarning: Ignoring cached namespace 'core' version 2.4.0 because version 2.6.0-alpha is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
/home/heberto/miniconda3/envs/neuroconv_env/lib/python3.10/site-packages/hdmf/spec/namespace.py:531: UserWarning: Ignoring cached namespace 'hdmf-experimental' version 0.1.0 because version 0.5.0 is already loaded.
  warn("Ignoring cached namespace '%s' version %s because version %s is already loaded."
```

It comes from here:
https://github.com/hdmf-dev/hdmf/blob/41fabd4eea4dc1e2a2c55af5b81358d1de86f3b1/src/hdmf/spec/namespace.py#L527-L532

I have the feeling that warnings should be actionable by the person that receives them. I feel that displaying this to end users is obscure (as I assume most users are not familiar with the concept of namespaces in the context of hdmf) and non-actionable. It is not clear what (and if) there is something wrong and what should be done about it.

If this is something that the end users should be concerned about, maybe enchance the warning to add instructions of what should be done or try.

If this is something that end users should not be concerned about maybe we could eliminate or restrict the warning so only the people that should be concerned get it. 

Can this be done?

Yes.

stephprince:
## Motivation

Fix #1018.

Combine the warnings about ignoring cached namespaces into a single warning, and only warn for core namespaces if the cached version is newer than the loaded version. 

## How to test the behavior?
with an example file from the pynwb test suite:
```python
from pynwb import NWBHDF5IO

io = NWBHDF5IO("tests/back_compat/2.1.0_nwbfile_with_extension.nwb", "r")
io.read()
```

```
UserWarning: Ignoring the following cached namespace(s) because another version is already loaded:
hdmf-experimental - cached version: 0.2.0, loaded version: 0.5.0
Ignore this warning if these versions are compatible.
```

## Checklist

### DIFF
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -602,7 +602,7 @@ class NamespaceCatalog:
             if name in self.__core_namespace and is_newer_version(cached_version, loaded_version):
                 core_warnings.append(version_info)  # for core namespaces, warn if the cached version is newer
             elif name not in self.__core_namespace:
-                other_warnings.append(version_info)  # for all other namespaces, issue a warning for compatability
+                other_warnings.append(version_info)  # for all other namespaces, issue a warning for compatibility
 
         if core_warnings:
             warning_msg.append(f'{("\n").join(core_warnings)}\nPlease update to the latest package versions.')
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -230,8 +230,8 @@ class NamespaceCatalog:
              'doc': 'the class to use for dataset specifications', 'default': DatasetSpec},
             {'name': 'spec_namespace_cls', 'type': type,
              'doc': 'the class to use for specification namespaces', 'default': SpecNamespace},
-            {'name': 'core_namespace', 'type': str,
-             'doc': 'the name of the core namespace', 'default': None})
+            {'name': 'core_namespaces', 'type': list,
+             'doc': 'the names of the core namespaces', 'default': list()})
     def __init__(self, **kwargs):
         """Create a catalog for storing  multiple Namespaces"""
         self.__namespaces = OrderedDict()
@@ -239,17 +239,8 @@ class NamespaceCatalog:
         self.__group_spec_cls = getargs('group_spec_cls', kwargs)
         self.__spec_namespace_cls = getargs('spec_namespace_cls', kwargs)
 
-        core_namespace = getargs('core_namespace', kwargs)
-        if core_namespace is None:
-            from ..common import CORE_NAMESPACE, EXP_NAMESPACE
-            # NOTE: even though HDMF does not guarantee backwards compatibility with schema
-            # using an older version of the experimental namespace, in practice, this has not been
-            # an issue, and it is costly to determine whether there is an incompatibility before issuing
-            # a warning. so, we ignore the experimental namespace warning by default.
-            # see https://github.com/hdmf-dev/hdmf/pull/1258
-            self.__core_namespace = [CORE_NAMESPACE, EXP_NAMESPACE]
-        else:
-            self.__core_namespace = [core_namespace]
+        core_namespaces = getargs('core_namespaces', kwargs)
+        self.__core_namespaces = core_namespaces
 
         # keep track of all spec objects ever loaded, so we don't have
         # multiple object instances of a spec
@@ -263,7 +254,7 @@ class NamespaceCatalog:
         ret = NamespaceCatalog(self.__group_spec_cls,
                                self.__dataset_spec_cls,
                                self.__spec_namespace_cls)
-        ret.__core_namespace = copy(self.__core_namespace)
+        ret.__core_namespaces = copy(self.__core_namespaces)
         ret.__namespaces = copy(self.__namespaces)
         ret.__loaded_specs = copy(self.__loaded_specs)
         ret.__included_specs = copy(self.__included_specs)
@@ -274,7 +265,7 @@ class NamespaceCatalog:
         for name, namespace in ns_catalog.__namespaces.items():
             self.add_namespace(name, namespace)
 
-        self.__core_namespace.extend(ns_catalog.__core_namespace)
+        self.__core_namespaces.extend(ns_catalog.__core_namespaces)
 
     @property
     @docval(returns='a tuple of the available namespaces', rtype=tuple)
@@ -298,9 +289,9 @@ class NamespaceCatalog:
         return self.__spec_namespace_cls
 
     @property
-    def core_namespace(self):
-        """The core namespace used in this NamespaceCatalog"""
-        return self.__core_namespace
+    def core_namespaces(self):
+        """The core namespaces used in this NamespaceCatalog"""
+        return self.__core_namespaces
 
     @docval({'name': 'name', 'type': str, 'doc': 'the name of this namespace'},
             {'name': 'namespace', 'type': SpecNamespace, 'doc': 'the SpecNamespace object'})
@@ -604,9 +595,9 @@ class NamespaceCatalog:
         warning_msg = list()
         for name, cached_version, loaded_version in ignored_namespaces:
             version_info = f"{name} - cached version: {cached_version}, loaded version: {loaded_version}"
-            if name in self.__core_namespace and is_newer_version(cached_version, loaded_version):
+            if name in self.__core_namespaces and is_newer_version(cached_version, loaded_version):
                 core_warnings.append(version_info)  # for core namespaces, warn if the cached version is newer
-            elif name not in self.__core_namespace:
+            elif name not in self.__core_namespaces:
                 other_warnings.append(version_info)  # for all other namespaces, issue a warning for compatibility
 
         if core_warnings:
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -9,7 +9,7 @@ from warnings import warn
 
 from .catalog import SpecCatalog
 from .spec import DatasetSpec, GroupSpec
-from ..utils import docval, getargs, popargs, get_docval
+from ..utils import docval, getargs, popargs, get_docval, is_newer_version
 
 _namespace_args = [
     {'name': 'doc', 'type': str, 'doc': 'a description about what this namespace represents'},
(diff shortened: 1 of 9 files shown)


src/hdmf/spec/namespace.py
```diff
@@ -508,36 +526,117 @@ def __register_dependent_types_helper(spec, inc_ns, catalog, registered_types):
...
+            version_info = f"{name} - cached version: {cached_version}, loaded version: {loaded_version}"
+            if name in self.__core_namespace and is_newer_version(cached_version, loaded_version):
+                core_warnings.append(version_info)  # for core namespaces, warn if the cached version is newer
+            elif name not in self.__core_namespace:
+                other_warnings.append(version_info)  # for all other namespaces, issue a warning for compatability
+
+        if core_warnings:
+            warning_msg.append(f'{("\n").join(core_warnings)}\nPlease update to the latest package versions.')
+        if other_warnings:
+            warning_msg.append(f'{("\n").join(other_warnings)}\nIgnore this warning if these versions are compatible.')
```
stephprince: Are there any more actionable steps that we can recommend users to take here if they are not compatible?

src/hdmf/utils.py
```diff
@@ -876,6 +876,16 @@ def is_ragged(data):
...
 
+def is_newer_version(version_a, version_b):
+    # this method could be replaced by packaging.version if packaging is added as a dependency
+    for a, b in zip(version_a, version_b):
+        if a > b:
+            return True
+        elif a < b:
+            return False
+
+    return False
```
stephprince: the packaging library has more robust version comparison functions, but I wasn't sure if we wanted to add it as a dependency just for this case.

rly: I don't think we need to just for this case. We decided against it in pynwb and previously in hdmf as well.

rly:
I would really like to not show the hdmf-experimental warning but I know we do not ensure backwards compatibility between schema versions there. What if we also hide the warning if none of the data types in the namespace are used in the file? Most files do not involve data types used in hdmf-experimental.

On rewrite/append, the loaded namespaces are written to the file, so we should not have a problem with writing new data that is incompatible with old namespaces.

stephprince:
> I would really like to not show the hdmf-experimental warning but I know we do not ensure backwards compatibility between schema versions there. What if we also hide the warning if none of the data types in the namespace are used in the file? Most files do not involve data types used in hdmf-experimental.

I think that's a good idea! I agree it would be nice to not display that warning unless necessary. 

Are there any places you can point me to in the code that get all the data types used in the file? I am thinking this would be an optional argument to `NamespaceCatalog.load_namespaces` since the NamespaceCatalog seems isolated from the file object?

rly:
After talking with @oruebel , we realized that it would be hard or hacky to get all the data types used in the file. The hacky way is to iterate through all groups and datasets in the file and look for the `neurodata_type` or `data_type` attributes, but this involves iterating through the whole file just to check for a warning. This iteration is expensive, especially for streaming data.

Another approach we discussed is to have the IO object build all the builders before loading the namespace from the file. Then we can iterate through the in-memory builders. But this requires refactoring the logic in NWBHDF5IO and HDF5IO, which currently takes in a BuildManager that has the loaded namespace. This would be complicated.

The primary use case for the hdmf-experimental warning is to warn users of incompatible/breaking changes introduced in, say 0.4.0 when the data was written in an older version, say 0.3.0. This really only impacts `ExternalResources`. (`EnumData` is also in hdmf-experimental but has not changed). To my knowledge, no one is using older versions of `ExternalResources` (and perhaps also the latest version), it is relatively stable now, and no one has yet raised an issue about incompatibilities. Therefore the impact of just ignoring this warning is minimal.

So, I would say, let's include hdmf-experimental into the "core" set of namespaces to ignore from hdmf.

stephprince:
> So, I would say, let's include hdmf-experimental into the "core" set of namespaces to ignore from hdmf.

Thanks for explaining in detail. That sounds good to me - I'll add the hdmf-experimental namespace to the `NamespaceCatalog.__core_namespace` list. 

Note that the pynwb [`__init__.py`](https://github.com/NeurodataWithoutBorders/pynwb/blob/1b613aa0b1c42be3eebc2c7bab87c157df081512/src/pynwb/__init__.py#L90) file will also need to be updated to add the "core" core namespace and fully eliminate the pynwb/hdmf namespace related warnings when the cached namespace version is not newer. (In the example output I demonstrated above with "tests/back_compat/2.1.0_nwbfile_with_extension.nwb", I think I was accidentally using a cached typemap in which I had added the "core" namespace in pynwb for testing purposes). 

Edit: see https://github.com/NeurodataWithoutBorders/pynwb/issues/2064 as the open issue to add "core" as a core namespace

stephprince:
@rly this is ready for you to look at again

src/hdmf/spec/namespace.py
```diff
@@ -229,13 +229,23 @@ class NamespaceCatalog:
...
         """Create a catalog for storing  multiple Namespaces"""
         self.__namespaces = OrderedDict()
         self.__dataset_spec_cls = getargs('dataset_spec_cls', kwargs)
         self.__group_spec_cls = getargs('group_spec_cls', kwargs)
         self.__spec_namespace_cls = getargs('spec_namespace_cls', kwargs)
+
+        core_namespace = getargs('core_namespace', kwargs)
+        if core_namespace is None:
+            from ..common import CORE_NAMESPACE, EXP_NAMESPACE
+            self.__core_namespace = [CORE_NAMESPACE, EXP_NAMESPACE]
```
rly:
```suggestion
            # NOTE: even though HDMF does not guarantee backwards compatibility with schema
            # using an older version of the experimental namespace, in practice, this has not been
            # an issue, and it is costly to determine whether there is an incompatibility before issuing
            # a warning. so, we ignore the experimental namespace warning by default. 
            # see https://github.com/hdmf-dev/hdmf/pull/1258
            self.__core_namespace = [CORE_NAMESPACE, EXP_NAMESPACE]
```

src/hdmf/utils.py
```diff
@@ -876,6 +876,16 @@ def is_ragged(data):
 
     return False
 
+def is_newer_version(version_a, version_b):
```
rly: It looks like the inputs in the code are strings, and then this code splits the strings into lists of characters and compares them element by element, so:
```
>>> is_newer_version("2.0.0", "10.0.0")
True
```
which is incorrect. You can use part of this code to get the version string as a list of ints: https://github.com/NeurodataWithoutBorders/pynwb/blob/1b613aa0b1c42be3eebc2c7bab87c157df081512/src/pynwb/io/utils.py#L7
and then this code should run fine.

src/hdmf/spec/namespace.py
```diff
@@ -229,13 +229,28 @@ class NamespaceCatalog:
...
+
+        core_namespace = getargs('core_namespace', kwargs)
+        if core_namespace is None:
+            from ..common import CORE_NAMESPACE, EXP_NAMESPACE
+            # NOTE: even though HDMF does not guarantee backwards compatibility with schema
+            # using an older version of the experimental namespace, in practice, this has not been
+            # an issue, and it is costly to determine whether there is an incompatibility before issuing
+            # a warning. so, we ignore the experimental namespace warning by default.
+            # see https://github.com/hdmf-dev/hdmf/pull/1258
+            self.__core_namespace = [CORE_NAMESPACE, EXP_NAMESPACE]
```
rly: I think this should be handled in https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/common/__init__.py#L236

This class should not know anything about hdmf.common.

To make this work, I think the `core_namespace` string argument should change to `core_namespaces` as a list of strings.

tests/unit/spec_tests/test_load_namespace.py
```diff
@@ -278,11 +286,55 @@ def test_catch_dup_name_same_version(self):
...
+        ns_builder1 = NamespaceBuilder('Extension doc', self.core_ns, version=new_ns_version)
+        ns_builder1.add_spec(self.ext_source1, GroupSpec('doc', data_type_def='MyType'))
+        ns_builder1.export(self.ns_path1, outdir=self.tempdir)
+
+        # create new catalog and merge the loaded core namespace catalog
+        ns_catalog = NamespaceCatalog()
+        ns_catalog.merge(self.ns_catalog)
+
+        # test no warning if loading older namespace than one already loaded
+        ns_catalog.load_namespaces(os.path.join(self.tempdir, self.ns_path1))
```
rly: Although we try to look at all warnings returned by tests, in CI, we do not always. I think it would be better to follow the pattern from one of the tests above to test that the namespace warning does not appear.

stephprince: That makes sense, I wasn't sure the best approach on this one. It also felt risky that when I changed the warning message, the test checking for the exact warning string would continue to pass even when the case was raising a warning... but I guess either approach has the potential to miss some things.

src/hdmf/utils.py
stephprince: thanks for catching that!

stephprince:
@rly I think I addressed all of your comments

I believe the test failures are due to an ImportError issue with the schemasheets package. The warning in the logs is a little confusing because schemasheets is actually installed, but does not seem to be compatible with the latest versions of `linkml_runtime>=1.9.0`. I'll open up an issue on that repo (edit: see https://github.com/linkml/schemasheets/issues/147).

src/hdmf/spec/namespace.py
rly: For the core warnings, your instructions sound good.

For the other warnings, I would say "The loaded extension(s) may not be compatible with the cached extension(s) in the file. Please check the extension documentation and ignore this warning if these versions are compatible."
What do you think?

In the future, when we have the ability to unload namespaces/extensions, we can update this warning.

tests/unit/spec_tests/test_load_namespace.py
rly: Good point. I suggest we re-raise any caught warnings in the for loop `for w in ws`. Could you update that in all places we use that pattern?

rly:
Thanks! I made a couple minor comments. Otherwise looks good.

src/hdmf/spec/namespace.py
stephprince: Those instructions sound good to me! And yes we can provide instructions to unload namespaces when we add those capabilities.

stephprince:
@rly I think I addressed all your remaining comments! This should be ready to merge

rly (approved):

rly:
Looks good. Thank you! Squash and merge anytime.

rly:
I'll just do it since Matt and I are managing this package and the release.

h-mayorquin:
This is great! Thanks for moving this forward!

PR #1261, opened 2025-04-29
Adjust append for appending to Array with multi-dim arg

mavaylon1:
## Motivation

In pynwb there is a case for appending to an array that is neither a scalar or a row vector. The arg matches the dim of the data and need essentially a vstack.

## Checklist

### DIFF
--- a/tests/unit/utils_test/test_data_utils.py
+++ b/tests/unit/utils_test/test_data_utils.py
@@ -22,13 +22,18 @@ class TestAppendData(TestCase):
         with self.assertRaises(ValueError):
             append_data(data, 4)
 
+    def test_append_same_dim_array_arg(self):
+        data = np.array([[1, 2], [3, 4]])
+        arg = np.array([[5, 6]])
+        new = append_data(data, arg)
+        assert_array_equal(new, np.array([[1, 2],
+                                       [3, 4],
+                                       [5, 6]]))
 
 class TestZarrAppendData(TestCase):
-
     def setUp(self):
         if not ZARR_INSTALLED:
             self.skipTest("optional Zarr package is not installed")
-
 
     def test_append_data_zarr(self):
         zarr_array = zarr.array([1,2,3])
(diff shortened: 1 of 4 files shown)


mavaylon1:
Review Notes:
All the failing tests are due to schemasheets. [TBD for link to issue ticket]

tests/unit/utils_test/test_data_utils.py
```diff
@@ -22,14 +22,19 @@ def test_append_exception(self):
         with self.assertRaises(ValueError):
             append_data(data, 4)
 
+    def test_append_same_dim_array_arg(self):
+        data = np.array([[1, 2], [3, 4]])
+        arg = np.array([[5, 6]])
```
rly, line 35: `append_data` should be used to append a single element to an existing array. So, arg should be just `np.array([5, 6])`. np.append does not work in this case. Either you have to add a dimension to the arg, or just use np.vstack. I think vstack will be easier to understand.

rly: Ah, I see that the `elif np.ndim(arg) < np.ndim(data)` will catch this case, because arg will be 1D and data is 2D. It should always be the case that the arg has one fewer dimension than the data. So when would the `else` statement be hit?

tests/unit/utils_test/test_data_utils.py
```diff
@@ -22,14 +22,19 @@ def test_append_exception(self):
...
+        new = append_data(data, arg)
+        assert_array_equal(new, np.array([[1, 2],
+                                       [3, 4],
+                                       [5, 6]]))
```
rly, line 39:
```suggestion
    def test_append_1D_to_2D(self):
        data = np.array([[1, 2], [3, 4]])
        arg = np.array([5, 6])
        new = append_data(data, arg)
        assert_array_equal(new, np.array([[1, 2],
                                       [3, 4],
                                       [5, 6]]))

    def test_append_same_dim_array_arg(self):
        data = np.array([[1, 2], [3, 4]])
        arg = np.array([[5, 6]])
        new = append_data(data, arg)
        assert_array_equal(new, np.array([[1, 2],
                                       [3, 4],
                                       [5, 6]]))
```

I recommend we also add this case to the test suite.

rly (approved):

rly:
LGTM!

PR #1273, opened 2025-05-08
Pydantic Explore

mavaylon1:
This marks the initial investigation of pydantic integration as a full replacement to docval.

1. **Using existing Pydantic Tools**
- The idea is to convert classes to being a pydantic `model`. This will validate the arguments used to instantiate the class.
- Methods within the classes that have docval will replace docval with the pydantic `@validate_call` decorator. This will give type validation and run custom validators. The latter is needed to ensure the functionality of docval is covered completely. 

2. **Qwirks**
- In this PR, you will find the initial conversion of a class to the pydantic model. Pydantic takes private class variables and hides them away from the public schema. This changes the way we are able to access these variables. We currently have methods to "get" them, but the issue is the value returned is not the value of the variable, but an instance of `ModelPrivateAttr`. This is wrapped around the variable, making it difficult to fully access the variable. 

3. **Needed Features**
-  Pydantic's native support for NumPy arrays is limited, especially regarding validation of array shape and data type. The numpydantic library addresses this limitation by providing robust tools for specifying and validating NumPy arrays within Pydantic models. https://github.com/p2p-ld/numpydantic 
- There is a use case of enum validation in pynwb that appears once. This can be done in pydantic, but it might be easier to just do something local in that use case.

### DIFF
--- a/src/hdmf/spec/catalog.py
+++ b/src/hdmf/spec/catalog.py
@@ -1,5 +1,6 @@
 import copy
 from collections import OrderedDict
+from pydantic import validate_call
 
 from .spec import BaseStorageSpec, GroupSpec
 from ..utils import docval, getargs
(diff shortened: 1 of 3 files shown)


rly:
This is superseded by #1331.

PR #1276, opened 2025-05-15
Raise Error when using a compound dtype that is not formally defined in the spec or extension.

### LINKED ISSUE #1265: [Bug]: When data type is unspecified, prevent compound dtypes
### What happened?

Follow-up to https://github.com/hdmf-dev/hdmf-zarr/issues/273

As @oruebel wrote: 
> While TimeSeries does not restrict the data type, it assume the use of a basic numeric data type (int, float) or string. Using a compound data type with a TimeSeries without defining an extension is currently not supported.

Our current docval-based methods for validation does not validate the dtype of arrays, and so we cannot validate on instantiation of the TimeSeries. This year, we plan to replace docval, so I suggest we punt the ability to prevent a TimeSeries with a structured array as the data from being instantiated to that time.

As mentioned in the linked issue, the HDF5 backend raises an unclear error on write, and the Zarr backend does not raise an error at all.

An error should be raised before the backends are involved, probably during the convert_dtype step of the build process, where the user data is converted to the dtype specified in the spec. If no dtype is specified in the spec, then we should check that the data is of a basic numeric data type or string. This behavior should also be documented in the spec language.

### Steps to Reproduce

```python
See linked issue
```

### Traceback

```pytb

```

### Operating System

Windows

### Python Version

3.13

### Package Versions

_No response_

mavaylon1:
Using a compound data type without defining an extension is currently not supported. This is to add a check and test to prevent this.

Fix #1265 
Original Issue: https://github.com/hdmf-dev/hdmf-zarr/issues/273

## Checklist


tests/unit/build_tests/mapper_tests/test_build.py
```diff
@@ -35,6 +35,25 @@ def test_unwrap(self):
...
+
+class TestBuildNoDtypeSpec(TestCase):
+    def test_structured_array_without_dtype_raises(self):
+        # Create a structured (compound) array
+        compound_data = np.array(
+            [(1.0, True), (2.0, False)],
+            dtype=[('volume', 'f4'), ('autorewarded', 'bool')]
+        )
+
+        # Set up build manager with core NWB type map
```
rly:
```suggestion
        # Set up build manager with core type map
```

rly (approved):

rly:
Looks good, thanks!

Issue #1123, opened 2024-06-03
[Feature]: Add `load_namespaces` and `get_namespaces` to `HDMFIO`


rly:
### What would you like to see added to HDMF?

It's useful in validation to load or get namespaces from an IO object but right now those are implemented per-backend. HDF5IO supports `load_namespaces` and `get_namespaces` and `ZarrIO` supports only `load_namespaces`. It would be useful if they and any future `HDMFIO` class supported both of these methods.

cc @stephprince 

### What solution would you like?

Add `load_namespaces` and `get_namespaces` as abstract methods to `HDMFIO`

### Do you have any interest in helping implement the feature?

Yes.

mavaylon1:
Interesting. Do you have a timeline in mind?

rly:
> Interesting. Do you have a timeline in mind?

This will require some coordination with hdmf-zarr. Let's target July.

rly:
Notes from discussion with @oruebel and @stephprince today:

AWS Region is common to both hdf5 and Zarr.

```python
@classmethod
@abstractmethod
def load_namespaces_from_path(namespace_catalog, path, aws_region=None):
    pass

In HDF5IO:
def load_namespaces_from_path(namespace_catalog, path, aws_region=None):
    h5py.File(path, driver=config.driver, aws_region=aws_region)

In ZarrIO:
def load_namespaces_from_path(namespace_catalog, path, aws_region=None):
    storage_options = dict()
    if aws_region:
        storage_options["region"] = aws_region
    zarr.open(path, storage_options=storage_options)

# instance method
@abstractmethod
def load_namespaces(namespace_catalog):  # act on this IO object. note that the file arg is not necessary
    pass
```

If custom IO options are required, the user must create the corresponding `HDMFIO` object with those options and call `io.load_namespaces(namespace_catalog)`.

Else, the user doesn't have to create the IO object first and can just call `io_cls.load_namespaces(namespace_catalog, path, aws_region)`

PR #1299, opened 2025-08-11
Move load_namespaces to HDMFIO and add load_namespaces_io

rly:
## Motivation

Implement solution for https://github.com/NeurodataWithoutBorders/pynwb/issues/2119.

Added new abstract methods to `HDMFIO`
- class method `load_namespaces` so that pynwb/other software can rely on `HDMFIO.load_namespaces` regardless of backend
- instance method `load_namespaces_io` which basically calls `load_namespaces` on the open file object

I had initially implemented `load_namespaces_io` with a new keyword argument set in `load_namespaces`, but the error checking was tedious (half of the other arguments in HDF5IO should not be supplied at the same time as an `io` argument), and it is easier to track down different ways that a method is used in an IDE if there are separate methods for different argument sets. The downside is that users have to remember there are two different methods, but these methods are not designed to be used by end-users, just other APIs, such as pynwb.

Since adding new abstract methods imposes a new requirement on downstream software - mostly `ZarrIO`, this is a breaking change. This PR targets the release/5.0.0 branch. We should add the implementing method in hdmf-zarr first and make sure PyNWB is compatible first to make the transition smooth.

- Also refactored tests that involve a second HDMFIO object to use the same helper/utility class, renamed from `OtherIO` to `DoNothingIO`.

## How to test the behavior?
```
```

## Checklist


rly:
Fixes part of #1123. ~TODO: Also implement `get_namespaces`~

rly:
@stephprince this is now ready for review

stephprince (approved):

PR #1302, opened 2025-08-12
Fix copying of typemap and refactor TypeConfigurator

### LINKED ISSUE #1301: [Bug]: deepcopy of TypeMap uses same TypeConfig
### What happened?

https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/common/__init__.py#L176

`get_type_map` in `hdmf/common/__init__.py` with no keyword arguments returns a deep-copy of the global hdmf-common type map. This creates a copy of all the objects _except_ the TypeConfigurator(), for which the same object is used in each deep-copy. Instead, each type config should be different, and handling of global type map / type config should be handled at a higher level, i.e., in hdmf/common/__init__.py

### Steps to Reproduce

```python
from copy import deepcopy
from hdmf.build import TypeMap

type_map = TypeMap()
print("Original Type Config", type_map.type_config)

c1 = deepcopy(type_map)
print("Copy 1 Type Config", c1.type_config)
c2 = deepcopy(type_map)
print("Copy 2 Type Config", c2.type_config)
c3 = deepcopy(type_map)
print("Copy 3 Type Config", c3.type_config)
c4 = deepcopy(type_map)
print("Copy 4 Type Config", c4.type_config)
```

### Traceback

```pytb
Original Type Config <hdmf.term_set.TypeConfigurator object at 0x10624f320>
Copy 1 Type Config <hdmf.term_set.TypeConfigurator object at 0x10624f320>
Copy 2 Type Config <hdmf.term_set.TypeConfigurator object at 0x10624f320>
Copy 3 Type Config <hdmf.term_set.TypeConfigurator object at 0x10624f320>
Copy 4 Type Config <hdmf.term_set.TypeConfigurator object at 0x10624f320>
```

### Operating System

macOS

### Python Version

3.13

### Package Versions

_No response_

rly:
## Motivation

Fix #1301, work toward fixing https://github.com/NeurodataWithoutBorders/pynwb/issues/1958

Also deprecates calling `get_type_map` with any 'extensions' arguments, a use case I have never seen. IMO it's easy for the rare user to just call `load_namespaces` on the type map than us maintain all this extra convenience logic that even we don't use. 

## Checklist

### DIFF
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -431,7 +431,7 @@ class TypeMap:
         return self.__container_types
 
     def __copy__(self):
-        ret = TypeMap(copy(self.__ns_catalog), self.__default_mapper_cls, self.type_config)
+        ret = TypeMap(copy(self.__ns_catalog), self.__default_mapper_cls, TypeConfigurator(self.type_config.paths))
         ret.merge(self)
         return ret
 
@@ -464,6 +464,7 @@ class TypeMap:
         for custom_generators in reversed(type_map.__class_generator_manager.custom_generators):
             # iterate in reverse order because generators are stored internally as a stack
             self.register_generator(custom_generators)
+        # NOTE: the type config is not merged from the input type map to the new one. add if there is a clear use case
 
     @docval({"name": "generator", "type": type, "doc": "the CustomClassGenerator class to register"})
     def register_generator(self, **kwargs):
--- a/src/hdmf/common/__init__.py
+++ b/src/hdmf/common/__init__.py
@@ -4,6 +4,7 @@ for reading and writing data in according to the HDMF-common specification
 import os.path
 from copy import deepcopy
 from collections.abc import Callable
+import warnings
 
 CORE_NAMESPACE = 'hdmf-common'
 EXP_NAMESPACE = 'hdmf-experimental'
@@ -26,11 +27,10 @@ global __TYPE_MAP
         is_method=False)
 def load_type_config(**kwargs):
     """
-    This method will either load the default config or the config provided by the path.
-    NOTE: This config is global and shared across all type maps.
+    This method will either load the config at the given path into either the global type map or a specific type map.
     """
     config_path = kwargs['config_path']
-    type_map = kwargs['type_map'] or get_type_map()
+    type_map = kwargs['type_map'] or __TYPE_MAP
 
     type_map.type_config.load_type_config(config_path)
 
@@ -38,23 +38,23 @@ def load_type_config(**kwargs):
         is_method=False)
 def get_loaded_type_config(**kwargs):
     """
-    This method returns the entire config file.
+    This method returns a dictionary with the configuration for each namespace and data type.
     """
-    type_map = kwargs['type_map'] or get_type_map()
+    type_map = kwargs['type_map'] or __TYPE_MAP
 
     if type_map.type_config.config is None:
         msg = "No configuration is loaded."
         raise ValueError(msg)
-    else:
-        return type_map.type_config.config
+
+    return type_map.type_config.config
 
 @docval({'name': 'type_map', 'type': TypeMap, 'doc': 'The TypeMap.', 'default': None},
         is_method=False)
 def unload_type_config(**kwargs):
     """
-    Unload the configuration file.
+    Unload all type configurations from the global type map or a specific type map.
     """
-    type_map = kwargs['type_map'] or get_type_map()
+    type_map = kwargs['type_map'] or __TYPE_MAP
 
     return type_map.type_config.unload_type_config()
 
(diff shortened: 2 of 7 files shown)


src/hdmf/container.py
```diff
@@ -132,39 +135,39 @@ def _field_config(self, arg_name, val, type_map):
...
+        config_data_type = config_namespace['data_types'][data_type]
+        try:
+            config_termset_path = config_data_type[mapped_attr_name]
+        except KeyError:
+            return val
+
+        termset_path = os.path.join(CUR_DIR, config_termset_path['termset'])
+        termset = TermSet(term_schema_path=termset_path)
+        val = TermSetWrapper(value=val, termset=termset)
+        return val
```
rly, line 167: The changes here
1. refactor the logic so that we have fewer nested if/else statements. Return breaks the if so we can dedent
2. Updates `configurator.path` to `configurator.paths`

src/hdmf/term_set.py
```diff
@@ -355,14 +355,13 @@ class TypeConfigurator:
...
         self.config = None
-        if kwargs['path'] is None:
-            self.path = []
-        else:
-            self.path = [kwargs['path']]
-            self.load_type_config(config_path=self.path[0])
+        self.paths = []
+        if kwargs["paths"]:
+            for p in kwargs["paths"]:
+                self.load_type_config(p)
```
rly, line 364: It was confusing for `path` to be a list of paths, so change `path` to `paths`. It was also confusing for init to take only one path when there may be multiple paths you want to load, and so subsequent calls require calling `load_type_config`.

src/hdmf/common/__init__.py
```diff
@@ -26,35 +27,34 @@
         is_method=False)
 def load_type_config(**kwargs):
     """
-    This method will either load the default config or the config provided by the path.
-    NOTE: This config is global and shared across all type maps.
+    This method will either load the config at the given path into either the global type map or a specific type map.
     """
     config_path = kwargs['config_path']
-    type_map = kwargs['type_map'] or get_type_map()
+    type_map = kwargs['type_map'] or __TYPE_MAP
```
rly, line 33: We want to use the global type map here, not a copy of the global type map. This worked before because copying the global type map maintained a reference to the same type configurator, so the configurator was unintentionally global.

rly:
I added notes to aid review inline in the PR.

src/hdmf/container.py
```diff
@@ -114,57 +116,55 @@ def _field_config(self, arg_name, val, type_map):
...
-            msg = "%s not found within loaded configuration." % self.namespace
+            return val
+
+        # check to see that the container type is in the config under the namespace
+        config_namespace = termset_config['namespaces'][self.namespace]
+        data_type = self.data_type
+
+        # return the value if the data type for the container is not in the config
+        if data_type not in config_namespace['data_types']:
+            return val
```
rly, line 143: I removed the warning here because not all data types used will be configured

src/hdmf/container.py
```diff
@@ -114,57 +116,55 @@ def _field_config(self, arg_name, val, type_map):
...
         if isinstance(val, TermSetWrapper):
             msg = "Field value already wrapped with TermSetWrapper."
             warn(msg)
             return val
 
-        # check to see that the namespace for the container is in the config
+        # return the value if the namespace for the container is not in the config
         if self.namespace not in termset_config['namespaces']:
-            msg = "%s not found within loaded configuration." % self.namespace
+            return val
```
rly, line 135: I removed the warning here because not all namespaces used will be configured

rly:
It is expected that PyNWB tests fail. They only worked because the TypeConfigurator was the same instance across copies and PyNWB was using copies of TypeMap when loading configurations. This is fixed in https://github.com/NeurodataWithoutBorders/pynwb/pull/2121

oruebel (approved):
Looks good to me. @stephprince could you also take a look. The TypeMap stuff is pretty deep in the weeds and it would be good to get another pair of eyes on this just to make sure.

src/hdmf/build/manager.py
```diff
@@ -464,6 +464,7 @@ def merge(self, type_map, ns_catalog=False):
         for custom_generators in reversed(type_map.__class_generator_manager.custom_generators):
             # iterate in reverse order because generators are stored internally as a stack
             self.register_generator(custom_generators)
+        # NOTE: the type config is not merged from the input type map to the new one. add if there is a clear use case
```
stephprince, line 467: Does this have any implications for `_field_config` if the type configs are not merged when merging the hdmf/pynwb type maps? 

It's not clear to me when all the type configuration is happening so I might just be misunderstanding the workflow

rly: This would occur only when someone loads a type config into the HDMF type map before merging the hdmf/pynwb type maps, which I don't foresee happening. Users are either 1) using pynwb and loading type configs through pynwb after the hdmf/pynwb type maps have been merged, or 2) using hdmf only and no merging happens.

But if someone were to do the above, then the type config loaded into the HDMF type map would be ignored/lost when using the merged hdmf/pynwb type map.

rly:
This also resolves https://github.com/NeurodataWithoutBorders/pynwb/issues/1939

src/hdmf/build/manager.py
stephprince: That makes sense - thanks for explaining!

stephprince (approved):

PR #1309, opened 2025-09-15
Handle conflicts between core schema and extension

stephprince:
## Motivation

First step towards fixing https://github.com/NeurodataWithoutBorders/pynwb/issues/2108, which will also require remapping Device objects during read in pynwb.

Check for conflicts between the core and extension schemas. These changes check for pretty specific link/attribute conflicts from the `Device` / `DeviceModel` changes in NWB Schema 2.9.0. @rly and I discussed that ideally these checks would be more generalized in the future, but until then additional conflicts could be added to the `check_cross_type_conflicts` function.

todo: 
- [x] add tests

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -649,6 +649,25 @@ class NamespaceCatalog:
             self.__order_deps_aux(order, deps, subk)
         order.append(key)
 
+    def _get_namespace_for_type(self, type_name: str, catalog: SpecCatalog) -> str:
+        """
+        Get the namespace name that contains the given type by looking up its source file.
+        
+        Args:
+            type_name: The name of the type to find the namespace for
+            catalog: The catalog containing the type
+            
+        Returns:
+            The namespace name containing the type, or None if not found
+        """
+        # Get the source file for this type
+        source_file = catalog.get_spec_source_file(type_name)
+        for ns_name, sources in self.__included_sources.items():
+            if source_file in sources:
+                return ns_name
+                
+        return None
+
     def _check_namespace_conflicts(self, extension_ns_name: str, extension_ns_source: str, catalog: SpecCatalog):
         """
         Check for conflicts between extension namespaces and core namespace.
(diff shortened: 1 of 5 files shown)


src/hdmf/spec/namespace.py
```diff
@@ -641,3 +648,118 @@ def __order_deps_aux(self, order, deps, key):
         for subk in subdeps:
             self.__order_deps_aux(order, deps, subk)
         order.append(key)
+
+    def _get_namespace_for_type(self, type_name: str, catalog: SpecCatalog) -> str:
+        """
```
stephprince, line 653: @rly I added this function after we last met to handle finding the first parent dtype that is from a core namespace in the dtype hierarchy. Is there an existing function or better way to do this?

rly:
(note to self: after releasing this, check that zenodo sync worked)

src/hdmf/spec/namespace.py
```diff
@@ -641,3 +648,118 @@ def __order_deps_aux(self, order, deps, key):
...
+        if len(warning_msg) > 0:
+            warning_msg = "\n".join(warning_msg)
+            warn(f"Schema conflict(s) detected in namespace '{extension_ns_name}': \n {warning_msg} \n"
+                 f"This may cause compatibility issues. Please update the extension version if possible or "
+                 f"install an older version of the core schema that is compatible.",
+                 category=UserWarning, stacklevel=2)
+
+    def _check_cross_type_conflicts(self, extension_ns_name, spec_name, extension_spec, core_spec):
+        """
+        Check for type conflicts between extension and core specs (e.g., attribute vs link).
```
rly, line 728:
```suggestion
        Check for type conflicts between extension and core specs (e.g., attribute vs link).
        
        The extension spec data type should be a descendent of the core spec data type.
```

src/hdmf/spec/namespace.py
```diff
@@ -641,3 +648,118 @@ def __order_deps_aux(self, order, deps, key):
...
+                        warning_msg.append(conflict_msg)
+
+        if len(warning_msg) > 0:
+            warning_msg = "\n".join(warning_msg)
+            warn(f"Schema conflict(s) detected in namespace '{extension_ns_name}': \n {warning_msg} \n"
+                 f"This may cause compatibility issues. Please update the extension version if possible or "
+                 f"install an older version of the core schema that is compatible.",
+                 category=UserWarning, stacklevel=2)
+
+    def _check_cross_type_conflicts(self, extension_ns_name, spec_name, extension_spec, core_spec):
```
rly:
```suggestion
    def _check_cross_type_conflicts(self, extension_ns_name, extension_spec, core_spec):
```

You should be able to get the spec name from `extension_spec.data_type_def`

src/hdmf/spec/namespace.py
```diff
@@ -641,3 +648,118 @@ def __order_deps_aux(self, order, deps, key):
...
+
+        Note that detection of conflicts in which the same type_name is defined in both the extension and core
+        are detected by the SpecCatalog when registering the spec.
+        """
+
+        # get all the types in the catalog that are from the extension namespace
+        extension_types = [reg_type for reg_type in catalog.get_registered_types() if
+                           catalog.get_spec_source_file(reg_type) == extension_ns_source]
+
+        # check all types in the extension namespace for conflicts with core namespaces
```
rly:
```suggestion
        # check all types in the extension namespace for conflicts with types in core namespaces
        # that the extension types extend. for example, `ExcitationSource` is a `DeviceInstance` that
        # extends `Device` to have a "model" link. a later version of the core namespace defines 
        # `Device` to have a "model" link that conflicts with `DeviceInstance.model` in its target type.
        # Therefore both `DeviceInstance` and `ExcitationSource` have conflicts and will raise warnings
        # in the code below.
```

src/hdmf/spec/namespace.py
rly: Looks good. I defined a separate `get_namespace_for_type` that is namespace-catalog-wide in #1312. This assumes a type exists only once in the namespace catalog. Let's keep yours, merge this, and then we can think about refactoring in #1312.

rly:
I made a couple small suggestions. Otherwise looks good!

src/hdmf/spec/namespace.py
stephprince: updated to get the spec name from `extension_spec.data_type_def`

rly (approved):

rly:
Looks good. Thanks @stephprince !

PR #1312, opened 2025-09-18
Refactor inc spec resolution to pass array properties

rly:
## Motivation
Fix #320

The key changes are:

1. __New resolution system__: Instead of resolving includes during spec loading, resolution now happens after all specs are loaded via `NamespaceCatalog.resolve_all_specs()`
2. __New methods__: `BaseStorageSpec.resolve_inc_spec()` replaces the old `BaseStorageSpec.resolve_spec()` method
3. __Resolution tracking__: New properties `BaseStorageSpec.resolved` and `BaseStorageSpec.inc_spec_resolved` track resolution state
4. __Cross-namespace resolution__: The system can now resolve specs that include types from different namespaces
5. `dtype`, `shape`, `dims`, `value`, and `default_value` in `DatasetSpec` and `AttributeSpec` are now inherited and validated from the parent data type spec
6. If `dims` are not provided in a `DatasetSpec` or `AttributeSpec`, but `shape` is provided, `dims` will be set to a list of dummy dimension names, e.g., "dim_0", "dim_1", etc.

This PR also adds a few TODOs for extra checks for cases that should not be allowed but are currently allowed (even before this PR). These will be handled in future PRs from #1316.

Supersedes / closes #321.

## How to test the behavior?
```
pytest
```

## Checklist

- [x] Did you update `CHANGELOG.md` with your changes?
- [x] Does the PR clearly describe the problem and the solution?
- [x] Have you reviewed our [Contributing Guide](https://github.com/hdmf-dev/hdmf/blob/dev/docs/CONTRIBUTING.rst)?
- [x] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged?

PR #1313, opened 2025-09-19
Support get_class and roundtrip on datetime datasets

### LINKED ISSUE #429: get_class to handle datetimes
Handling a datetime attribute requires e.g.

```python
@ObjectMapper.constructor_arg('session_start_time')
    def dateconversion(self, builder, manager):
        datestr = builder.get('session_start_time').data
        date = dateutil_parse(datestr)
        return date
```

in the io mapper for that class. We would need to implement this for get_class to work on neurodata_types that take datetimes.

rly:
## Motivation

Fixes #429. Related to (now-merged) #1312.

Previously, neurodata types with `isodatetime`/`datetime` datasets could not be roundtripped through `get_class`-generated classes without manually wiring up an `@ObjectMapper.constructor_arg` converter to parse the stored ISO string back into a `datetime`. This PR makes that work end to end.

## What changed

- **Write path.** New `_isoformat` dtype helper converts `datetime`/`date` values to ASCII-encoded ISO bytes on write so they land as fixed-length text in HDF5.
- **Read path.** New `_parse_isoformat` parses stored values back to `datetime` (or `date` for date-only strings) for any dataset/attribute whose spec declares `dtype='isodatetime'` or `dtype='datetime'`, including arrays.
- **Inc-site dtype overrides.** `construct()` now accepts an optional `spec_ext` so a parent group that declares `dtype='isodatetime'` on a `data_type_inc`-extending dataset whose own def has no dtype is honored on read. `spec_ext` is plumbed through `classgenerator`, `manager`, and `objectmapper`.
- **Scalar datetime in generated classes.** The `scalar_data` docval macro now includes `datetime`/`date`, so `get_class`-generated dtype-less classes accept scalar datetime values without per-class shims.
- **Test scaffolding.** `_assert_data_equal` skips the length comparison for scalar `Data` (a 0-d datetime has no `len`), preventing false negatives in roundtrip assertions.

## How to test

The new tests cover the main paths:

- `tests/unit/build_tests/mapper_tests/test_build_datetime.py` — datetime/date build round-trips for both regular specs and extension specs (ported from #1314).
- `tests/unit/test_hdf5_roundtrip_extensions.py` — full HDF5 roundtrip including the inc-site dtype override case.
- `tests/unit/build_tests/test_classgenerator.py` — `get_class` with datetime fields, including scalar datetime acceptance via the widened `scalar_data` macro.

```bash
pytest tests/unit/build_tests/mapper_tests/test_build_datetime.py \
       tests/unit/test_hdf5_roundtrip_extensions.py \
       tests/unit/build_tests/test_classgenerator.py
```

## Checklist


rly:
Superseded by #1458

PR #1333, opened 2025-10-08
Deprecate unused, superfluous methods in spec module

rly:
The `hdmf.spec.spec` module contain a number of unused methods and convenience methods that as far as I know are used only in tests. They add unnecessary upkeep and can be deprecated and removed.
- `BaseStorageSpec.add_attribute` (`BaseStorageSpec.set_attribute` exists)
- `GroupSpec.add_group` (`set_group` exists)
- `GroupSpec.add_dataset` (`set_dataset` exists)
- `GroupSpec.add_link` (`set_link` exists)
- `BaseStorageSpec.get_data_type_spec` - not sure why this exists
- `BaseStorageSpec.get_namespace_spec` - same as above
 
## How to test the behavior?
```
pytest
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   - Cross-namespace resolution: The system can now resolve specs that include types from different namespaces
   - `dtype`, `shape`, `dims`, `value`, and `default_value` in `DatasetSpec` and `AttributeSpec` are now inherited and validated from the parent data type spec
 - If `dims` are not provided in a `DatasetSpec` or `AttributeSpec`, but `shape` is provided, `dims` will be set to a list of dummy dimension names, e.g., "dim_0", "dim_1", etc. @rly [#1312](https://github.com/hdmf-dev/hdmf/pull/1312)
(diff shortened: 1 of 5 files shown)


oruebel:
I think the docs are using the `add_dataset` etc. methods, so we should update the docs, e.g,.:

https://nwb-overview.readthedocs.io/en/latest/extensions_tutorial/extension_examples/labmetadata_extension.html

rly:
Thanks @oruebel . PyNWB has its own `add_group` and `add_dataset` that should also be deprecated. https://github.com/NeurodataWithoutBorders/pynwb/blob/6ff7e44567083fb16c14dd8f2d9e59c643f78ca8/src/pynwb/spec.py#L191

stephprince:
It looks like several catalystneuro extensions use the `add_dataset`, `add_group`, and `add_attribute` convenience methods. Maybe a good approach would be to deprecate the HDMF spec methods and leave the ones on the NWBGroupSpec since some are already redefined there anyway?

Edit: Some of those extensions look older, so I'm not sure if they are compatible with the latest versions of HDMF/PyNWB anyway. I'm fine with deprecating but just wanted to note where these methods might also be used.

rly:
Yeah, I saw those too. Those scripts are not run when extension users use the package, but rather only during development by a small group of knowledgeable people. Perhaps I am being too drastic, but to reduce maintenance burden and simplify the API, I favor deprecating here and in PyNWB.

Edit: I saw those after @oruebel mentioned them being used in extensions

stephprince (approved):
> Those scripts are not run when extension users use the package, but rather only during development by a small group of knowledgeable people.

Got it - reducing maintenance / simplifying sounds good!

PR #1349, opened 2025-12-01
Fix TypeMap.merge to preserve original namespace associations for DynamicTableRegion

### LINKED ISSUE #1347: `DynamicTableRegion` written with `hdmf_experimental` instead of `hdmf_common` when using NWB extensions
Hi all,

```traceback
Error using feval

Error in nwbRead (line 99)
Error in <script> (line 17)
```

Image

Image

bendichter:
fix #1347

Generated by Claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF (Unreleased)
+
+### Fixed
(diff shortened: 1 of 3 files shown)


rly:
This PR doesn't actually resolve the problem using the MWE in the original issue... The namespace is still "hdmf-experimental".

PR #1354, opened 2025-12-10
HTML repr: Remove double nesting in processing modules

h-mayorquin:
This PR modifies the html repr of the NWBFiles so the processing modules do not display a double nesting with the data_interfaces attribute:

image

Left is the new behavior right is the old behavior

This uses duck typing in the MultiContainerinterface to eliminate the double nesting (right on the image above) for the `ProcessingModule`. As `ProcessingModule` lives on pynwb I can't do the more natural solution of either overriding the method there (which would imply having some of the html repr code in pynwb) or checking of class instance here (which risks a circular dependency). 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 
 ### Changed
 - Changed how to call `BuildManager.build`, `TypeMap.build`, and `ObjectMapper.build` when exporting. The `export` argument is no longer accepted by `TypeMap.build` and `ObjectMapper.build`; `BuildManager.build` still accepts the `export` argument but now uses it to set an internal flag instead of passing it through the call chain. @rly [#1358](https://github.com/hdmf-dev/hdmf/pull/1358)
-- Changed HTML representation of `MultiContainerInterface` to flatten the `data_interfaces` field,
-  removing the redundant nesting level for `ProcessingModule` in PyNWB. @h-mayorquin [#1354](https://github.com/hdmf-dev/hdmf/pull/1354)
+- Changed HTML representation of `MultiContainerInterface` to flatten the grouping attribute when only
+  one grouping is defined (`len(__clsconf__) == 1`), removing the redundant nesting level since users
+  can access children directly via `container["name"]`. @h-mayorquin [#1354](https://github.com/hdmf-dev/hdmf/pull/1354)
 
 ### Fixed
 - Fixed bug when validating string datasets in NWB Zarr files. @stephprince [#1348](https://github.com/hdmf-dev/hdmf/pull/1348)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   If `copy=False`, the returned type map will be a direct reference to the global type map. @rly
   [#1352](https://github.com/hdmf-dev/hdmf/pull/1352)
 
+### Changed
+- Changed HTML representation of `MultiContainerInterface` to flatten the `data_interfaces` field,
+  removing the redundant nesting level for `ProcessingModule` in PyNWB. @h-mayorquin
```
```suggestion
  removing the redundant nesting level for `ProcessingModule` in PyNWB. @h-mayorquin [#1354](https://github.com/hdmf-dev/hdmf/pull/1354)
```

oruebel (commented):
Looks good to me

rly:
> As `ProcessingModule` lives on pynwb I can't do the more natural solution of either overriding the method there (which would imply having some of the html repr code in pynwb)

I understand that it's nice to have all the html code in one place, but I don't think it is such a bad thing to have HTML repr code in PyNWB. Each subclass should be able to control how it's displayed in HTML. We already have overrides for TimeSeries in PyNWB: https://github.com/NeurodataWithoutBorders/pynwb/blob/7e100c2d06ba73cd9ad873a7b1ee0a2d5156ce8a/src/pynwb/base.py#L352

In the approach of this PR, it might be confusing to maintainers who are wondering why there is a special case for PyNWB in HDMF, and this function could grow in complexity as more special cases get added. If someone builds an extension that happens to have a MultiContainerInterface with attribute name "data_interfaces", they may be confused as to why this behavior is different. These are rare cases though.

I would prefer to have this in PyNWB.

h-mayorquin:
> I would prefer to have this in PyNWB.

I am OK with it. 

The only downside is the slower cycle of release : (

h-mayorquin:
Moved to pynwb here:
https://github.com/NeurodataWithoutBorders/pynwb/pull/2149

Feel free to close this then.

rly:
Now that I have tested this and the other PR locally, it would be nice to make this work for all MultiContainerInterface classes that have only one grouping attribute. These classes allow the user to get the child by name without knowing the attribute name, e.g., `nwbfile.processing["behavior"]["Position"]["SpatialSeries"]` instead of `nwbfile.processing["behavior"].data_interfaces["Position"].spatial_series["SpatialSeries"]`. Could you update this so that if there is only one grouping attribute, the grouping attribute is removed from the html repr (e.g., "data_interfaces", "spatial_series")? The pynwb PR would not be necessary anymore, I think.

See https://github.com/hdmf-dev/hdmf/blob/35f123ebb1b8a3b9701db6540ebf76246948e8fd/src/hdmf/container.py#L1278-L1282

h-mayorquin:
Ok, I fixed this and the tooltip:

@rly 

image

The tooltip looks like this here because I was testing on the dev branch but is fixed in this branch.

rly:
Great! Thank you so much @h-mayorquin !

rly (approved):

PR #1355, opened 2025-12-10
Add type to container html repr

h-mayorquin:
As discussed with @rly @stephprince and @rly 

Take a look at the format. 

image

image

## Checklist


rly:
@h-mayorquin could you please resolve the conflicts? 

Why exclude top-level fields? I think it's useful, e.g., for the Units and Electrodes tables.

The html repr is really about how the file is represented in memory in Python, as opposed to on-disk (for example, the "general" group is not shown). In almost all cases, the objects in the repr correspond to PyNWB/HDMF classes, and those match the neurodata type. In rare cases, they might deviate, e.g., when a class with a different name from the data type is mapped to the data type. Could we just use the class name instead of the data type name?

h-mayorquin:
> Why exclude top-level fields? I think it's useful, e.g., for the Units and Electrodes tables.

I was looking at the trials and epochs that are also inside an intervals accessor at the top level and inferring wrongly that there all the top level things are just general containers (e.g. acquisition, processing, intervals, etc). But then, we have units and electrodes that are not like this so I guess let's just always display this.

h-mayorquin:
@rly I modified the code to match the suggestions.

src/hdmf/container.py
```diff
@@ -747,10 +749,17 @@ def _generate_field_html(self, key, value, level, access_code):
         else:
             html_content = f'<span class="field-key">{value}</span>'
 
+        display_name = str(key)
+        if isinstance(value, AbstractContainer):
+            class_name = type(value).__name__
+            is_name_different_from_class = str(key) != class_name
+            if is_name_different_from_class:
+                class_name_str = f" <span style='font-weight: normal; color: #888;'>({class_name})</span>"
+                display_name += class_name_str
```
rly: Why not show the class name if it matches the key? It might be confusing, but I think always showing the class name is nice in that it is predictable. Sorry about all the back and forth!

rly: Not trying to criticize - just trying to understand your thinking here!

h-mayorquin: Yes, I was thinking that having the same thing displaying twice would be confusing but on a second thought I think I prefer consistency.

h-mayorquin: Changed.

rly: Thank you!

rly (approved):

PR #1369, opened 2026-01-02
Add column descriptions to DynamicTable HTML representation

h-mayorquin:
Previously, there was no way to view column descriptions in the HTML display of DynamicTables. I think this will be useful to add. When working on conversion projects I use the HTML to check that everthing is as intended. In addition, this feature will be useful for people reading new NWBFiles through the notebok (e.g. from Dandi) as it improves the representation of common but important tables such as trials, electrodes, and units. This is especially important for custom-added columns, where most columns are hard to interpret without a description.

For the implementation, a new collapsible "columns" section is added to the `generate_html_repr()` method in `DynamicTable`. Each column is displayed as an individually collapsible item (collapsed by default) that shows the column's description when expanded. Column names include an access code tooltip (e.g., `['column_name']`) that can be clicked to copy to clipboard, consistent with other field displays in the HTML representation.

This is how it looks:

image

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   [#1352](https://github.com/hdmf-dev/hdmf/pull/1352)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


rly (approved):

rly:
Great idea. Thanks @h-mayorquin !

PR #1372, opened 2026-01-07
Refactor load_namespaces to return source types, fix registration

### LINKED ISSUE #1347: `DynamicTableRegion` written with `hdmf_experimental` instead of `hdmf_common` when using NWB extensions

rly:
- Larger breaking change and refactor of `load_namespaces` (from TypeMap, NamespaceCatalog, and the IO classes) to fix #1347 and improve the code structure for clarity and utility. This change will target HDMF 5.0 and needs to be rebased. 
- A separate PR will handle a smaller, non-breaking fix. 

TODO:
- Add tests
- Target 5.0 branch

## How to test the behavior?
```
Run tests and pynwb tests
```

## Checklist


src/hdmf/build/manager.py
```diff
@@ -379,25 +382,33 @@ def is_sub_data_type(self, **kwargs):
...
-    @property
-    def data_type(self):
-        return self.__data_type
+      - Calling get_dt_container_cls on the source namespace/type (which may auto-generate the class)
+      - Replacing the TypeSource with the actual class via register_container_type
+
+    - When searching for a type across namespaces, TypeSource entries are skipped to avoid returning the wrong
+      namespace.
+
+    This class should by used only by TypeMap.
```
oruebel:
```suggestion
    This class should be used only by TypeMap.
```

src/hdmf/build/manager.py
```diff
@@ -421,39 +432,46 @@ def __init__(self, **kwargs):
...
-        # XXX: From @nicain: All of a sudden legacy tests started
-        #      needing this argument in deepcopy. Doesn't hurt anything, though.
         return self.__copy__()
 
     def copy_mappers(self, type_map):
+        warnings.warn(
+            "copy_mappers is deprecated and will be removed in a future version. "
+            "Use merge instead with the argument ns_catalog=False to copy only mappers without namespaces.",
+            DeprecationWarning,
+        )
```
oruebel, line 469: Since `copy_mappers` is probably only used internally, I think it would be Ok to remove it with the upcoming major release. Up to you.

src/hdmf/build/manager.py
```diff
@@ -463,9 +481,9 @@ def copy_mappers(self, type_map):
     def merge(self, type_map, ns_catalog=False):
```
oruebel: Would be nice to add a docstring for `merge` since that seems to be the recommended approach

oruebel (dismissed):

src/hdmf/build/manager.py
rly: I thought so too, but it was used by pynwb, and it doesn't really hurt to do this a little gradually just in case.

oruebel (approved):

PR #1373, opened 2026-01-07
Fix container class mapped to incorrect namespace

### LINKED ISSUE #1347: `DynamicTableRegion` written with `hdmf_experimental` instead of `hdmf_common` when using NWB extensions

rly:
## Motivation

- Fix #1347 with a minimal change. Although this resolves the issue, the logic within load_namespaces and TypeMap regarding container class registration is not intuitive, and could be improved with a larger refactor as shown in #1372.

## How to test the behavior?
```
Run tests
```

## Checklist

### DIFF
--- a/src/hdmf/build/manager.py
+++ b/src/hdmf/build/manager.py
@@ -521,7 +521,7 @@ class TypeMap:
             for ns_key, ns_data_types in self.__container_types.items():
                 # NOTE that the type_name may appear in multiple namespaces based on how they were resolved
                 # but the same type_name should point to the same class
-                if data_type in ns_data_types:
+                if data_type in ns_data_types and not isinstance(ns_data_types[data_type], TypeSource):
                     namespace = ns_key
                     break
         if namespace is None:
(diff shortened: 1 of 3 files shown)


src/hdmf/build/manager.py
```diff
@@ -521,7 +521,7 @@ def get_dt_container_cls(self, **kwargs):
             for ns_key, ns_data_types in self.__container_types.items():
                 # NOTE that the type_name may appear in multiple namespaces based on how they were resolved
                 # but the same type_name should point to the same class
-                if data_type in ns_data_types:
+                if data_type in ns_data_types and not isinstance(ns_data_types[data_type], TypeSource):
```
oruebel, line 524: Does the comment also need to be updated? I'm not sure what `isinstance(ns_data_types[data_type], TypeSource` actually checks so may be worth clarifying. Otherwise this looks good so I approve and you can decide

oruebel (approved):

src/hdmf/build/manager.py
rly: The comment could be a little clearer. I'll make a new PR. Sorry, this was on auto-merge.

PR #1376, opened 2026-01-09
Add MeaningsTable

### LINKED ISSUE #1375: [Feature]: Add MeaningsTable and group to hold them in DynamicTable

To support https://github.com/hdmf-dev/hdmf-common-schema/issues/89

Update DynamicTable, add MeaningsTable class, update the schema submodule

Yes.

rly:
## Motivation

Requires #1374 

Fix #1375 

See https://github.com/hdmf-dev/hdmf-common-schema/pull/91 for schema changes

## How to test the behavior?
```
pytest
```

## Checklist


src/hdmf/common/table.py
```diff
@@ -1721,3 +1781,65 @@ def add_row(self, **kwargs):
...
+
+    __columns__ = (
+        {'name': 'value', 'description': 'The value in the linked VectorData object.', 'required': True},
+        {'name': 'meaning', 'description': 'The meaning of the value.', 'required': True},
+    )
+
+    @docval({'name': 'target', 'type': VectorData,
+             'doc': 'the VectorData object for which this table provides meanings'},
+            {'name': 'name', 'type': str,
+             'doc': 'the name of this table. If not provided, defaults to "{target.name}_meanings".',
```
oruebel: To avoid user error and avoid unexpected behavior, I would propose to in this case remove the the `name` argument from `__init__` and always set it to `f"{target.name}_meanings"`. Since these are always tied to a DynamicTable, I think it should be Ok to restrict the naming in such a way here.

src/hdmf/common/table.py
```diff
@@ -1721,3 +1781,65 @@ def add_row(self, **kwargs):
...
+
+    All possible values of the linked VectorData object should be present in the 'value' column
+    of this table, even if the value is not observed in the data. Additional columns may be
+    added to store additional metadata about each value.
+
+    The name of the MeaningsTable should correspond to the name of the linked VectorData object
+    with a "_meanings" suffix. For example, if the linked VectorData object is named
+    "stimulus_type", the corresponding MeaningsTable should be named "stimulus_type_meanings".
+
+    Note: MeaningsTable does not support containing nested MeaningsTables.
```
oruebel: One possible solution to avoid this issue (and maybe also the circular dependency) would be to:
- Create a new class `DynamicTableBase`, which would essentially be the orginal `DynamicTable` class without the `MeaningsTable` support
- Have `DynamicTable` inherit from `DynamicTableBase` and add the `MeaningsTable` support there
- Have `MeaningsTable` inherit from `DynamicTableBase` so that it doesn't even have the option at all for containing meanings

src/hdmf/common/table.py
```diff
@@ -353,10 +358,17 @@ def __gather_columns(cls, name, bases, classdict):
                      'added to the table if it is not already present (i.e., when it is optional).'),
              'type': dict,
              'default': None},
+            {'name': 'meanings_tables',
+             'doc': ('MeaningsTable objects that provide meanings for values in VectorData columns. '
+                     'The key should be the name of the VectorData column and the value should be the '
+                     'MeaningsTable object. The MeaningsTable will be named "{column_name}_meanings".'),
```
oruebel: >  The MeaningsTable will be named "{column_name}_meanings".'),

I don't think this is actually enforced in the  code. I think we should restrict to not allow any other naming scheme in this case.

src/hdmf/common/table.py
```diff
@@ -533,6 +545,54 @@ def __init__(self, **kwargs):  # noqa: C901
...
+            for mt in val:
+                self.add_meanings_table(mt)
+
+    @docval({'name': 'meanings_table', 'type': 'MeaningsTable',
+             'doc': 'The MeaningsTable to add. It should have a name ending with "_meanings".'})
+    def add_meanings_table(self, **kwargs):
+        """Add a MeaningsTable to this DynamicTable."""
+        meanings_table = getargs('meanings_table', kwargs)
+        if meanings_table.name in self.__meanings_tables:
+            raise ValueError(f"MeaningsTable '{meanings_table.name}' already exists in this DynamicTable")
```
oruebel, line 589: I think we should add a couple of additional checks here:

1. Raise a `ValueError`if the `meanings_table.target` is not a column of this `DynamicTable`
2. Raise a `ValueError` if the `meanings_table.name` is not as expected. This check may be unnecessary or redundant if we enforce the naming elsewhere

src/hdmf/common/table.py
```diff
@@ -533,6 +545,54 @@ def __init__(self, **kwargs):  # noqa: C901
...
+    def get_meanings_table(self, **kwargs):
+        """Get a MeaningsTable from this DynamicTable."""
+        name = getargs('name', kwargs)
+        if name is None:
+            if len(self.__meanings_tables) == 0:
+                raise ValueError(f"No MeaningsTable objects in DynamicTable '{self.name}'")
+            if len(self.__meanings_tables) > 1:
+                raise ValueError(f"Multiple MeaningsTable objects in DynamicTable '{self.name}', "
+                                 "must specify a name")
+            return next(iter(self.__meanings_tables.values()))
```
oruebel: Could you clarify why this is needed? It seems like this code only applies if `len(self.__meanings_tables) == 1` since for `==0` and for `>1` this raises a `ValueError`. I think it may be simpler to just remove this if block and always require a `name` to be provided instead of allowing `None` as default. 

More broadly, I am not sure this function is needed, since I think a user can get the dict with `DynamicTable.get_meanings_tables` and do this lookup easily there. Instead, I think it would be useful to change the function to something like `get_meanings_for_column` where a user provides the `name` of the column and we look up if there is a meanings_table for that by checking for `return self.__meanings_tables[f"{name}_meanings"`

oruebel (changes requested):
Overall, the code looks good. I added a few suggestions/comments.

src/hdmf/common/table.py
rly: Since `DynamicTable` is now effectively a `MultiContainerInterface` of `MeaningsTable` objects, for consistency, this code is copying the logic that would have been generated here for `MultiContainerInterface.get_meanings_table`:
https://github.com/hdmf-dev/hdmf/blob/e24a19c331f4a1f2412928a8bab84a21186065e1/src/hdmf/container.py#L1103-L1127

rly: > I think it would be useful to change the function to something like `get_meanings_for_column` where a user provides the name of the column and we look up if there is a `meanings_table` for that by checking for return `self.__meanings_tables[f"{name}_meanings"]`

Good idea

rly: I think it's fine to change the copied behavior since this is not actually a `MultiContainerInterface` and I agree, it would be cleaner to just require a name.

src/hdmf/common/table.py
```diff
@@ -533,6 +546,61 @@ def __init__(self, **kwargs):  # noqa: C901
...
+        if target_name not in self:
+            raise ValueError(f"MeaningsTable target '{target_name}' is not a column in DynamicTable '{self.name}'")
+        if not isinstance(meanings_table.parent, Container):
+            meanings_table.parent = self
+        else:
+            self.set_modified()
+        self.__meanings_tables[meanings_table.name] = meanings_table
+
+    @docval({'name': 'name', 'type': str,
+             'doc': 'The name of the MeaningsTable to get.'},
```
oruebel, line 601: I'm wondering whether that should be the name of meanings table or the name of the column? Or maybe it could make sense to add something like:

```
def get_column_meanings(self, col_name):
     return self.get_meanings_table(f"{col_name}_meanings")
```

rly: How is that different from the new `get_meanings_for_column`?

src/hdmf/common/table.py
rly: Done

rly: The name check is in the MeaningsTable constructor

src/hdmf/common/table.py
rly: I updated the code around this

src/hdmf/common/table.py
rly: If we do this, should we update the schema as well to make it consistent with the API? I would call it `BaseDynamicTable` in that case to be consistent with `BaseImage` https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/core/nwb.base.yaml#L26

src/hdmf/common/table.py
```diff
@@ -533,6 +564,61 @@ def __init__(self, **kwargs):  # noqa: C901
...
+        """Get a MeaningsTable from this DynamicTable by name."""
+        name = getargs('name', kwargs)
+        if name not in self.__meanings_tables:
+            raise KeyError(f"MeaningsTable '{name}' not found in DynamicTable '{self.name}'")
+        return self.__meanings_tables[name]
+
+    @docval({'name': 'col_name', 'type': str,
+             'doc': 'The name of the column to get the MeaningsTable for.'},
+            returns='the MeaningsTable for the given column', rtype='MeaningsTable')
+    def get_meanings_for_column(self, **kwargs):
```
oruebel, line 613:
```suggestion
            returns='the MeaningsTable for the given column', rtype='MeaningsTable')
            
    def get_meanings_for_column(self, **kwargs):
```

src/hdmf/common/table.py
oruebel: Same thing. Sorry, I didn't see that `get_meanings_for_column` was already added.

src/hdmf/common/table.py
oruebel: > If we do this, should we update the schema as well to make it consistent with the API?

Yes, that makes sense. I think there will be other cases where an extension will only want a table without the options for adding meanings, so I think preserving that functionality by creating a `BaseDynamicTable` schema and then `DynamicTable` adds the `MeaningsTable` feature makes sense.

src/hdmf/common/table.py
rly: Did you mean to add a line of whitespace? I don't see any other differences

oruebel: Yes, that was all. It looked like the line break was missing.

src/hdmf/common/table.py
```diff
@@ -79,6 +79,24 @@ def extend(self, ar, **kwargs):
...
+        parent = self.parent
+        if parent is None:
+            return None
+        # Check if parent is a DynamicTable (avoid circular import by checking for the method)
+        if not hasattr(parent, 'get_meanings_for_column'):
+            return None
+        try:
+            return parent.get_meanings_for_column(self.name)
+        except KeyError:
+            return None
```
oruebel:
```suggestion
        try:
            return parent.get_meanings_for_column(self.name)
        except KeyError:
            return None
        except AttributeError:  # parent is not a DynamicTable, i.e, missing  get_meanings_for_column
            return None
```

Just a suggestion, but it's also fine as is. Just seemed a bit cleaner since the call is already in a try/except block. Your choice.

src/hdmf/common/table.py
```diff
@@ -1412,6 +1430,85 @@ def copy(self):
...
+        col_name = getargs('col_name', kwargs)
+        meanings_table_name = f"{col_name}_meanings"
+        if meanings_table_name not in self.__meanings_tables:
+            raise KeyError(f"No MeaningsTable found for column '{col_name}' in DynamicTable '{self.name}'")
+        return self.__meanings_tables[meanings_table_name]
+
+
 @register_class('DynamicTableRegion')
 class DynamicTableRegion(VectorData):
     """
```
oruebel, line 1501: I think we probably also want to change `DyanmicTableRegion` to have `BaseDynamicTable` as target to allow referencing of rows in any subclass of `BaseDynamicTable`, incl. `DynamicTable`.

oruebel: I think this can be done a separate PR or here, whichever you prefer.

src/hdmf/common/table.py
rly: We don't usually put a line break between docval and the method though

src/hdmf/common/table.py
```diff
@@ -1721,3 +1818,58 @@ def add_row(self, **kwargs):
...
+    @docval({'name': 'value', 'type': None, 'doc': 'the value in the linked VectorData object'},
+            {'name': 'meaning', 'type': str, 'doc': 'the meaning of the value'},
+            {'name': 'id', 'type': int, 'doc': 'the ID for the row', 'default': None},
+            {'name': 'enforce_unique_id', 'type': bool, 'doc': 'enforce that the id in the table must be unique',
+             'default': False},
+            allow_extra=True)
+    def add_row(self, **kwargs):
+        """Add a row to the table mapping a value to its meaning."""
+        value, meaning = popargs('value', 'meaning', kwargs)
+        super().add_row(value=value, meaning=meaning, **kwargs)
```
oruebel:
```suggestion
        super().add_row(**kwargs)
```

Again, this up to you. The code is fine as is. I just wasn't sure why we need popargs when we just pass them through as keyword arguments anyways.

src/hdmf/common/table.py
rly: Ah, I missed that. Thanks.

src/hdmf/spec/namespace.py
```diff
@@ -470,7 +470,8 @@ def __load_namespace(self, namespace, reader, resolve=True):
                     types_to_load = inc_ns.get_registered_types()  # load all types in namespace
                 registered_types = set()
                 for ndt in types_to_load:
-                    self.__register_type(ndt, inc_ns, catalog, registered_types)
+                    in_progress_registrations = set()
+                    self.__register_type(ndt, inc_ns, catalog, registered_types, in_progress_registrations)
```
oruebel:
```suggestion
                    self.__register_type(ndt, inc_ns, catalog, registered_types)
```

I think this change might be a straggler from the other PR, where I think this was changed to have the default for `in_progress_registrations=None`

oruebel (changes requested):
I added a few minor additional suggestion. Feel free to accept/reject as you see fit and ping me again

src/hdmf/common/table.py
oruebel: Sounds good. I think I misread it as a missing space to the code of another function, rather than docval.

rly:
## Question: Feedback on BaseDynamicTable / DynamicTable Schema Refactoring

We're considering refactoring the DynamicTable schema hierarchy to support MeaningsTable. I'd like feedback on whether this approach is the right design choice.

### Proposed Implementation

**Schema (hdmf-common-schema/common/table.yaml):**
```yaml
BaseDynamicTable:
  data_type_inc: Container
  # Core table functionality: id, colnames, description, VectorData columns

DynamicTable:
  data_type_inc: BaseDynamicTable
  groups:
    - name: meanings_tables (quantity: '?')
      groups:
        - data_type_inc: MeaningsTable (quantity: '*')

MeaningsTable:
  data_type_inc: BaseDynamicTable  # NOT DynamicTable
  datasets: [value, meaning]
  links: [target -> VectorData]
```

**Python (hdmf/common/table.py):**
- `BaseDynamicTable`: Core table functionality (add_row, add_column, DataFrame conversion, etc.)
- `DynamicTable`: Adds `meanings_tables` field and related methods
- `MeaningsTable`: Extends BaseDynamicTable directly

### Rationale
MeaningsTable should not have its own MeaningsTable (a meanings table for a meanings table is nonsensical), so MeaningsTable extends BaseDynamicTable to avoid inheriting the `meanings_tables` group.

### Pros
1. **Clean semantics**: MeaningsTable cannot have nested MeaningsTable objects — enforced at schema level, providing type safety and clear intent

### Cons
1. **Design/codebase changes**:
   - Code checking `isinstance(obj, DynamicTable)` won't match MeaningsTable
   - DynamicTableRegion points to BaseDynamicTable instead of DynamicTable
   - API code and documentation across all table-related types (e.g., VectorData, DynamicTableRegion, AlignedDynamicTable) now references BaseDynamicTable instead of DynamicTable
2. **Increased complexity**: Three schema types instead of two; users must understand when to use each

Since we expect most users to use DynamicTable, documentation and tutorials will continue to demonstrate use of DynamicTable instead of BaseDynamicTable.

### Alternative Approach
Keep a single DynamicTable type and:
- Make `meanings_tables` group optional (already is with `quantity: '?'`)
- Add validation/restriction in each API that MeaningsTable cannot contain meanings_tables
- Simpler hierarchy, same end result

### Question
Is the structural enforcement (separate schema types) worth the added complexity, or would validation/restriction in each API on a single DynamicTable type be preferable?

oruebel:
> We're considering refactoring the DynamicTable schema hierarchy to support MeaningsTable.

Thanks for the helpful summary. 

> Pros

Here I would add: 
- Enables extensions to use a table as is (without Meanings) rather than always having MeaningsTable in there. On the flip-side this is a Con of the alternative approach, which modifies all existing schema to add MeaningsTable without having the ability to remove it. 

> Cons

Here, I think the main con is the effort and potential unexpected larger downstream code-changes this may require.

> Alternative Approach
> * Add validation/restriction in each API that MeaningsTable cannot contain meanings_tables

Here, I would argue that we should not artificially restrict this in the API if the schema allows it. A check in the nwb-inspector would make sense or maybe even a warning in the API, but if the schema allows it, then adding this as custom logic seems unwieldy. Theoretically, defining meanings for columns of a meanings table is I think a valid thing to do, in particular since users can add custom columns in a MeaningsTable (which could have information that requires further definition), however, I agree that in most cases this is probably not the right way to do it. 

I.e., I would argue that we either use:

1. use the proposed approach of adding `BaseDynamicTable` as a new base class, or
2. keep the single DynamicTable class (with options for MeaningsTables), which means allowing users to define meanings for columns of MeaningsTable (with maybe a best practice check in the nwb-inspector)

Personally, I think option 1 is cleaner, but will require significant effort such that the cost for implementing this may out-way the benefit. I think option 2 should be fine, but may require changes down-the-line if/when a real use-case for a `BaseDynamicTable` should arise. 

Long story short, I think either option should meet the current needs reasonably well, and I think it is mainly a question of spending effort now to prepare for potential use cases vs. waiting for real use cases and spending the effort then. I am personally fine with either approach right now, but the more I think about, I'm starting to lean towards Option 2 (even-though I originally proposed Option 1 😉 )

h-mayorquin:
I vote for option 2. I think that option 1 is a large modification just to exclude what seems like an edge case.

Also, is there similar art on the data frame data model that we can use as a compass?

bendichter:
Sorry for the delay on this. I discussed with Claude we came up with this:

I'd side with Option 2 as well, for these reasons:                                                                 
                                                                                                                     
1. [**YAGNI principle:**](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) The `BaseDynamicTable` abstraction is being introduced primarily to prevent an edge case (meanings of meanings) that users are unlikely to encounter. If no real use case exists yet for `BaseDynamicTable`  beyond this restriction, it's premature abstraction.
2. **Breaking change surface area:** As noted in the cons, code checking `isinstance(obj, DynamicTable)` won't match `MeaningsTable`, and `DynamicTableRegion` now points to `BaseDynamicTable`. This creates downstream compatibility work for a relatively minor semantic benefit.
3. **Schema vs. validation tradeoff:** While schema-level enforcement is cleaner in theory, the practical benefit is minimal here. A meanings-of-meanings scenario would be unusual but not necessarily invalid (as @oruebel noted -custom columns in MeaningsTable could theoretically need definitions).
4. **Reversibility:** Starting with Option 2 doesn't preclude adding `BaseDynamicTable` later if a real use case emerges. Going the other direction would be harder.

rly:
Thanks @bendichter! I was also leaning toward Option 2. We will go with Option 2 and also not artificially restrict the addition of MeaningsTables to MeaningsTables in the API.

oruebel:
> I'd side with Option 2 as well, for these reasons:

Thanks @bendichter  for the thorough assessment!

> We will go with Option 2 and also not artificially restrict the addition of MeaningsTables to MeaningsTables in the API.

This sounds good to me. Thanks for your patience.

oruebel (dismissed):

oruebel (approved):

PR #1381, opened 2026-01-22
Add HTML repr to `LabelledDict`

h-mayorquin:
## Motivation

Right now it looks like this:
image

After this PR:
image

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Added support for Python 3.14. @bendichter [#1366](https://github.com/hdmf-dev/hdmf/issues/1366)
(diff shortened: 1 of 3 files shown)


rly:
Can you add tests for the branches (empty case and nested case)? Otherwise looks good to me

rly:
I can add this into the 4.3.1 release today if it's ready today.

h-mayorquin:
> I can add this into the 4.3.1 release today if it's ready today.

Let me add the tests

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # HDMF Changelog
 
+## HDMF 4.3.1 (Upcoming)
```
h-mayorquin, line 3: So we probably want to move this if this is being released today, right?

rly (approved):

CHANGELOG.md
rly: This is fine. I'll update the changelog to the appropriate date in a separate PR.

PR #1388, opened 2026-02-02
Fix circular dependency handling in spec resolution

rly:
## Summary
- Fix `ValueError: Cycle detected in specification dependencies` caused by the DynamicTable ↔ MeaningsTable circular dependency in the updated hdmf-common-schema
- Fix infinite recursion during spec resolution for types involved in circular dependencies
- Fix non-deterministic spec resolution ordering that caused roundtrip test failures in some environments

## Problem
PR #1312 (`4c6f1ef`) introduced topological sorting for spec resolution, and PR #1374 (`aae6b62`) added circular dependency support during type registration. These two features conflict: the topological sort in `__resolve_namespace_specs` does not tolerate cycles, and the resolution logic in `__resolve_local` can infinitely expand specs that reference each other.

Additionally, `graphlib.TopologicalSorter.static_order()` does not guarantee ordering among types at the same topological level — the order depends on insertion order into the graph. This caused non-deterministic resolution: e.g., MeaningsTable and AlignedDynamicTable are at the same level, and if AlignedDynamicTable was resolved before MeaningsTable, it would get a stale (incompletely resolved) copy of MeaningsTable, causing `test_write_cache_spec` failures in CI.

## Changes
All changes are in `src/hdmf/spec/namespace.py`:

1. **`__get_spec_dependencies`**: Skip edges that would create cycles in the topological sort. Uses `catalog.get_hierarchy` to detect when a nested subspec type inherits (at any depth) from the current type.

2. **`__resolve_local`**: Add a `resolving_in_progress` set to track types currently being resolved, preventing infinite recursion. After resolving a spec's included type, add it to the set (via copy, not mutation, so siblings are unaffected). Iterate direct children only, letting recursion handle deeper nesting.

3. **`__resolve_namespace_specs`**: Replace `static_order()` with level-by-level processing (`prepare()`/`get_ready()`/`done()`) and sort types alphabetically within each topological level. This ensures deterministic resolution order regardless of catalog insertion order.

4. **Rename** `in_progress_registrations` → `registering_in_progress` for consistency with `resolving_in_progress`.

## Checklist

### DIFF
--- a/src/hdmf/spec/namespace.py
+++ b/src/hdmf/spec/namespace.py
@@ -431,23 +431,20 @@ class NamespaceCatalog:
         return nested_subspecs
 
     def __get_spec_dependencies(self, spec: BaseStorageSpec, catalog: SpecCatalog) -> set[tuple[str, str]]:
-        """Get the set of edges representing the dependencies of the given spec."""
+        """Get the set of edges representing the dependencies of the given spec.
+
+        Edges that would create cycles are skipped. This includes self-referential types
+        (A contains A) and indirect cycles (A contains B, B extends A).
+        """
         edges = set()
         if spec.data_type_inc is not None:
-            # The included spec should be resolved before this spec
             edges.add((spec.data_type_def, spec.data_type_inc))
         if isinstance(spec, GroupSpec):
-            # For each nested subspec, the included specs of that nested subspec should be resolved before
-            # this spec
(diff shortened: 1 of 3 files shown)


oruebel (approved):

rly:
Oops. I did not mean to merge this while there are still unresolved CI failures for "x-upgraded-optional" workflows. I cannot reproduce these locally so this is proving difficult.

PR #1389, opened 2026-02-03
Fix issues with optional deps in #1388

rly:
Fix issues with optional deps in #1388. There are unresolved CI failures for "x-upgraded-optional" workflows. I cannot reproduce these locally so this is proving difficult.


rly:
Closing to make merge clean after force push to release/5.0.0

PR #1392, opened 2026-02-05
Fix issues with optional deps in #1388, take 2

rly:
Fix issues with optional deps in https://github.com/hdmf-dev/hdmf/pull/1388. There are unresolved CI failures for "x-upgraded-optional" workflows. I cannot reproduce these locally so this is proving difficult.

Refactored the spec resolution approach to use a recursive DFS approach instead of graphlib.TopologicalSorter, which is more complex and can be non-deterministic.


src/hdmf/backends/utils.py
```diff
@@ -65,44 +65,24 @@ def convert_namespace(cls, **kwargs):
                 builder.include_namespace(inc_ns)
             else:
                 source = elem['source']
-                for dt in ns_catalog.get_types(source):
-                    spec = ns_catalog.get_spec(namespace, dt)
-                    if spec.parent is not None:
-                        continue
-                    h5_source = cls.get_source_name(source)
-                    spec = cls.__copy_spec(spec)
```
rly, line 73: Previously, `convert_namespace` would take the in-memory, partially resolved spec, and unresolve the spec in `__copy_spec`. Downstream code would cache that to the file. Now, we resolve the spec more completely, and instead of trying to unresolve the spec, we just retrieve and cache the originally loaded spec, because I believe the cached spec should match the spec YAML files.

rly:
@oruebel This is finally ready for your review

src/hdmf/backends/utils.py
```diff
@@ -65,44 +65,16 @@ def convert_namespace(cls, **kwargs):
...
+                h5_source = cls.get_source_name(source)
```
oruebel: Should `h5_source` be called `file_source`? I think ZarrIO probably uses the class too.

oruebel (approved):
Looks good to me. Although, I have to admit that some of the details of the logic are not 100% clear to me without taking more of a deep-dive into the code.

src/hdmf/backends/utils.py
rly: Good catch. Updating now

oruebel (approved):

PR #1399, opened 2026-02-11
Require timezone information for isodatetime validation

### LINKED ISSUE #1284: [Bug]: Validator does not raise warning for missing time zone
### What happened?

See the original issue here: https://github.com/NeurodataWithoutBorders/pynwb/issues/2084.

The validator currently does not raise a warning if a datetime is provided without required time zone information. I think this would require an update to the HDMF validator

### Steps to Reproduce

```python
See the original issue.
```

### Operating System

macOS

### Python Version

3.13

### Package Versions

_No response_

sejalpunwatkar:
This PR adds validation to ensure that values with string_format="isodatetime" include timezone information.

Specifically:

- Added a has_timezone() helper function to detect timezone presence.

- Integrated timezone validation in both AttributeValidator and DatasetValidator.

- Raises a validation Error if an ISO datetime string is missing timezone information.

- Existing type and shape validation behavior remains unchanged.

Please let me know if enforcing timezone as a requirement aligns with the intended specification behavior.

Fix #1284


rly:
Thank you @sejalpunwatkar . Please add tests to verify that your fix worked. The tests should fail prior to this change and succeed with this change.

sejalpunwatkar:
Thanks for the feedback, @rly . I will add unit tests to tests/unit/test_validator.py (or the relevant test suite) that include ISO datetime strings both with and without timezones to ensure the new validation logic triggers correctly and prevents regressions.

src/hdmf/validate/validator.py
```diff
@@ -89,6 +89,19 @@ def get_string_format(data):
         pass
     return None
 
+def has_timezone(datetime_string):
+    """Return True if ISO datetime string contains timezone info."""
+    s= pystr(datetime_string)
+
+    if s.endswith("Z"):
+        return True
```
rly, line 97: Can you also add a test for this condition?

rly:
Thanks @sejalpunwatkar . This is looking good. Can you also add a test for validating DatasetSpec with dtype=isodatetime?

rly:
Ideally the tests in each PR cover each change made in the PR. You can look at the codecov report above to see what lines were not touched by the tests. It's easy for a bug to appear if the code isn't tested.

rly:
When you're done, could you please also add an entry to the CHANGELOG.md for version 5.0.0 with a brief description of the change. Follow the format of the other changelog entries. This would go under "Fixed"

sejalpunwatkar:
Thanks for the detailed feedback @rly ! I’ll get started on these updates right away, including the 'Z' suffix test, the DatasetSpec validation, and the CHANGELOG entry. I'll also make sure to hit the missing lines to get that Codecov report back to green.
I'll have the commits pushed shortly so they're ready for your review whenever you're back. enjoy the break

sejalpunwatkar:
Updates made:

- Enforced timezone requirement for isodatetime dtype validation.

Added tests for:

- Valid timestamp with explicit offset (+02:00)

- Missing timezone (expected validation error)

- UTC "Z" suffix case

- Added a corresponding entry in CHANGELOG.md.

All tests are passing locally.

rly:
Thanks for the updates @sejalpunwatkar . Could you please address the issues raised by copilot? They are mostly the same issue. I do think we need to test arrays of datetimes, that checking for "+"/"-" can lead to false positives, and that the variable names could be made clearer. The arrays issue is causing a test failure in the pynwb tests. Any existing hdmf test(s) where a datetime without time zone is used should be updated because that is not valid. Thank you!

sejalpunwatkar:
Updated validator.py to enforce that isodatetime datasets include timezone information.

- Modified AttributeValidator and DatasetValidator to check timezone for scalar and array data (lists, tuples, numpy arrays).

- Added comprehensive unit tests in test_validate.py to cover:

- Scalar datetimes with/without timezone

- Arrays with mixed timezone presence

- Empty arrays

- Numpy arrays

- Python datetime objects

- Ensures validation raises an Error when timezone is missing and passes when present.

oruebel:
@sejalpunwatkar thanks for the updates! We are in a bit of crunch this week so it'll likely be later next week before we get back to this on our end. Feel free to re-ping at the end of next week if you have not seen activity here.

sejalpunwatkar:
No problem at all, @oruebel  . I appreciate the update! I'll use this time to verify that every test runs smoothly and keep the PR ready for your review. I'll re-ping you at the end of next week as suggested. Hope the crunch goes well!

src/hdmf/validate/validator.py
```diff
@@ -89,11 +89,37 @@ def get_string_format(data):
...
+
+
+    if s.endswith("Z"):
+        return True
+
+    # Timezone offsets only valid if there is a time component
+    if "T" not in s:
+        return False
+
+     # Check for valid timezone offset: +HH:MM or -HH:MM at the end
```
```suggestion
    # Check for valid timezone offset: +HH:MM or -HH:MM at the end
```

src/hdmf/validate/validator.py
```diff
@@ -349,7 +394,6 @@ def validate(self, **kwargs):
                         value_type, _ = get_type(value)
                         ret.append(DtypeError(self.get_spec_loc(spec), expected, value_type))
                     except EmptyArrayError:
-                        # do not validate dtype of empty array. HDMF does not yet set dtype when writing a list/tuple
```
rly, line 352: I believe this change is unrelated, and this comment should not be deleted.

src/hdmf/validate/validator.py
```diff
@@ -331,14 +359,31 @@ class AttributeValidator(Validator):
...
+                                 "Datetime is missing required timezone information."))
+
     @docval({'name': 'value', 'type': None, 'doc': 'the value to validate'},
             returns='a list of Errors', rtype=list)
     def validate(self, **kwargs):
         value = getargs('value', kwargs)
         ret = list()
         spec = self.spec
-        if spec.required and value is None:
-            ret.append(MissingError(self.get_spec_loc(spec)))
```
rly, line 341: This code that validates that a required element is present should not be deleted. I believe the new check should be added as an `elif` here.

src/hdmf/validate/validator.py
```diff
@@ -413,6 +458,34 @@ class DatasetValidator(BaseStorageValidator):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
+    def _validate_isodatetime_timezone(self, data, builder, ret):
```
rly: This function and the `_validate_isodatetime_timezone` function above are nearly the same and should be extracted into a utility function that can be called from both places. The `Error` constructor is different. I suggest having the utility function return False and the calling code do:
```python
                    ret.append(
                        Error(
                            self.get_spec_loc(self.spec),
                            "Datetime is missing required timezone information.",
                            location=self.get_builder_loc(builder)
                    )
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,321 @@ def test_utf8_for_ascii_hdf5(self):
         self.assertEqual(len(results), 1)
         self.assertIsInstance(results[0], DtypeError)
         self.assertEqual("Foo/data (my_foo/data): incorrect type - expected 'bytes', got 'utf'", str(results[0]))
+
+class TestISODateTimeTimezone(ValidatorTestBase):
+    """Test that isodatetime dtype requires timezone information."""
```
```suggestion
class TestISODateTimeAttributeTimezone(ValidatorTestBase):
    """Test that an AttributeSpec with isodatetime dtype requires timezone information."""
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,321 @@ def test_utf8_for_ascii_hdf5(self):
...
+                'data_type': 'DateTimeTest',
+                'dt': '2026-02-12T10:30:00'
+            }
+        )
+        result = self.vmap.validate(builder)
+        self.assertEqual(len(result), 1)
+        self.assertIsInstance(result[0], Error)
+
+class TestISODateTimeDatasetTimezone(ValidatorTestBase):
+    """Test that isodatetime dtype in DatasetSpec requires timezone information."""
```
```suggestion
    """Test that a DatasetSpec with isodatetime spec requires timezone information."""
```

sejalpunwatkar:
Hi @rly , I've addressed all the feedback and pushed the updates:
Refactored Validation: Extracted the timezone check into a standalone utility function _is_missing_timezone to keep the code DRY and reusable.
Restored Original Logic: Re-inserted the spec.required validation and the EmptyArrayError catch/comments that were previously removed.
Updated Error Handling: Used the specific Error constructors with correct location information for both attributes and datasets.
Tests Updated: Renamed the test classes to TestISODateTimeAttributeTimezone and TestISODateTimeDatasetTimezone and updated the docstrings as suggested.
All 1663 local tests are passing. Ready for your final review!

src/hdmf/validate/validator.py
```diff
@@ -413,6 +450,18 @@ class DatasetValidator(BaseStorageValidator):
...
```
rly: This function is very simple now. Could you please integrate it into line 474 below where it is called, and can you make it consistent with the attribute validation code on line 371?

src/hdmf/validate/validator.py
```diff
@@ -735,3 +787,17 @@ def is_unsatisfied(spec_matches):
...
+            # Flatten array/iterator for check
+            iterator = [data.item()] if isinstance(data, np.ndarray) and data.ndim == 0 else \
+                       (data.flat if isinstance(data, np.ndarray) else iter(data))
+            for v in iterator:
+                if isinstance(v, str) and not has_timezone(v):
+                    return True  # Found one missing a TZ
+        elif isinstance(data, str):
+            if not has_timezone(data):
+                return True  # Single string is missing a TZ
+        return False
```
```suggestion
    """Utility to check if isodatetime data is missing timezone info."""
    if isinstance(data, (list, tuple, np.ndarray)):
        # Flatten array/iterator for check
        iterator = [data.item()] if isinstance(data, np.ndarray) and data.ndim == 0 else \
                   (data.flat if isinstance(data, np.ndarray) else iter(data))
        for v in iterator:
            if isinstance(v, str) and not has_timezone(v):
                return True  # Found one missing a TZ
    elif isinstance(data, str):
        if not has_timezone(data):
            return True  # Single string is missing a TZ
    return False
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,321 @@ def test_utf8_for_ascii_hdf5(self):
...
+
+        builder = GroupBuilder(
+            name='test_group',
+            attributes={'data_type': 'DateTimeDatasetTest'},
+            datasets=[ds_builder]
+        )
+        result = self.vmap.validate(builder)
+        self.assertTrue(any("timezone" in str(e).lower() for e in result))
+
+    def test_dataset_isodatetime_numpy_scalar_no_timezone(self):
```
rly: This test is basically the same as `test_dataset_isodatetime_numpy_scalar_without_timezone`. You can add the `"timezone" in str(result[0])` check to that test.

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,321 @@ def test_utf8_for_ascii_hdf5(self):
...
+        builder = GroupBuilder(
+            name='dt',
+            attributes={'data_type': 'DateTimeDatasetTest'},
+            datasets=[DatasetBuilder(name='dt', data=np.array('2026-02-12T10:30:00'))]
+        )
+        result = self.vmap.validate(builder)
+        self.assertTrue(any("timezone" in str(e).lower() for e in result))
+
+    def test_dataset_isodatetime_list_no_timezone(self):
```
rly: Same comment as above

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,321 @@ def test_utf8_for_ascii_hdf5(self):
...
+        builder = GroupBuilder(
+            name='dt',
+            attributes={'data_type': 'DateTimeDatasetTest'},
+            datasets=[DatasetBuilder(name='dt', data=['2026-02-12T10:30:00'])]
+        )
+        result = self.vmap.validate(builder)
+        self.assertTrue(any("timezone" in str(e).lower() for e in result))
+
+    def test_dataset_isodatetime_raw_string_no_timezone(self):
```
rly: Same comment as above, for `test_dataset_isodatetime_without_timezone`

src/hdmf/validate/validator.py
```diff
@@ -337,40 +365,50 @@ def validate(self, **kwargs):
...
+                data_type = value.attributes.get(target_spec.type_key())
+                hierarchy = self.vmap.namespace.catalog.get_hierarchy(data_type)
+                if spec.dtype.target_type not in hierarchy:
+                    ret.append(IncorrectDataType(self.get_spec_loc(spec), spec.dtype.target_type, data_type))
+
+        else:
+            try:
+                dtype, string_format = get_type(value)
+
+                if not check_type(spec.dtype, dtype, string_format):
```
rly: For consistency with DatasetValidator below, I would actually add the new if statement from above `spec.dtype == "isodatetime" and _is_missing_timezone(value):` here, with the additional guard on string_format == "isodatetime"

src/hdmf/validate/validator.py
```diff
@@ -89,11 +89,37 @@ def get_string_format(data):
...
+    """
+    # If input is array-like, validate only first element
+    if isinstance(data, (list, tuple)):
+        data = data[0]
+    elif hasattr(data, "shape") and len(data) == 1:
+        data = data.item()
+
+    s = pystr(data)
+
+
```
```suggestion
```

src/hdmf/validate/validator.py
```diff
@@ -89,11 +89,36 @@ def get_string_format(data):
         pass
     return None
 
+def has_timezone(data):
```
rly: Looking at how this is called in `_is_missing_timezone`, `has_timezone` has dead array-handling code. `_is_missing_timezone` iterates containers and always passes scalar strings to `has_timezone`. But `has_timezone` still handles lists/tuples/arrays below. 

This is unreachable in practice and makes the function confusing. The docstring says "Handles scalar strings or single-element arrays/lists" but it's only ever called with scalars. Should be simplified to just accept a string.

src/hdmf/validate/validator.py
```diff
@@ -735,3 +786,17 @@ def is_unsatisfied(spec_matches):
...
 
         return list(filter(is_unsatisfied, candidates))
+
+def _is_missing_timezone(data):
+    """Utility to check if isodatetime data is missing timezone info."""
+    if isinstance(data, (list, tuple, np.ndarray)):
+        # Flatten array/iterator for check
+        iterator = [data.item()] if isinstance(data, np.ndarray) and data.ndim == 0 else \
+                   (data.flat if isinstance(data, np.ndarray) else iter(data))
+        for v in iterator:
```
rly: This will work only for scalar and 1-dimensional arrays/lists/tuples. Please add support for arrays/lists/tuples with more than 1 dimension. (This doesn't happen in core NWB right now, but it could happen in extensions or custom table columns.)

rly:
Sorry for all the back and forth @sejalpunwatkar . I left a few more comments for you to address. Thank you!

sejalpunwatkar:
No worries at all, @rly ! I really appreciate the detailed feedback and the opportunity to make the code cleaner and more robust. It’s been a great learning process.
I’ve addressed all the latest comments:
1. Integrated Validation: Moved the timezone check directly into AttributeValidator and DatasetValidator and removed the standalone helper method for better consistency.
2. Simplified has_timezone: Refactored the function to focus strictly on scalar strings, as the container iteration is now handled upstream.
3. Multi-dimensional Support: Updated _is_missing_timezone to handle multi-dimensional numpy arrays (using np.nditer) and nested lists/tuples.
4. Test Cleanup: Consolidated the redundant test cases into the original project tests and added the 'timezone' error message verification as suggested.
All 1663 local tests are passing. Ready for your final review whenever you have a moment

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,296 @@ def test_utf8_for_ascii_hdf5(self):
...
+                    '2026-02-12T12:00:00+02:00'
+                   ]
+               )
+            ]
+        )
+        result = self.vmap.validate(builder)
+        self.assertEqual(len(result), 0)
+
+    def test_dataset_isodatetime_with_datetime_objects(self):
+        builder = GroupBuilder(
```
```suggestion
        # NOTE: In practice, builders should never contain isodatetime objects, 
        # but we test this just in case.
        builder = GroupBuilder(
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,298 @@ def test_utf8_for_ascii_hdf5(self):
...
+                   data=[
+                       '2026-02-12T10:30:00',
+                       '2026-02-12T12:00:00'
+                   ]
+               )
+            ]
+        )
+        result = self.vmap.validate(builder)
+        self.assertEqual(len(result), 1)
+        self.assertIsInstance(result[0], Error)
```
```suggestion
        self.assertIsInstance(result[0], Error)
        self.assertTrue(any("timezone" in str(e).lower() for e in result))
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,299 @@ def test_utf8_for_ascii_hdf5(self):
...
+        )
+
+        builder = GroupBuilder(
+            name='test_group',
+            attributes={'data_type': 'DateTimeDatasetTest'},
+            datasets=[ds_builder]
+        )
+        result = self.vmap.validate(builder)
+        self.assertTrue(any("timezone" in str(e).lower() for e in result))
+
```
```suggestion
```
This is now covered by `test_dataset_isodatetime_array_all_without_timezone` above

rly:
Looks good @sejalpunwatkar ! I made a couple minor changes. 

One last comment: `TestISODateTimeAttributeTimezone` only tests a couple cases, whereas `TestISODateTimeDatasetTimezone` is thorough. We should test the same cases for attributes as for datasets. However, I do not want all of that code to be simply copied and pasted with slight modifications for testing attributes vs datasets. Can you refactor the boilerplate code for these tests and add the same tests for attributes that you have for datasets? It might help to have `getSpecs` return a GroupSpec with an optional DatasetSpec (`quantity='?'`) and an optional AttributeSpec (`required=False`) so that you can use that for both attribute testing and dataset testing.

sejalpunwatkar:
Hi @rly , I've finished the refactor! I ran into some merge conflicts earlier which have now been resolved. The tests are passing locally (1659 passed), and the code is now unified into a single class with the shared helper as requested. Ready for another look!

tox.ini
```diff
@@ -51,6 +51,7 @@ commands =
 [testenv:pytest-py{313,314}-upgraded-optional]
 [testenv:pytest-py314-prerelease]
 [testenv:pytest-py{313,314}-prerelease-optional]
+
```
rly:
```suggestion
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1723,3 +1724,153 @@ def test_utf8_for_ascii_hdf5(self):
...
+                self.assertIsInstance(result[0], Error)
+                self.assertTrue("timezone" in str(result[0]).lower())
+
+    def test_isodatetime_list_without_timezone(self):
+        data = ['2026-02-12T10:30:00', '2026-02-12T11:30:00']
+        for is_attr in [True, False]:
+            with self.subTest(is_attr=is_attr):
+                result = self.check_iso(data, is_attr=is_attr)
+                self.assertTrue(any("timezone" in str(e).lower() for e in result))
+
```
```suggestion
```
This is already covered by `test_isodatetime_array_all_without_timezone` above

src/hdmf/validate/validator.py
```diff
@@ -337,40 +383,51 @@ def validate(self, **kwargs):
...
+                        )
+                    )
+
+                if not check_type(spec.dtype, dtype, string_format):
+                    ret.append(DtypeError(self.get_spec_loc(spec), spec.dtype, dtype))
+            except EmptyArrayError:
+                # do not validate dtype of empty array. HDMF does not yet set dtype when writing a list/tuple
+                pass
+
+        shape = get_data_shape(value)
```
rly: After refactoring the if/else into if/elif/else, this shape check now runs when value is None. That should not happen. I'll move it back.

rly (approved):
Looks good. Thank you for your sustained effort on this @sejalpunwatkar !

sejalpunwatkar:
Thank you, @rly ! I appreciate the support in getting this across the finish line. Looking forward to helping out with more issues soon.

PR #1403, opened 2026-02-13
Reduce memory allocations in ragged column writing by using `np.concatenate`

h-mayorquin:
## Motivation

I am working on converting sessions from the IBL project to NWB, and some of these sessions contain large ragged array columns on the units table. During writing, I observed total memory allocations of around 24 GiB, with some sessions exceeding 32 GiB, so I am concerned about reducing the memory footprint. Profiling reveals that a meaningful part of the allocations comes from `DynamicTable.add_column(..., index=True)`: when it receives ragged data, it flattens it using `list(itertools.chain.from_iterable(data))`, which iterates element-by-element through every numpy array, boxing each value into an individual Python object and appending it to a growing list.

This PR adds a `_flatten_one_ragged_level` helper that detects when all entries are compatible numpy arrays and flattens them with a single `np.concatenate` call. Unlike the `itertools.chain` path, `np.concatenate` computes the total output size upfront, allocates one contiguous buffer, and copies raw memory without creating any intermediate Python objects. The function validates shape compatibility (matching ndim and trailing dimensions) and falls back to the original `itertools.chain` path for non-array or mixed-type data, so behavior is unchanged for edge cases. 

## How to test the behavior?
<details>
<summary>Profiling script</summary>

```python

import itertools
import os
import resource

import numpy as np

import hdmf.common.table as table_mod
from hdmf.common.table import DynamicTable

try:
    import psutil
except ImportError:  # pragma: no cover
    psutil = None

# Toggle between "current" and "old" for before/after profiling.
MODE = "old"
NUM_UNITS = 800
SPIKES_PER_UNIT = 4000
REPEAT = 2

if MODE == "old":
    table_mod._flatten_one_ragged_level = lambda data: list(itertools.chain.from_iterable(data))

table = DynamicTable(name="units", description="memray ragged flatten profile")
table.add_column(name="label", description="unit label")
for i in range(NUM_UNITS):
    table.add_row(label=f"u{i}")

for rep in range(REPEAT):
    rng = np.random.default_rng(rep)
    rows = []
    for _ in range(NUM_UNITS):
        n = int(rng.poisson(SPIKES_PER_UNIT))
        arr = np.sort(rng.random(max(n, 1), dtype=np.float64) * 1000.0)
        rows.append(arr)
    ragged = np.array(rows, dtype=object)
    table.add_column(
        name=f"spike_times_{rep}",
        description="ragged spike times",
        data=ragged,
        index=True,
    )

print(f"mode={MODE} rows={len(table)} columns={len(table.colnames)}")

# Peak resident set size from the OS (KB on Linux, bytes on macOS).
ru_maxrss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
peak_rss_bytes = ru_maxrss if os.uname().sysname == "Darwin" else ru_maxrss * 1024
peak_rss_mib = peak_rss_bytes / (1024 ** 2)

if psutil is not None:
    process = psutil.Process(os.getpid())
    current_rss_mib = process.memory_info().rss / (1024 ** 2)
    print(f"current_rss_mib={current_rss_mib:.2f} peak_rss_mib={peak_rss_mib:.2f}")
else:
    print(f"peak_rss_mib={peak_rss_mib:.2f} (psutil not installed; current RSS unavailable)")
```

</details>

Using the script above with 800 units and ~4,000 spikes each, the old path uses ~424 MiB RSS versus ~203 MiB with the new path (~2x reduction). Memray reports a larger gap: peak tracked memory drops from 361 MB to 148 MB (~2.4x) and total allocations from 1.05 GB to 387 MB (~2.7x), while allocation count stays roughly the same (~59k vs ~56k). 

I am not very familiar with a lot of the machinery I am touching here, so I am wary of unintended side effects, but I thought the best way to start the discussion would be with a PR.

## Checklist

### DIFF
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -21,15 +21,9 @@ from ..term_set import TermSetWrapper
 
 def _flatten_one_ragged_level(data):
     """Flatten one ragged nesting level with a fast path for numpy arrays."""
-    if len(data) == 0:
-        return []
-
     all_entries_are_numpy_arrays = all(isinstance(value, np.ndarray) for value in data)
     if all_entries_are_numpy_arrays:
-        try:
-            return np.concatenate(data)
-        except (TypeError, ValueError):
-            pass
+        return np.concatenate(data)
 
     return list(itertools.chain.from_iterable(data))
 
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -24,22 +24,12 @@ def _flatten_one_ragged_level(data):
     if len(data) == 0:
         return []
 
-    # Explicit NumPy fast path with readability-oriented checks.
     all_entries_are_numpy_arrays = all(isinstance(value, np.ndarray) for value in data)
     if all_entries_are_numpy_arrays:
-        first_array = data[0]
-        # Check whether all arrays have the same nonzero number of dimensions
-        first_array_has_nonzero_dimension = first_array.ndim > 0
-        all_arrays_have_same_dimension = all(value.ndim == first_array.ndim for value in data)
-        if first_array_has_nonzero_dimension and all_arrays_have_same_dimension:
-            all_arrays_have_same_trailing_shape = (
-                first_array.ndim == 1 or all(value.shape[1:] == first_array.shape[1:] for value in data)
-            )
-            if all_arrays_have_same_trailing_shape:
-                try:
-                    return np.concatenate(data)
-                except (TypeError, ValueError):
-                    pass
+        try:
+            return np.concatenate(data)
+        except (TypeError, ValueError):
+            pass
 
     return list(itertools.chain.from_iterable(data))
 
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -21,6 +21,9 @@ from ..term_set import TermSetWrapper
 
 def _flatten_one_ragged_level(data):
     """Flatten one ragged nesting level with a fast path for numpy arrays."""
+    if len(data) == 0:
+        return []
+
     all_entries_are_numpy_arrays = all(isinstance(value, np.ndarray) for value in data)
     if all_entries_are_numpy_arrays:
         return np.concatenate(data)
--- a/src/hdmf/common/table.py
+++ b/src/hdmf/common/table.py
@@ -28,9 +28,10 @@ def _flatten_one_ragged_level(data):
     all_entries_are_numpy_arrays = all(isinstance(value, np.ndarray) for value in data)
     if all_entries_are_numpy_arrays:
         first_array = data[0]
-        all_arrays_have_nonzero_dimension = first_array.ndim > 0
+        # Check whether all arrays have the same nonzero number of dimensions
+        first_array_has_nonzero_dimension = first_array.ndim > 0
         all_arrays_have_same_dimension = all(value.ndim == first_array.ndim for value in data)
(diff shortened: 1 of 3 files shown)


h-mayorquin:
Profile with IBL data:

I profiled peak RSS during units table writing on two sessions with large spike counts (1,366 and 2,185 units across 2 probes). The larger session has 3.3 GiB of per-spike ragged data on disk (spike times, amplitudes, and depths). Before these optimizations, it hit ~29.5 GB RSS and was OOM-killed on a 32 GiB machine. The `itertools.chain` flattening amplifies each ragged column ~4.5x due to Python object boxing (28 bytes per float instead of 8). Combined with neuroconv's column-first insertion strategy (`catalystneuro/neuroconv#1667`), peak RSS dropped to 13.8 GiB, a reduction of over 50%. On the smaller session (918 MiB on disk), I saw a similar 49% reduction (7,242 MiB to 4,597 MiB). With `np.concatenate`, each ragged column now costs exactly 1x its raw data size instead of 4.5x.

src/hdmf/common/table.py
```diff
@@ -19,6 +19,30 @@
...
+    if len(data) == 0:
+        return []
+
+    # Explicit NumPy fast path with readability-oriented checks.
+    all_entries_are_numpy_arrays = all(isinstance(value, np.ndarray) for value in data)
+    if all_entries_are_numpy_arrays:
+        first_array = data[0]
+        all_arrays_have_nonzero_dimension = first_array.ndim > 0
+        all_arrays_have_same_dimension = all(value.ndim == first_array.ndim for value in data)
+        if all_arrays_have_nonzero_dimension and all_arrays_have_same_dimension:
```
```suggestion
        # Check whether all arrays have the same nonzero number of dimensions
        first_array_has_nonzero_dimension = first_array.ndim > 0
        all_arrays_have_same_dimension = all(value.ndim == first_array.ndim for value in data)
        if first_array_has_nonzero_dimension and all_arrays_have_same_dimension:
```

rly:
@h-mayorquin This looks good and reasonable. Could you please add tests? Thanks.

src/hdmf/common/table.py
h-mayorquin: I applied this and added a regression test.

h-mayorquin: actually thinking in terms of the coverage patch here np.concatenate already does those checks here so I will be simplifying this.

src/hdmf/common/table.py
```diff
@@ -19,6 +19,15 @@
 from ..term_set import TermSetWrapper
 
 
+def _flatten_one_ragged_level(data):
```
rly, line 22: Looks good but the earlier version had an `if len(data) == 0: return []` guard that was dropped. In a rare edge case, if `data` is empty, `all()` returns True on an empty iterable, and `np.concatenate([])` raises a ValueError. This can happen if `index > 1` and the first flatten (in loop `for i in range(index):` produces an empty result (e.g., `data = [[], []]` flattens to `[]`, then the second iteration calls `_flatten_one_ragged_level([]))`. The old `itertools.chain.from_iterable([])` returned `[]` fine. It's unlikely that someone will add a column with data `[[], []]`, but I can see it happening, especially if more non-empty rows will be added. I think it's worth adding the empty guard back.  (this was found with help from Claude)

h-mayorquin: Yes, Ok, I added the guard and a regression test.

rly: Thanks!

rly (approved):

PR #1405, opened 2026-02-24
Fix `VectorData.extend()` corrupting 1D numpy arrays

h-mayorquin:
## Motivation

`extend_data` in `data_utils.py` uses `np.vstack` to extend numpy arrays, which silently reshapes 1D arrays into 2D matrices instead of concatenating them. For example, extending `[1.0, 2.0]` with `[3.0, 4.0]` produces `[[1.0, 2.0], [3.0, 4.0]]` instead of the expected `[1.0, 2.0, 3.0, 4.0]`. This PR replaces `np.vstack` with `np.concatenate`, which handles all dimensionalities correctly.

## How to test the behavior?
For example, building an electrode table with numpy-backed columns and extending it corrupts the data:

```python
from hdmf.common.table import DynamicTable, VectorData

locations = VectorData(name="location", description="brain region", data=np.array(["CA1", "CA1", "CA3", "CA3"]))
impedances = VectorData(name="impedance", description="impedance in ohms", data=np.array([1.2e6, 1.3e6, 1.1e6, 1.4e6]))

table = DynamicTable(name="electrodes", description="extracellular electrodes", columns=[locations, impedances])

# Add a second batch of electrodes
table["location"].extend(np.array(["VISp", "VISp", "VISl", "VISl"]))
table["impedance"].extend(np.array([1.5e6, 1.6e6, 1.7e6, 1.8e6]))

table["location"].data.shape  # Expected: (8,)  Actual: (2, 4)
table["location"].data
# Expected: ["CA1", "CA1", "CA3", "CA3", "VISp", "VISp", "VISl", "VISl"]
# Actual:   [["CA1", "CA1", "CA3", "CA3"],
#            ["VISp", "VISp", "VISl", "VISl"]]
```

The table goes from 8 electrodes to a 2x4 matrix. Each row now returns an array of four locations instead of a single string, breaking any downstream code that iterates over electrodes.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Added abstract methods `HDMFIO.load_namespaces` and `HDMFIO.load_namespaces_io`. @rly [#1299](https://github.com/hdmf-dev/hdmf/pull/1299)
 
 ### Fixed
+- Fixed `VectorData.extend()` silently corrupting 1D numpy arrays by reshaping them into 2D matrices. Replaced `np.vstack` with `np.concatenate` in `extend_data`. @h-mayorquin [#1405](https://github.com/hdmf-dev/hdmf/pull/1405)
 - Fixed a broken test and refactored `VectorIndex.get`. @rly, @mavaylon1 [#1293](https://github.com/hdmf-dev/hdmf/pull/1293)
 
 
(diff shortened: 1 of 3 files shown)


rly:
This looks good. Good catch @h-mayorquin! Thanks!

rly (approved):

PR #1407, opened 2026-02-25
Fix reverse class-to-namespace map overwrite when extensions include core

### LINKED ISSUE #1406: `register_container_type` overwrites reverse lookup, mapping shared types to wrong namespace
I have been getting `AttributeError: 'Proxy' object has no attribute 'read_io'` when displaying a read NWB file in my notebooks. The file was written with `ndx-events` and `ndx-hed` both loaded. `nwbfile.children` is empty, and `_repr_html_()` crashes when it tries to walk up the parent chain and hits an unresolved Proxy.

I think the problem is that PR #1372 changed `register_container_type` (manager.py:835) so that the reverse lookup (`class -> (namespace, data_type)`) uses direct assignment instead of `setdefault`. This means the last call to `register_container_type` for a given class wins, instead of the first. When any extension calls `include_namespace("core")`, `load_namespaces` re-registers every core type (NWBFile, TimeSeries, DynamicTable, etc.) under the extension namespace, overwriting the original `'core'` entry. Minimal reproducer:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "pynwb>=3.1.0",
#     "hdmf @ git+https://github.com/hdmf-dev/hdmf@dev",
# ]
# ///

import tempfile
from pathlib import Path

import pynwb
from pynwb import NWBFile
from pynwb.spec import NWBGroupSpec, NWBNamespaceBuilder, export_spec

tmp_dir = tempfile.mkdtemp()

# Create a minimal extension that includes core, like ndx-events or ndx-hed would
ns_builder = NWBNamespaceBuilder(name="ndx-events", version="0.1.0", doc="ndx-events")
ns_builder.include_namespace("core")
spec = NWBGroupSpec(
    neurodata_type_def="EventsTable",
    neurodata_type_inc="NWBContainer",
    doc="ndx-events type",
)
export_spec(ns_builder, [spec], tmp_dir)
pynwb.load_namespaces(str(Path(tmp_dir) / "ndx-events.namespace.yaml"))

# The bug: NWBFile is now mapped to 'ndx-events' instead of 'core'
ns, dt = pynwb.__TYPE_MAP.get_container_cls_dt(NWBFile)
assert ns == "core", f"BUG: NWBFile mapped to namespace='{ns}' instead of 'core'"
```

Coming back to my bug, I think this is caused because after the reverse map is contaminated, `build()` stamps `namespace='ndx-events'` on every core type's builder (including the root NWBFile). This wrong namespace gets written into the HDF5 file. When the file is read back the builder-side Proxy carries the namespace from HDF5 (`'ndx-events'`) while the container-side Proxy gets its namespace from the reverse map (which may now say `'ndx-hed'`, or `'core'`, depending on load order). `Proxy.__eq__` requires all four fields to match, so the namespace mismatch causes parent resolution to fail, which can end with the file getting the wrong namespace for a core data type. 

Maybe thi is is related to #1391 as well.

h-mayorquin:
Closes #1406.

The approach here is to guard the reverse map assignment in `register_container_type` with `if container_cls not in self.__container_cls_to_ns_dt`, restoring the first-registration-wins semantics that `setdefault` provided before PR #1372. This is safe because `NamespaceCatalog._order_deps` topologically sorts namespaces so base namespaces always load before extensions, meaning the first registration is always the defining one.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Added abstract methods `HDMFIO.load_namespaces` and `HDMFIO.load_namespaces_io`. @rly [#1299](https://github.com/hdmf-dev/hdmf/pull/1299)
 
 ### Fixed
+- Fixed `register_container_type` overwriting the reverse class-to-namespace map when an extension calls `include_namespace("core")`, which caused core types to be stamped with the extension's namespace. @h-mayorquin [#1407](https://github.com/hdmf-dev/hdmf/pull/1407)
 - Fixed a broken test and refactored `VectorIndex.get`. @rly, @mavaylon1 [#1293](https://github.com/hdmf-dev/hdmf/pull/1293)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@
 - Reduced memory allocations in `DynamicTable.add_column(..., index=True)` ragged column flattening by using `np.concatenate` instead of `list(itertools.chain.from_iterable(...))` when all entries are compatible numpy arrays. @h-mayorquin [#1403](https://github.com/hdmf-dev/hdmf/pull/1403)
 
 ### Fixed
-- Fixed `register_container_type` overwriting the reverse class-to-namespace map when an extension calls `include_namespace("core")`, which caused core types to be stamped with the extension's namespace. @h-mayorquin [#1407](https://github.com/hdmf-dev/hdmf/pull/1407)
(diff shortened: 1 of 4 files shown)


rly:
In researching this fix and the code around it, I (with Claude's help) caught additional issues - calling setattr on the namespace and type key should also be guarded, and the earlier pop from the reverse map should happen only if its entry actually belongs to the current (namespace, data_type). I've resolved these in the latest commit here

rly (approved):

h-mayorquin:
Thanks for the finishing this.

PR #1408, opened 2026-02-25
Make project dependencies uv compatible

h-mayorquin:
I have been working with uv lately and would like to make this package compatible with it. Currently `uv lock` and `uv run` fail because the resolver cannot build a single lockfile that satisfies all dependency groups and extras together: `test-min-deps` pins `numpy==1.22.0` while `hdmf-zarr` (via the `zarr` extra) requires `numpy>=1.24.0`, `test-min-deps` pins `jsonschema==3.2.0` while `linkml` (via the `termset` extra) requires `jsonschema>=4.0.0`, and pinned versions like `pandas==1.4.0` cannot even build on Python >=3.12. In short, the dependencies are locally consistent but not globally. 

To fix this, this PR:
- Raises the dependency lower bounds for `h5py`, `numpy`, and `pandas` to versions that are installable on Python >=3.12
- Adds `python_version == '3.10'` markers to all `test-min-deps` pins so the resolver does not attempt to use them on newer Python versions
- Declares `[tool.uv] conflicts` between `test-min-deps` and the optional extras (`all`, `zarr`, `termset`) so uv resolves them in separate forks.

The last addition is the only one that is uv-specific, the first two are standard [PEP 508](https://peps.python.org/pep-0508/)/[PEP 735](https://peps.python.org/pep-0735/) dependency specifications and should help any resolver while making the specification more precise.

If this make sense maybe it should go after the next release to avoid introducing risk into the current release cycle.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ coverage.xml
 # Version
 _version.py
 .python-version
+uv.lock
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Deprecated `BaseStorageSpec.add_attribute`, `GroupSpec.add_group`, `GroupSpec.add_dataset`, and `GroupSpec.add_link`. Use `set_attribute`, `set_group`, `set_dataset`, and `set_link` instead. @rly [#1333](https://github.com/hdmf-dev/hdmf/pull/1333)
 - Deprecated unused `BaseStorageSpec.get_data_type_spec` and `BaseStorageSpec.get_namespace_spec`. @rly [#1333](https://github.com/hdmf-dev/hdmf/pull/1333)
 - Moved `test`, `docs`, and `min-reqs` from `[project.optional-dependencies]` to `[dependency-groups]` (PEP 735). `min-reqs` was renamed to `test-min-deps`. @rly [#1395](https://github.com/hdmf-dev/hdmf/pull/1395)
+- Added Python version markers to dependency lower bounds (`h5py`, `numpy`, `pandas`) and `test-min-deps` pins, and declared `[tool.uv] conflicts` to make the project compatible with uv. @h-mayorquin [#1408](https://github.com/hdmf-dev/hdmf/pull/1408)
 - Changed `get_data_shape` to check `shape` before `maxshape`, so that objects with both attributes (e.g., h5py datasets) return their actual shape rather than their maximum shape. @rly [#1180](https://github.com/hdmf-dev/hdmf/pull/1180)
 
 ### Removed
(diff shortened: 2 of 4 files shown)


rly:
> Raises the dependency lower bounds for h5py, numpy, and pandas to versions that are installable on Python >=3.12

`pip` and `uv` should be able to handle this though, right?

I propose a slightly different solution. test-min-deps is used only in CI, and the dependencies are derived from the main dependencies - just pinned. `uv` is currently the only package manager that supports resolving dependencies with the lowest version pinned via their [`--resolution lowest` flag](https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy). Let's use `uv` in CI for this specific case and remove `test-min-deps` from `pyproject.toml`. `tox.ini` would then have:
```
install_command =
      minimum: uv pip install --resolution lowest {opts} {packages}
      upgraded: python -I -m pip install -U {opts} {packages}
      ...
```

What do you think @h-mayorquin ?

h-mayorquin:
> > Raises the dependency lower bounds for h5py, numpy, and pandas to versions that are installable on Python >=3.12
> 
> `pip` and `uv` should be able to handle this though, right?
> 
> I propose a slightly different solution. test-min-deps is used only in CI, and the dependencies are derived from the main dependencies - just pinned. `uv` is currently the only package manager that supports resolving dependencies with the lowest version pinned via their [`--resolution lowest` flag](https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy). Let's use `uv` in CI for this specific case and remove `test-min-deps` from `pyproject.toml`. `tox.ini` would then have:
> 
> ```
> install_command =
>       minimum: uv pip install --resolution lowest {opts} {packages}
>       upgraded: python -I -m pip install -U {opts} {packages}
>       ...
> ```
> 
> What do you think @h-mayorquin ?

I think this makes more sense. See the latest changes for my version of this to see if this is what you envisioned.

rly:
Yes, exactly! Can you update the ``tox.ini`` to get ``uv`` to work?

h-mayorquin:
> Yes, exactly! Can you update the `tox.ini` to get `uv` to work?

Ok, check the last commit.

rly:
Thanks. I was exploring a solution to the last failing issue as well. Since there are only two "minimum" environments, I worry that the !minimum negation approach (which is clever) is hard to understand and clutters the main `testenv` env definition. Let me push a small refactor.

rly:
It's a little redundant, but explicit and hopefully easier to understand for this use case. @h-mayorquin, let me know what you think.

h-mayorquin:
> It's a little redundant, but explicit and hopefully easier to understand for this use case. @h-mayorquin, let me know what you think.

I think this is better, more surgical. Restrict the special configuration to that environment might be more verbose but is more decoupled.

rly (approved):

PR #1414, opened 2026-03-03
Replace `hasattr(data, "__len__")` duck typing with `ndim`-based collection detection

h-mayorquin:
Motivated by hdmf-dev/hdmf-zarr#325 (zarr v2 to v3 migration).

hdmf uses `hasattr(data, "__len__")` across hte library ([`utils.py`](https://github.com/hdmf-dev/hdmf/blob/a27f6061/src/hdmf/utils.py#L852), [`h5tools.py`](https://github.com/hdmf-dev/hdmf/blob/a27f6061/src/hdmf/backends/hdf5/h5tools.py#L808), [`validator.py`](https://github.com/hdmf-dev/hdmf/blob/a27f6061/src/hdmf/validate/validator.py#L201)) to decide whether data is a collection or a scalar. This has worked so far because numpy implements `__len__` and other array libraries (h5py, zarr v2) followed that convention. But `__len__` is not part of any array specification (see the [array API standard object definition](https://data-apis.org/array-api/draft/API_specification/array_object.html#attributes) for the required attributes and methods), so we cannot rely on future array libraries continuing to provide it. The check is also implicit about what it is actually trying to detect. The zarr v2 to v3 migration (hdmf-dev/hdmf-zarr#325) exposed this, as zarr v3 arrays no longer implement `__len__`. The `__len__` heuristic also has a bug in the current release: 0-d ndarrays have `__len__` defined but `len()` raises `TypeError`, so passing one to `HDF5IO.get_type` or `write_dataset` crashes. This already happens when converting results from array-API-conforming libraries (which return 0-d arrays from reductions like `mean` and `sum`) to numpy via `np.asarray()`. A regression test for this is included.

This PR centralizes the "is this a collection?" logic into a single helper (`_is_collection`) that is both more general and more explicit: check `ndim > 0` first (which the array API standard guarantees for any conforming array), then fall back to `__len__` for plain Python containers. A companion `_get_length` helper safely gets first-dimension length via `shape[0]` or `len()`. This replaces all scattered checks with one clear, documented policy that covers both legacy types and array API standard conforming libraries, and shoudl eliminate the need for monkey-patching in hdmf-zarr.

The tests that are failing are because pynwb has not updated its ceiling on hdmf I think

Related to #1235
## Checklist


src/hdmf/validate/validator.py
```diff
@@ -198,7 +199,11 @@ def get_type(data, builder_dtype=None):
     # Numpy bool data
     elif isinstance(data, np.bool_):
         return 'bool', None
-    if not hasattr(data, '__len__'):
+    # Numpy 0-d structured array with compound dtype (ndim=0 but has named fields)
+    if isinstance(data, np.ndarray) and data.ndim == 0 and len(data.dtype) > 1:
```
```suggestion
    if isinstance(data, np.ndarray) and data.ndim == 0 and dtype.names is not None:
```
See the similar change here: https://github.com/hdmf-dev/hdmf/pull/1420/changes

h-mayorquin: I did this.

tests/unit/utils_test/test_utils.py
```diff
@@ -5,7 +5,109 @@
...
+        zarr v3 scalar indexing (e.g. zarr_array[0]) returns 0-d ndarrays.
+        The old hasattr(data, '__len__') check would incorrectly treat these
+        as collections, then crash on len(data).
+        """
+        zero_d = np.array(5.0)
+        self.assertTrue(hasattr(zero_d, '__len__'))  # confirms the old check would pass
+        with self.assertRaises(TypeError):
+            len(zero_d)  # confirms len() crashes
+        self.assertFalse(_is_collection(zero_d))  # our helper handles it correctly
+
```
rly: This is a nice demonstration of the change, but it is redundant with the
```python
    def test_numpy_scalar(self):
        self.assertFalse(_is_collection(np.float64(3.14)))
```
in the above class. It was good to document this change in the PR, but I suggest we remove this test.

(sorry, I thought I had submitted this comment in the last review)

h-mayorquin: I agree, removed it. I put effort on this test first and then the bot added the other one when I asked for edge cases and I did not catch that this made this redundant.

rly: Great, thanks

rly (approved):

PR #1415, opened 2026-03-03
Handle 0-d ndarrays in scalar isinstance checks

h-mayorquin:
Chained after #1414 (`_is_collection`). Motivated by hdmf-dev/hdmf-zarr#325 (zarr v2 to v3 migration).

hdmf's `get_type` functions infer element dtype by recursively indexing with `data[0]` until reaching a scalar, then calling `type()` on it. With numpy and zarr v2, `data[0]` on a 1-d float array returns a numpy scalar (e.g., `numpy.float64`), which passes `isinstance(val, float)` and has no `__len__`, so all downstream checks work. With zarr v3 (following the Python array API standard), `data[0]` returns a 0-d ndarray instead. A 0-d ndarray fails `isinstance(val, (int, float, str, bool))` and `type()` returns `numpy.ndarray` rather than the element dtype. PR #1414 fixed the crash path (`__len__` heuristic), but `isinstance` checks in other parts of the codebase still silently take the wrong branch when they encounter a 0-d ndarray.

This PR adds a `_unwrap_scalar` helper in `hdmf.utils` that converts 0-d ndarrays to numpy scalars via `.item()`, and applies it at the remaining `isinstance` checks that compare against Python scalar types. 

Together with #1414, this eliminates the need for the `__getitem__` monkey-patch in hdmf-zarr PR #325.

## Checklist


rly:
@h-mayorquin could you please resolve the conflicts in this PR? Thanks.

src/hdmf/container.py
```diff
@@ -976,7 +977,7 @@ def __bool__(self):
         return False
 
     def __len__(self):
-        return len(self.__data)
+        return _get_length(self.__data)
```
h-mayorquin, line 980: I am unsure why those were not merged with #1414

h-mayorquin: Ok, I think I only found those errors later and did some further additions here.

h-mayorquin: But maybe, we should have tests for those?

rly: Yeah, let's take this opportunity to add some. Those tests were missing in the first place.

h-mayorquin: Added I also improved the docstring and unify some uses len that were not going through the new centralized function. 

Let me know if you think there is something else to add

rly:
I haven't searched through where else `len` could be replaced, but these changes here look good to me. If you spot others, please feel free to open a new PR.

rly (approved):

rly:
Thanks @h-mayorquin !

h-mayorquin:
> I haven't searched through where else `len` could be replaced, but these changes here look good to me. If you spot others, please feel free to open a new PR.

Yes, I double checked. I will be on the look for more uses.

PR #1420, opened 2026-03-06
Fix writing compound dtype datasets with a single field

### LINKED ISSUE #1419: Writing HERD as a child container produces shape/maxshape rank mismatch

rly:
## Summary

- Fix compound dtype detection in `__list_fill__` to handle single-field compound dtypes (e.g., HERD's `keys` table with dtype `[('key', 'O')]`)
- The old check `len(dtype) > 1` missed single-field compound dtypes, causing `get_data_shape` to return a 2-D shape that mismatched the 1-D `maxshape`, producing an h5py error
- Use `dtype.names is not None` to correctly detect all compound dtypes

Fixes #1419

## Test plan

- [x] `test_write_dataset_list_single_field_compound_datatype` — unit test for writing a single-field compound dtype dataset
- [x] `test_write_herd_as_child` — integration test writing/reading HERD as a child group of a container (mirrors the NWB use case)
- [x] Existing compound dtype and HERD tests continue to pass

### DIFF
--- a/tests/unit/test_io_hdf5_h5tools.py
+++ b/tests/unit/test_io_hdf5_h5tools.py
@@ -208,13 +208,13 @@ class H5IOTest(TestCase):
         """Test that a compound dtype with a single field can be written."""
         a = np.array([('val1',), ('val2',)], dtype=[('key', 'O')])
         dset_builder = DatasetBuilder(
-                    name='test_dataset',
-                    data=a.tolist(),
-                    attributes={},
-                    dtype=[
-                        DtypeSpec('key', doc='key', dtype='text'),
-                    ],
-                )
+              name='test_dataset',
+              data=a.tolist(),
+              attributes={},
+              dtype=[
+                  DtypeSpec('key', doc='key', dtype='text'),
+              ],
+          )
         self.io.write_dataset(self.f, dset_builder)
         dset = self.f['test_dataset']
         self.assertEqual(dset.shape, (2,))
--- a/tests/unit/test_io_hdf5_h5tools.py
+++ b/tests/unit/test_io_hdf5_h5tools.py
@@ -23,6 +23,7 @@ from hdmf.backends.errors import UnsupportedOperation
 from hdmf.build import GroupBuilder, DatasetBuilder, BuildManager, TypeMap, OrphanContainerBuildError, LinkBuilder
 from hdmf.container import Container, HERDManager
 from hdmf import Data, docval
+from hdmf.common import SimpleMultiContainer, get_hdf5io
 from hdmf.data_utils import DataChunkIterator, GenericDataChunkIterator, InvalidDataIOError, append_data
 from hdmf.spec.catalog import SpecCatalog
 from hdmf.spec.namespace import NamespaceCatalog, SpecNamespace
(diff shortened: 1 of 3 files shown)


tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -204,6 +204,23 @@ def test_write_dataset_list_compound_datatype(self):
...
+        """Test that a compound dtype with a single field can be written."""
+        a = np.array([('val1',), ('val2',)], dtype=[('key', 'O')])
+        dset_builder = DatasetBuilder(
+                    name='test_dataset',
+                    data=a.tolist(),
+                    attributes={},
+                    dtype=[
+                        DtypeSpec('key', doc='key', dtype='text'),
+                    ],
+                )
```
```suggestion
              name='test_dataset',
              data=a.tolist(),
              attributes={},
              dtype=[
                  DtypeSpec('key', doc='key', dtype='text'),
              ],
          )
```

tests/unit/test_io_hdf5_h5tools.py
```diff
@@ -204,6 +204,23 @@ def test_write_dataset_list_compound_datatype(self):
...
+        """Test that a compound dtype with a single field can be written."""
+        a = np.array([('val1',), ('val2',)], dtype=[('key', 'O')])
+        dset_builder = DatasetBuilder(
+              name='test_dataset',
+              data=a.tolist(),
+              attributes={},
+              dtype=[
+                  DtypeSpec('key', doc='key', dtype='text'),
+              ],
+          )
```
```suggestion
        dset_builder = DatasetBuilder(
            name='test_dataset',
            data=a.tolist(),
            attributes={},
            dtype=[
                DtypeSpec('key', doc='key', dtype='text'),
            ],
        )
```

rly:
~~Zarr tests and optional req tests are expected to fail until https://github.com/NeurodataWithoutBorders/pynwb/issues/2171 is resolved.~~

Update: Zarr tests and optional req tests are expected to fail until a new version of PyNWB is released, but in the meantime, this can be tested locally.

oruebel (approved):

PR #1421, opened 2026-03-06
Fix #775: handle empty datasets in get_type with informative error

### LINKED ISSUE #775: [Bug]: Validation returns "IndexError: Index (0) out of range for empty dimension" for dataset with empty shape
### What happened?

See https://github.com/NeurodataWithoutBorders/pynwb/issues/1569

### Steps to Reproduce

```shell
See https://github.com/NeurodataWithoutBorders/pynwb/issues/1569
```

### Traceback

```shell
^
```

### Operating System

Linux

### Python Executable

Python

### Python Version

3.10

### Package Versions

^

### Code of Conduct

sejalpunwatkar:
This PR addresses issue #775 where validation of datasets with empty shapes caused an `IndexError`. 

Changes: 
- Updated `get_type` in `validator.py` to check for empty/null data before indexing.
- Replaced the generic `IndexError` with a more informative `EmptyArrayError`.
- Added a unit test in `tests/unit/test_validator.py` to verify the fix and prevent regressions.

Fixes #775


sejalpunwatkar:
Hi @rly , I've updated the PR to fix #775. I added the safety check for empty datasets in get_type and included a corresponding unit test.
Note on file changes: You will notice a large number of line changes (1018 insertions, 845 deletions) in the modified files. This is because I ran ruff format to resolve the linting failures in CI and align these specific files with the project's updated style standards. I also simplified the str and bytes checks in get_type to maintain the required complexity score (C901).
I see that some environment-specific tests are still failing; I am looking into those logs now and will provide an update shortly.

rly:
The optional and zarr tests are expected to fail until the next pynwb release, so those are OK. 

The ruff / style changes are significant and should be moved to a separate PR so that this one is easy to review and self-contained. However, I don't think ruff should be complaining about changing single quote to double quote if you are running it from the repo, which should use the configuration specified in `pyproject.toml`. 

Could you please limit this PR to just the changes relevant to the fix?

sejalpunwatkar:
@rly, I've cleaned up the PR. It now contains only the fix for #775 and the new test (16 lines total). I've avoided the bulk formatting changes to keep this PR self-contained. The Zarr and optional test failures are the ones you mentioned are expected to fail.

rly:
@sejalpunwatkar My apologies. I accidentally recommended that you fix a bug that no longer exists. After reviewing your changes, it looks like the base code already contained fixes for the originally reported IndexError on an empty HDF5 dataset from three years ago. I made a test script:
```python
import os
import tempfile

import h5py
from datetime import datetime
from dateutil.tz import tzutc
from pynwb import NWBFile, NWBHDF5IO, validate as pynwb_validate


"""Validate an NWB file with an empty vlen string dataset (related_publications)."""
fd, tmpfile = tempfile.mkstemp(suffix=".nwb")
os.close(fd)

# Create a normal NWB file with related_publications
nwbfile = NWBFile(
    session_description="test session",
    identifier="test_empty_pubs",
    session_start_time=datetime(2024, 1, 1, tzinfo=tzutc()),
    related_publications=("pub1",),
)
with NWBHDF5IO(tmpfile, "w") as io:
    io.write(nwbfile)

# Manually replace related_publications with an empty vlen string dataset
# to simulate the problematic file from issue #775
with h5py.File(tmpfile, "a") as f:
    del f["general/related_publications"]
    dt = h5py.special_dtype(vlen=str)
    f.create_dataset("general/related_publications", shape=(0,), dtype=dt)

# Verify the dataset is indeed empty
with h5py.File(tmpfile, "r") as f:
    ds = f["general/related_publications"]
    assert ds.shape == (0,), f"Expected shape (0,), got {ds.shape}"
    assert ds.dtype.metadata and ds.dtype.metadata.get("vlen") is str

# Validate — before the fix, this would crash with:
# IndexError: Index (0) out of range for empty dimension
with NWBHDF5IO(tmpfile, "r") as io:
    errors = pynwb_validate(io=io)

print(f"PASS: Validation completed without crash. {len(errors)} error(s) found.")
for e in errors:
    print(f"  {e}")

os.unlink(tmpfile)
```
and sure enough, that particular bug is not there. It's covered already by lines 160-164 in `_get_type_from_dtype_attr` which is called from `get_type` when a dataset has a "dtype" attribute.

You also added a None check but I believe it's not possible for `builder.data` to be None.

So I think this PR is not needed. Sorry for leading you astray!

sejalpunwatkar:
@rly No problem! Since the fix is already there, I’m happy to see the PR closed. I’ll go ahead and delete my local branch to keep things clean. Thanks for the clarification and the test script!

PR #1429, opened 2026-03-20
Add validation for dataset reference target types (#632)

### LINKED ISSUE #632: Support validation of dataset references
The validator currently does not validate dataset references. At a minimum it should validate that the reference points to builders with the right data_type.

Example: the following code should produce a validation error because the data_type of the builder within the reference does not match the target_type in the RefSpec for the dataset containing the reference. However, no error is generated because the reference data_type is not checked during validation.

```python
from hdmf.spec import DatasetSpec, SpecCatalog, SpecNamespace, RefSpec
from hdmf.build import DatasetBuilder, ReferenceBuilder

foo_spec = DatasetSpec(
    doc='a dataset containing a reference',
    dtype=RefSpec('NonexistantType', 'object'),
    data_type_def='Foo',
    shape=None
)
bar_spec = DatasetSpec(
    doc='a simple scalar dataset',
    data_type_def='Bar',
    dtype='int',
    shape=None
)
spec_catalog = SpecCatalog()
for spec in [foo_spec, bar_spec]:
namespace = SpecNamespace('a test namespace', 'test_namespace',
                                [{'source': 'test.yaml'}], version='0.1.0', catalog=spec_catalog)

bar = DatasetBuilder('bar', 12, attributes={'data_type': 'Bar'})
foo = DatasetBuilder('foo', ReferenceBuilder(bar), attributes={'data_type': 'Foo'})
result = vmap.validate(bar)
print(result)

```

I want to ensure the validator performs a complete and correct validation of builder objects.

On encountering a dataset reference type (possibly compound), the validator should verify that the data_type of the referenced builder matches the target_type defined in the spec.

**Describe alternatives you've considered**

**Additional context**
There could be additional things related to dataset references that are currently not being validated, so it would be good to evaluate if anything else should be included.

## Checklist

sejalpunwatkar:
## Motivation

The validator currently does not verify that dataset references point to builders with the correct `data_type` as defined by `RefSpec.target_type`.

This can lead to invalid references passing validation. This PR adds validation to ensure that referenced builders match the expected target type.

Fix #632

## How to test the behavior?
Run the validator tests:

pytest tests/unit/validator_tests -k reference

A new test demonstrates that a dataset reference with an incorrect `data_type` now raises a validation error.

## Checklist

### DIFF
--- a/src/hdmf/validate/validator.py
+++ b/src/hdmf/validate/validator.py
@@ -479,7 +479,6 @@ class DatasetValidator(BaseStorageValidator):
     @docval({'name': 'spec', 'type': DatasetSpec, 'doc': 'the specification to use to validate'},
             {'name': 'validator_map', 'type': ValidatorMap, 'doc': 'the ValidatorMap to use during validation'})
     def __init__(self, **kwargs):
-        self.validator_map = getargs('validator_map', kwargs)
         super().__init__(**kwargs)
 
     def _check_ref_target_type(self, val, expected_type, type_key, builder, ret):
@@ -489,10 +488,10 @@ class DatasetValidator(BaseStorageValidator):
             ref_type = target.attributes.get(type_key)
 
             if expected_type is not None and ref_type is not None:
-                hierarchy = self.validator_map.namespace.catalog.get_hierarchy(ref_type)
+                hierarchy = self.vmap.namespace.catalog.get_hierarchy(ref_type)
                 if expected_type not in hierarchy:
                     ret.append(
-                        DtypeError(
+                        IncorrectDataType(
                             self.get_spec_loc(self.spec),
                             f"{expected_type} (or subtype)",
                             ref_type,
@@ -532,7 +531,8 @@ class DatasetValidator(BaseStorageValidator):
                                 location=self.get_builder_loc(builder)
                             )
                         )
-                if isinstance(self.spec.dtype, RefSpec):
+
+                else:
                     if dtype != 'object':
                         ret.append(
                             DtypeError(
@@ -547,6 +547,7 @@ class DatasetValidator(BaseStorageValidator):
                     self._check_ref_target_type(data, expected_target, type_key, builder, ret)
 
             except EmptyArrayError:
+                # do not validate dtype of empty array. HDMF does not yet set dtype when writing a list/tuple
                 pass
 
         if isinstance(builder.dtype, list) and len(np.shape(builder.data)) == 0:
--- a/src/hdmf/validate/validator.py
+++ b/src/hdmf/validate/validator.py
@@ -429,6 +429,28 @@ class DatasetValidator(BaseStorageValidator):
                         expected = self.spec.dtype
                     ret.append(DtypeError(self.get_spec_loc(self.spec), expected, dtype,
                                           location=self.get_builder_loc(builder)))
+                if isinstance(self.spec.dtype, RefSpec) and check_type(self.spec.dtype, dtype, string_format):
+                    expected_type = self.spec.dtype.target_type
+
+                    def _check_ref(val):
+                        if isinstance(val, ReferenceBuilder):
+                            target = val.builder
+                            ref_type = target.attributes.get("data_type")
+
+                            if ref_type != expected_type:
+                                ret.append(
+                                    DtypeError(
+                                        self.get_spec_loc(self.spec),
+                                        expected_type,
+                                        ref_type,
+                                        location=self.get_builder_loc(builder)
+                                    )
+                                )
+                        elif isinstance(val, (list, tuple)):
+                            for v in val:
+                                _check_ref(v)
+
+                    _check_ref(data)
             except EmptyArrayError:
                 # do not validate dtype of empty array. HDMF does not yet set dtype when writing a list/tuple
                 pass
--- a/src/hdmf/validate/validator.py
+++ b/src/hdmf/validate/validator.py
@@ -514,15 +514,16 @@ class DatasetValidator(BaseStorageValidator):
                             target = val.builder
                             ref_type = target.attributes.get("data_type")
 
-                            if ref_type is not None and ref_type != expected_type:
-                                ret.append(
-                                    DtypeError(
-                                        self.get_spec_loc(self.spec),
-                                        expected_type,
-                                        ref_type,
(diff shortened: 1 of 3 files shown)


src/hdmf/validate/validator.py
```diff
@@ -506,6 +506,29 @@ def validate(self, **kwargs):
...
+                if isinstance(self.spec.dtype, RefSpec) and check_type(self.spec.dtype, dtype, string_format):
+                    expected_type = self.spec.dtype.target_type
+
+                    def _check_ref(val):
+                        if isinstance(val, ReferenceBuilder):
+                            target = val.builder
+                            ref_type = target.attributes.get("data_type")
+
+                            if expected_type is not None and ref_type is not None:
+                                if ref_type != expected_type:
```
rly: This equiality check should type hierarchy. This means if Bar is the expected type and SubBar, which inherits from Bar, is the actual type, this will incorrectly report an error. The validator correctly validates attribute references at `validator.py:411-414` with `catalog.get_hierarchy(data_type)`, and so this check should mirror that approach. Could you update this and also add a test for that?

src/hdmf/validate/validator.py
```diff
@@ -506,6 +506,29 @@ def validate(self, **kwargs):
                         expected = self.spec.dtype
                     ret.append(DtypeError(self.get_spec_loc(self.spec), expected, dtype,
                                           location=self.get_builder_loc(builder)))
+                if isinstance(self.spec.dtype, RefSpec) and check_type(self.spec.dtype, dtype, string_format):
+                    expected_type = self.spec.dtype.target_type
+
+                    def _check_ref(val):
+                        if isinstance(val, ReferenceBuilder):
+                            target = val.builder
+                            ref_type = target.attributes.get("data_type")
```
rly: Instead of `get("data_type")`, please use `get(target_spec.type_key())`. HDMF uses the key "data_type", but PyNWB uses the key "neurodata_type"

sejalpunwatkar:
Hi @rly , I've updated the reference validation in AttributeValidator to use get_hierarchy and type_key() as requested. I'm currently seeing some CI failures related to method complexity and tests, so I'll be refactoring this slightly into a helper method to keep Ruff happy and addressing the test failures shortly.

sejalpunwatkar:
@rly I've updated the reference validation to include hierarchy support and type_key() as you suggested. To address the Ruff complexity issues, I refactored the logic into a helper method, and all 1725 local unit tests are now passing successfully. Please take a look whenever your schedule allows!

src/hdmf/validate/validator.py
```diff
@@ -479,8 +479,30 @@ class DatasetValidator(BaseStorageValidator):
     @docval({'name': 'spec', 'type': DatasetSpec, 'doc': 'the specification to use to validate'},
             {'name': 'validator_map', 'type': ValidatorMap, 'doc': 'the ValidatorMap to use during validation'})
     def __init__(self, **kwargs):
+        self.validator_map = getargs('validator_map', kwargs)
```
rly: The base `Validator.__init__` already stores this as `self.__vmap` (accessible via `self.vmap`). We should use that instead.

src/hdmf/validate/validator.py
```diff
@@ -499,16 +521,34 @@ def validate(self, **kwargs):
...
+                                dtype,
+                                location=self.get_builder_loc(builder)
+                            )
+                        )
+                    expected_target = self.spec.dtype.target_type
+                    type_key = self.spec.type_key()
+                    self._check_ref_target_type(data, expected_target, type_key, builder, ret)
+
             except EmptyArrayError:
-                # do not validate dtype of empty array. HDMF does not yet set dtype when writing a list/tuple
```
rly, line 510: Was there a reason to remove this comment? I believe it documents a known limitation.

src/hdmf/validate/validator.py
```diff
@@ -479,8 +479,30 @@ class DatasetValidator(BaseStorageValidator):
...
+        """Helper to recursively validate reference target types and hierarchy."""
+        if isinstance(val, ReferenceBuilder):
+            target = val.builder
+            ref_type = target.attributes.get(type_key)
+
+            if expected_type is not None and ref_type is not None:
+                hierarchy = self.validator_map.namespace.catalog.get_hierarchy(ref_type)
+                if expected_type not in hierarchy:
+                    ret.append(
+                        DtypeError(
```
rly: We should use `IncorrectDataTypeError` instead of `DtypeError`. This is similar to the logic of `AttributeValidator` and is more appropriate here since the reference points to an object with the wrong data type.

src/hdmf/validate/validator.py
```diff
@@ -499,16 +521,34 @@ def validate(self, **kwargs):
...
-                                          location=self.get_builder_loc(builder)))
+                        ret.append(
+                            DtypeError(
+                                self.get_spec_loc(self.spec),
+                                expected,
+                                dtype,
+                                location=self.get_builder_loc(builder)
+                            )
+                        )
+                if isinstance(self.spec.dtype, RefSpec):
```
```suggestion
                else:
```

rly (commented):
Thanks @sejalpunwatkar ! I have a few suggested improvements

sejalpunwatkar:
@rly Thanks for the feedback! I've updated the PR to use self.vmap and IncorrectDataType as suggested. I also cleaned up the if/else logic, restored the missing comment, and updated the unit tests to match the new error class. All local tests are passing.

rly (dismissed):

tests/unit/validator_tests/test_validate.py
```diff
@@ -1907,3 +1998,48 @@ def test_isodatetime_no_time_component_fails(self):
                 # This confirms it fails because it lacks the 'T' and timezone
                 self.assertEqual(len(result), 1)
                 self.assertIsInstance(result[0], Error)
+
+    def test_dataset_reference_type_validation_hierarchy_success(self):
```
rly: This test is in `TestISODateTimeTimezone`. Please move this test to above the failure test above.

rly (approved):

rly:
This PR looks great! I just made a few minor changes. Thank you @sejalpunwatkar !

sejalpunwatkar:
Thanks for the help with the final polish and the merge, @rly

PR #1439, opened 2026-04-07
Make `expandable` a list; default expands only DynamicTable columns

### LINKED ISSUE #1437: Set expandable=True by default only for VectorData datasets
### Context

Sub-issue of #1436. See [discussion](https://github.com/hdmf-dev/hdmf/issues/1436#issuecomment-4202413189).

### Problem

Currently, `expandable=True` is set by default for **all** datasets in `H5DataIO`, which forces chunked storage even when it is unnecessary. For the vast majority of NWB files, datasets are written once and never appended to, so there is no need for them to be expandable.

Reviewing past user requests to append data ([pynwb#918](https://github.com/NeurodataWithoutBorders/pynwb/issues/918), [pynwb#1933](https://github.com/NeurodataWithoutBorders/pynwb/issues/1933), [pynwb#2055](https://github.com/NeurodataWithoutBorders/pynwb/issues/2055), [pynwb#2139](https://github.com/NeurodataWithoutBorders/pynwb/issues/2139)), these are all tables. There is no common use case for appending data to TimeSeries or other n-dimensional, non-table datasets.

### Proposed change

Change the default so that `expandable=True` is only set by default for `VectorData` datasets, not all datasets. Datasets with a known, fixed size should default to contiguous storage.

This is the higher-priority of the two changes proposed in #1436.

bendichter:
## Summary
- Change the default `expandable` behavior in `HDF5IO.write` / `write_builder` / `write_group` / `write_dataset` so that only `DynamicTable` columns and `id` are made expandable out of the box. Datasets of all other types now default to fixed-shape on-disk layout.
- **Breaking change:** `expandable` is now a list (or tuple) of data type names with default `("VectorData", "ElementIdentifiers")`, not a boolean. A dataset is made expandable if its data type (or an ancestor) is in the list — so subclasses of `VectorData` (e.g. `DynamicTableRegion`) and `ElementIdentifiers` are also covered. Pass an empty list/tuple to disable automatic expansion entirely. Passing `True`/`False` now raises a `TypeError` from docval.
- Subclass detection uses `BuildManager.is_sub_data_type`, with a `get_builder_dt` short-circuit so untyped dataset builders don't hit the hierarchy lookup.
- CHANGELOG entry added under a new `Breaking changes` heading with migration guidance.

Closes #1437

## Test plan
- [x] `TestDefaultExpandable` — `VectorData`, `VectorIndex`, indexed `VectorData`, and `ElementIdentifiers` are expandable by default
- [x] `TestDefaultExpandableSubclasses` — user-defined subclasses of `VectorData` / `ElementIdentifiers` are expandable by default (regression test for the `is_sub_data_type` fix)
- [x] `TestDefaultExpandableWithReferences` — reference-column dataset carries a reference dtype, is expandable, and refs resolve round-trip
- [x] `TestDefaultExpandableExplicitOverride` — explicit `H5DataIO(maxshape=...)` is preserved, not overridden
- [x] `TestExpandableArg` — `expandable=[]` disables expansion; custom list expands only listed types; `expandable=True`/`False` raises `TypeError`
- [x] `TestExpand.test_expand_set_shape` — `expandable=['QuxData']` still supports the append path on non-table types
- [x] Existing `TestDTRReferences` — `DynamicTableRegion` with references still works
- [x] Full test suite passes locally

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -311,7 +311,7 @@ class HDF5IO(HDMFIO):
              'doc': 'A HERD object to populate with references.',
              'default': None},
             {'name': 'expandable', 'type': bool, 'default': None,
-             'doc': ('If None (default), only VectorData (and subclasses) and ElementIdentifiers datasets '
+             'doc': ('If None (default), only VectorData, ElementIdentifiers, and their subclasses datasets '
                      'will be created as expandable. If True, all datasets will be created as expandable. '
                      'If False, no datasets will be automatically made expandable. Expandable datasets have '
                      'maxshape set based on the matching shape defined in the spec.')})
@@ -762,7 +762,7 @@ class HDF5IO(HDMFIO):
             {'name': 'export_source', 'type': str,
              'doc': 'The source of the builders when exporting', 'default': None},
             {'name': 'expandable', 'type': bool, 'default': None,
-             'doc': ('If None (default), only VectorData (and subclasses) and ElementIdentifiers datasets '
+             'doc': ('If None (default), only VectorData, ElementIdentifiers, and their subclasses datasets '
                      'will be created as expandable. If True, all datasets will be created as expandable. '
                      'If False, no datasets will be automatically made expandable. Expandable datasets have '
                      'maxshape set based on the matching shape defined in the spec.')})
@@ -944,7 +944,7 @@ class HDF5IO(HDMFIO):
             {'name': 'export_source', 'type': str,
              'doc': 'The source of the builders when exporting', 'default': None},
             {'name': 'expandable', 'type': bool, 'default': None,
-             'doc': ('If None (default), only VectorData (and subclasses) and ElementIdentifiers datasets '
+             'doc': ('If None (default), only VectorData, ElementIdentifiers, and their subclasses datasets '
                      'will be created as expandable. If True, all datasets will be created as expandable. '
                      'If False, no datasets will be automatically made expandable. Expandable datasets have '
                      'maxshape set based on the matching shape defined in the spec.')},
@@ -1049,7 +1049,7 @@ class HDF5IO(HDMFIO):
             {'name': 'export_source', 'type': str,
              'doc': 'The source of the builders when exporting', 'default': None},
             {'name': 'expandable', 'type': bool, 'default': None,
-             'doc': ('If None (default), only VectorData (and subclasses) and ElementIdentifiers datasets '
+             'doc': ('If None (default), only VectorData, ElementIdentifiers, and their subclasses datasets '
                      'will be created as expandable. If True, all datasets will be created as expandable. '
                      'If False, no datasets will be automatically made expandable. Expandable datasets have '
                      'maxshape set based on the matching shape defined in the spec.')},
@@ -1080,7 +1080,7 @@ class HDF5IO(HDMFIO):
             options['io_settings'] = {}
 
         # Set maxshape to make datasets expandable. When expandable is None (default), only
-        # VectorData (and subclasses) and ElementIdentifiers are made expandable — these are the
+        # VectorData, ElementIdentifiers, and their subclasses are made expandable — these are the
         # column and id datasets of DynamicTable, which users commonly need to append to. When
         # expandable is True, all datasets are made expandable. When False, none are.
         if (
@@ -1096,7 +1096,7 @@ class HDF5IO(HDMFIO):
                 if builder_dt is not None:
                     should_expand = (
                         self.manager.is_sub_data_type(builder, 'VectorData')
-                        or builder_dt == 'ElementIdentifiers'
+                        or self.manager.is_sub_data_type(builder, 'ElementIdentifiers')
                     )
             if should_expand:
                 options['io_settings']['maxshape'] = matched_spec_shape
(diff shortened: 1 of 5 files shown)


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1071,14 +1079,27 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
         ):
-            options['io_settings']['maxshape'] = matched_spec_shape
+            should_expand = expandable is True
+            if not should_expand:
+                # expandable is None (default): only expand VectorData and ElementIdentifiers
+                builder_dt = self.manager.get_builder_dt(builder)
+                if builder_dt is not None:
+                    should_expand = (
+                        self.manager.is_sub_data_type(builder, 'VectorData')
+                        or builder_dt == 'ElementIdentifiers'
```
```suggestion
                        or self.manager.is_sub_data_type(builder, 'ElementIdentifiers')
```
While there are no extensions of `ElementIdentifiers` that I am aware of, I think we should include that case here. I'll make the change

rly:
Looks good. I made some small changes:

## Summary

- Fixed `ElementIdentifiers` subclass detection in `HDF5IO.write_dataset`: replaced name equality (`builder_dt == 'ElementIdentifiers'`) with `self.manager.is_sub_data_type(builder, 'ElementIdentifiers')` so subclasses are also made expandable by default. Updated the matching docstring wording in the four `expandable` docval blocks.
- Added `CHANGELOG.md` entry under `Unreleased → Fixed` describing the full default-expandable behavior change.

### Tests added
- `TestDefaultExpandableSubclasses` (in `tests/unit/common/test_table.py`) — regression test for the `is_sub_data_type` fix. Uses `create_load_namespace_yaml` to define `CustomElementIdentifiers` and `CustomVectorData` subclasses, writes a `DynamicTable` using them, and asserts both `id` and `custom_col` get `maxshape=(None,)`.
- `TestExpandableFalse` — asserts `expandable=False` yields fixed shapes `(2,)` for both a VectorData column and the ElementIdentifiers `id`.
- `test_docval_none_default_with_non_none_type` (in `tests/unit/utils_test/test_docval.py`) — MWE confirming `docval` accepts `default=None` with `type=bool` (default preserved, wrong types still rejected), justifying the narrow `'type': bool` in the `expandable` docval.
- `test_roundtrip_basic_append` — now passes `expandable=True` explicitly and asserts `f['.../my_data'].maxshape == (None,)` to cover the force-all path on a non-VectorData dataset.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -759,9 +761,11 @@ def close_linked_files(self):
...
             {'name': 'export_source', 'type': str,
              'doc': 'The source of the builders when exporting', 'default': None},
-            {'name': 'expandable', 'type': bool, 'default': True,
-             'doc': ('If True (default), datasets will be created as expandable by setting the maxshape '
-                     'based on the matching shape defined in the spec.')})
+            {'name': 'expandable', 'type': bool, 'default': None,
+             'doc': ('If None (default), only datasets of type VectorData, ElementIdentifiers, or their '
+                     'subclasses will be created as expandable. If True, all datasets will be created as '
+                     'expandable. If False, no datasets will be automatically made expandable. Expandable '
+                     'datasets have maxshape set based on the matching shape defined in the spec.')})
```
oruebel: I think this definition for expandable may be confusing, because we are excluding certain types in the configuration and because there is no way to make `VectorData` not expandable here. To make this explicit we could have separate `dataset_expandable` (for regular datasets) and `table_expandable` to make columns of `DynamicTable` objects expandable (which includes `VectorData`, `ElementIdentifiers` and their subtypes. What do you think?

rly: `expandable=False` would make `VectorData` not expandable.

I don't like using two booleans when one of the combinations (dataset_expandable==True, table_expandable==False) isn't intuitive (would this mean all datasets except table types are expandable, or is this invalid? is table_expandable checked only if dataset_expandable is False?). 

Alternatively, what do you think about making 'expandable' be a list of data type names ("VectorData, "ElementIdentifiers") or boolean: True (all datasets), False (no datasets, same as empty list). Expands the scope of possible usages, but it is quite explicit.

oruebel: > Alternatively, what do you think about making 'expandable' be a list of data type names ("VectorData, "ElementIdentifiers") or boolean: True (all datasets), False (no datasets, same as empty list). Expands the scope of possible usages, but it is quite explicit.

Sure, that sounds reasonable.

bendichter: works for me

oruebel: [38d6698](https://github.com/hdmf-dev/hdmf/pull/1439/commits/38d6698890cd796ae36c4a3b8da5e922cb400443) looks good to me

oruebel:
Looks good to me. I'll let @rly  do final approval

rly:
I updated the changelog and moved the tests to a more appropriate location since expandable is no longer specific to tables. I also updated the PR summary so that the squash commit reflects the current state of changes. This is technically a breaking change and https://github.com/hdmf-dev/hdmf/pull/1440 will be a major change in behavior, so the next release will be 6.0.0.

rly (approved):

PR #1440, opened 2026-04-07
Set reasonable default chunk sizes (~4 MB) when chunks=True

### LINKED ISSUE #1438: Set more reasonable default chunk sizes (2-16 MB) when chunks=True
### Context

Sub-issue of #1436. See [discussion](https://github.com/hdmf-dev/hdmf/issues/1436#issuecomment-4202413189).

### Problem

When chunking is enabled (either explicitly via `chunks=True` or implicitly, e.g., when `max_shape` differs from the data shape), the default chunk sizes selected by h5py are too small — often well under 1 MB. Small chunks are particularly problematic for cloud access (e.g., files on DANDI), where each chunk may require a separate HTTP range request and per-request latency makes reading many small chunks far slower than reading fewer large chunks.

### Proposed change

When `chunks=True` is set (explicitly or implicitly), HDMF should select more reasonable default chunk sizes in the 2-16 MB range, rather than relying on h5py's defaults. This applies to all cases where chunking is active, including where it is implicit (e.g., `expandable=True` or `max_shape` set differently from data shape).

This is the lower-priority of the two changes proposed in #1436.

bendichter:
## Summary
- Add `_compute_chunk_shape()` static method that computes chunk shapes targeting ~4 MB per chunk, replacing h5py's default ~32 KB auto-chunking which is too small for cloud access (each chunk = one HTTP range request).
- Applied in `__list_fill__`, `__setup_chunked_dset__`, and `__setup_empty_dset__` whenever `chunks=True` (boolean, not a user-specified tuple).
- When `maxshape` forces chunking, explicitly sets `chunks=True` so the compute logic triggers instead of falling through to h5py's small defaults.
- The algorithm keeps all dimensions except the first at their full size and adjusts the first dimension to reach the target. Small datasets that fit within the target use their full shape.

Before (h5py defaults):
```
10000x100 float64: chunks=(63, 63),  chunk_bytes=31,752  (0.03 MB)
```

After:
```
10000x100 float64: chunks=(5242, 100), chunk_bytes=4,193,600 (4.00 MB)
```

Closes #1438

## Test plan
- [x] `TestComputeChunkShape` — unit tests for the chunk computation: target size, small datasets, empty datasets, 1D large, custom target
- [x] Full test suite passes (566 passed)

### DIFF
--- a/src/hdmf/backends/hdf5/h5tools.py
+++ b/src/hdmf/backends/hdf5/h5tools.py
@@ -1102,7 +1102,7 @@ class HDF5IO(HDMFIO):
             options['io_settings']['maxshape'] = matched_spec_shape
 
         # Ensure chunking is explicitly enabled when maxshape requires it, so that
-        # _compute_chunk_shape can replace it with appropriately-sized chunks later.
+        # compute_default_chunk_shape can replace it with appropriately-sized chunks later.
         if 'maxshape' in options['io_settings'] and 'chunks' not in options['io_settings']:
             options['io_settings']['chunks'] = True
 
@@ -1345,7 +1345,7 @@ class HDF5IO(HDMFIO):
                 io_settings['dtype'] = cls.__dtypes.get(io_settings['dtype'])
         # Replace chunks=True with computed chunk shape for better cloud access performance
         if io_settings.get('chunks') is True and 'shape' in io_settings and len(io_settings['shape']) > 0:
-            io_settings['chunks'] = HDF5IO._compute_chunk_shape(io_settings['shape'], io_settings.get('dtype'))
+            io_settings['chunks'] = HDF5IO.compute_default_chunk_shape(io_settings['shape'], io_settings.get('dtype'))
         try:
             dset = parent.create_dataset(name, **io_settings)
         except Exception as exc:
@@ -1377,7 +1377,7 @@ class HDF5IO(HDMFIO):
             io_settings['dtype'] = cls.__dtypes.get(io_settings['dtype'])
         # Replace chunks=True with computed chunk shape for better cloud access performance
         if io_settings.get('chunks') is True and 'shape' in io_settings and len(io_settings['shape']) > 0:
-            io_settings['chunks'] = HDF5IO._compute_chunk_shape(io_settings['shape'], io_settings.get('dtype'))
+            io_settings['chunks'] = HDF5IO.compute_default_chunk_shape(io_settings['shape'], io_settings.get('dtype'))
         try:
             dset = parent.create_dataset(name, **io_settings)
         except Exception as exc:
@@ -1431,7 +1431,7 @@ class HDF5IO(HDMFIO):
 
         # Replace chunks=True with computed chunk shape for better cloud access performance
         if io_settings.get('chunks') is True and data_shape is not None and len(data_shape) > 0:
-            io_settings['chunks'] = HDF5IO._compute_chunk_shape(data_shape, dtype)
+            io_settings['chunks'] = HDF5IO.compute_default_chunk_shape(data_shape, dtype)
 
         # Create the dataset
         try:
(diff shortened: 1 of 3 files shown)


src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1445,6 +1460,52 @@ def __get_ref(self, **kwargs):
     def _create_ref(self, **kwargs):
         return self.__get_ref(**kwargs)
 
+    @staticmethod
+    def _compute_chunk_shape(data_shape, dtype, target_chunk_bytes=4 * 1024 * 1024):
+        """Compute a chunk shape targeting a given number of bytes per chunk.
+
+        h5py's default auto-chunking targets ~32 KB, which is too small for cloud access where each
```
```suggestion
        h5py's default auto-chunking targets 8-500 KB chunks for datasets under 100 GB, depending on 
        dataset size. This is too small for cloud access where each
```

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1445,6 +1460,52 @@ def __get_ref(self, **kwargs):
...
+        """
+        try:
+            itemsize = np.dtype(dtype).itemsize
+        except TypeError:
+            return True  # fall back to h5py auto-chunking for unsupported dtypes
+
+        # Elements per "row" (all dimensions except the first)
+        elements_per_row = 1
+        for s in data_shape[1:]:
+            elements_per_row *= s
```
rly, line 1526: As mesoscale and volumetric imaging systems get better, `data_shape[1:]` will get bigger and result in chunks larger than 10 MB. e.g.,
- A `(1000, 4096, 4096) uint16` mesoscale imaging dataset would result in (1, 4096, 4096) = **32 MB chunks**. 
- Same for a `(1000, 1024, 1024, 16) uint16` volumetric imaging dataset. 

We don't typically see these dataset shapes now, but I expect we will in the next few years. I think it would be prudent to account for that, or simply raise a warning that auto chunk sizes are much larger than 4 MB and you should adjust the chunk size to your application.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1445,6 +1460,52 @@ def __get_ref(self, **kwargs):
...
+        bytes_per_row = elements_per_row * itemsize
+
+        if bytes_per_row == 0:
```
rly, line 1529: Can you add a test for this branch? It seems like a rare edge case.

src/hdmf/backends/hdf5/h5tools.py
```diff
@@ -1445,6 +1460,52 @@ def __get_ref(self, **kwargs):
     def _create_ref(self, **kwargs):
         return self.__get_ref(**kwargs)
 
+    @staticmethod
+    def _compute_chunk_shape(data_shape, dtype, target_chunk_bytes=4 * 1024 * 1024):
```
rly: I think this should be a public method, named something like `compute_default_chunk_shape`. Then, end users could use it to customize their chunk shapes for a different target, or inspect what chunk shape would be used if they set `chunks=True`

rly:
Looks good overall. I made a few suggestions. Could you also resolve the merge conflicts?

src/hdmf/backends/hdf5/h5tools.py
bendichter: OK that's a good point. We can think about splitting in other dimensions if a single slice in the first dimension exceeds our size target window

src/hdmf/backends/hdf5/h5tools.py
bendichter: good idea. Maybe it could accept a neurodata type, which would be unused by default but could be used by the someone overriding the function? We could add that arg if it's convenient

bendichter:
Addressed the review feedback in e62af031 (on top of a clean merge of `dev` in cbc353dd):

**API change**
- Renamed `_compute_chunk_shape` → `HDF5IO.compute_default_chunk_shape` so users can inspect or override the chunk shape that would be used when `chunks=True`.
- Added optional `neurodata_type` parameter (unused in the default implementation) as an extension hook for subclasses that want to specialize chunking per type — adding it now avoids a signature break later.

**Mesoscale / volumetric safety**
Previously, a `(1000, 4096, 4096) uint16` mesoscale frame would produce `(1, 4096, 4096) = 32 MB` chunks because the algorithm only adjusted the first dimension. The method now detects when a single first-axis slice already exceeds the target and iteratively halves the largest trailing axis until the chunk fits within the target (~4 MB by default), keeping results in the recommended 2-16 MB range regardless of frame size.

**Tests**
- `test_large_trailing_dims_get_split` — covers the mesoscale case; asserts `chunks[0] == 1` and chunk bytes ≤ 8 MB for `(1000, 4096, 4096) uint16`.
- `test_zero_trailing_dimension_falls_back` — exercises the `bytes_per_row == 0` branch that was flagged as uncovered; asserts it returns `True` so h5py's auto-chunking takes over.

**Other**
- Docstring wording updated to "8-500 KB chunks for datasets under 100 GB" per your suggestion.
- Added a CHANGELOG entry under Unreleased → Enhancements.
- Merge conflict with `dev` (in `test_io_hdf5_h5tools.py`) resolved by keeping both test classes.

rly (approved):

rly:
Thanks for the changes and looks good to me. @oruebel do you have any comments?

oruebel (approved):

PR #1442, opened 2026-04-14
Fix #624: Improve ExtraFieldError handling and validation behavior

### LINKED ISSUE #624: Create validation warnings on extra fields
These validation warnings were implemented in #542, but reverted in #548 due to the fact that it was raising warnings for pynwb types which were defined using a schema-language feature that the validator did not support. #585 adds support for this schema-language feature, so we can put back in the warnings, ideally for hdmf 3.0.

After #585, builders are sometimes evaluated against more than one spec object, which means that the implementation in #542 will no longer work. One possible solution is to take the intersection of extra fields from evaluating a builder against its specs, and then raise a warning for just those fields that are not defined in any of the specs.

Re-implementing the features from #542.

Create a validation warning for any field which is not defined in any of the specs that define a particular data type.

## Checklist

sejalpunwatkar:
Fixes #624
## Motivation

This PR improves validation handling for ExtraFieldError and reduces duplicate or incorrect validation errors.

Previously:
- Extra fields were sometimes incorrectly flagged (e.g., valid fields like "description")
- Duplicate validation errors were being returned
- Subtype validation and redundant attribute handling produced extra errors

This PR adjusts the validation logic to:
- Filter out valid/unintended extra fields
- Avoid duplicate error generation
- Improve compatibility with subtype validation

## How to test the behavior?
```
pytest tests/unit/validator_tests/test_validate.py
pytest tests/unit/common/test_table.py

Observe:
- Reduced duplicate validation errors
- Proper handling of ExtraFieldError
- Improved behavior for subtype and redundant attribute cases
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # HDMF Changelog
 
-## HDMF Unreleased
+## HDMF 5.1.0 (March 24, 2026)
 
 ### Enhancements
 - Made `HERDManager` an abstract interface (ABC) with an abstract `external_resources` property. Subclasses must now declare `external_resources` in their `__fields__` to satisfy the interface. Updated `__gather_fields` to allow auto-generated properties to override inherited abstract properties. Added `link_resources` and `get_external_resources` methods to `HERDManager` for managing a linked HERD separate from the primary one. @rly [#1431](https://github.com/hdmf-dev/hdmf/pull/1431)
--- a/src/hdmf/container.py
+++ b/src/hdmf/container.py
@@ -82,7 +82,7 @@ class HERDManager(ABC):
 
         Parameters
         ----------
-        linked : bool, optional
+        linked : bool
             If True, return the linked HERD set via ``link_resources``.
             If False (default), return the HERD set via ``__init__`` or the
             ``external_resources`` attribute.


sejalpunwatkar:
@rly I’ve been working on the extra field validation logic and wanted to clarify the intended behavior before refining further.

Currently, I’m iterating over matcher.unmatched_builders to raise ExtraFieldError, but I’m seeing cases where valid subtype datasets and some shape validation scenarios are getting flagged as extra.

Should ExtraFieldError be raised strictly for everything in unmatched_builders, or are there additional conditions I should consider (e.g., related to data_type or how matching is handled for subtypes)?

Also, I ran into some merge conflicts while iterating on this, which led to a large diff earlier, but I’ve since reverted to keep the changes minimal while I clarify the intended behavior.

I want to make sure I’m aligning with the expected validation behavior rather than over-filtering or under-reporting errors.

rly:
@sejalpunwatkar Thanks for starting on this! I have a few comments:
1. "valid subtype datasets and some shape validation scenarios are getting flagged as extra" - could you please give me some examples of this?
2. A warning should be generated, not an error. See https://github.com/hdmf-dev/hdmf/pull/542 for a starting point. This was reverted, as mentioned in #624. I hope it is not too much to re-apply the changes from #542 to the current state of the dev branch and address any new issues raised here and in pynwb. 
3. This branch currently removes some tests that were recently added to `main`.

sejalpunwatkar:
Thanks for the detailed feedback!

For the cases I mentioned, I’ll reproduce specific examples (e.g., subtype datasets and shape validation scenarios) and share them here clearly so we can confirm the expected behavior.

I also realize I was currently treating extra fields as errors, I’ll review #542 and #624 carefully and rework this to generate warnings instead, aligning with the intended design.

Regarding the tests, that wasn’t intentional, I’ll restore the removed tests and make sure the branch stays consistent with the latest main.

I’ll follow up shortly with concrete examples and an updated approach.

Issue #1448, opened 2026-04-26
Refactor: matched specs on builders (replace spec_ext plumbing)


rly:
# Epic: Matched Specs on Builders

## Problem

`ObjectMapper` instances are type-keyed singletons. `TypeMap.get_map()` (`src/hdmf/build/manager.py:793-818`) returns one mapper per neurodata type, and `ObjectMapper.spec` is the type's def-site spec from the namespace catalog (`src/hdmf/build/objectmapper.py:460-475`).

But spec overrides (dtype, shape, dims, value) live at the *position*, not the type. When `SubjectsTable(DynamicTable)` declares a `date_of_birth` column with `dtype='isodatetime'`, that override sits on the parent `SubjectsTable`'s subspec for that column, not on the `VectorData` def-site spec. The `VectorData` mapper's `self.spec` therefore knows nothing about the override.

Position information is plumbed two different ways today:

1. **Write path.** `__add_datasets` / `__add_groups` iterate the parent's subspecs and pass each as `spec_ext` to `build_manager.build()` (`src/hdmf/build/objectmapper.py:1318, 1343`). The child mapper merges def + override in `__check_dset_spec` (line 790).
2. **Read path.** PR #1313 adds the same `spec_ext` kwarg through `BuildManager.construct → TypeMap.construct → ObjectMapper.construct`, plus a `__parse_datetime_on_read` consumer in `ObjectMapper`.

Both rely on the parent's `__get_subspec_values` re-deriving the builder-to-subspec match by name and data_type (`src/hdmf/build/objectmapper.py:1431-1463, 1456`) and threading the result as a kwarg.

This works, but the architectural mismatch (type-keyed mapper, position-keyed override) leaks into every public `construct` and `build` signature, and each new read-time spec feature pays the same plumbing tax. Several call sites also currently consult `self.spec` directly without seeing overrides at all (see "Affected sites" below).

## Goal

Builders carry their own resolved (merged) spec. The mapper consults the builder's matched spec when present, falls back to `self.spec` otherwise. `spec_ext` kwargs disappear from public `construct` / `build` APIs, and any consumer that needs an override (dtype parsing, shape validation, string conversion, etc.) sees a single matched spec.

This is `Option 3` from the PR #1313 design discussion. It is a real refactor, not a minor change. It is worth doing because:

* It removes per-feature plumbing. Future read-time spec features (lazy decoders, dtype-driven validation, etc.) get the override for free.
[...]

PR #1447, opened 2026-04-26
Refactor spec_ext phase 1: add matched_spec slot to Builder

rly:
First phase of the resolved-specs-on-builders refactor. See epic issue #1448 for the full plan, design rationale, affected sites, risks, and downstream audits (pynwb, hdmf-zarr).

## Summary

Adds a write-once `matched_spec` property to the base `Builder` class (`src/hdmf/build/builders.py`). The slot defaults to `None`, accepts a one-time assignment, and raises `AttributeError` on overwrite, matching the existing `parent` and `source` setter discipline.

No callers are wired up yet. Subsequent phases (tracked in #1448):

- Phase 2: have `ObjectMapper.__get_subspec_values` set `sub_builder.matched_spec = subspec` after the match.
- Phase 3: migrate read consumers off the `spec_ext` kwarg and onto `builder.matched_spec`.
- Phase 4-5: same for the write path.
- Phase 6: drop `spec_ext` plumbing entirely.

This refactor supersedes the `spec_ext` plumbing introduced in #1313.

## Test plan

- [x] `pytest tests/unit/build_tests/test_builder.py` (4 new tests, all passing)
- [x] `pytest tests/unit/build_tests/` (no regressions)
- [x] CI green

## Notes for reviewers

Naming: `matched_spec` was chosen to pair with the existing `resolve_inc_spec` machinery in `src/hdmf/spec/spec.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1450, opened 2026-04-26
Refactor spec_ext phase 3: parse isodatetime/datetime values on read via matched_spec

rly:
Stacked on #1449 (Phase 2). Part of epic #1448.

This PR is the first read-time consumer of `builder.matched_spec`. It implements the inc-site dtype parsing case that motivated the epic, without introducing the `spec_ext` kwarg from #1313.

> Note: assumes #1313 is **not** merged. If #1313 lands first, Phase 3 becomes a migration (replace `spec_ext` reads with `matched_spec` reads, then drop the kwarg) instead of a greenfield consumer.

## Summary

- New module helper `_parse_isoformat(s)`: parses an ISO 8601 str/bytes into `datetime` (when a time component is present) or `date` (date-only). Idempotent for already-parsed values.
- New `ObjectMapper.__parse_datetime_on_read(value, spec)` static method: returns `value` unchanged unless the spec declares `dtype='isodatetime'` or `dtype='datetime'`. Handles scalars, lists, tuples, numpy arrays, h5py datasets, and zarr arrays via the numpy array protocol (`np.asarray(value).tolist()` covers v2 and v3, plus 0-d).
- Wired into three sites in `objectmapper.py`:
  - Attribute reads in `__get_subspec_values` (attribute specs are already position-resolved).
  - The inline DatasetSpec branch in `__get_subspec_values`, using `builder.matched_spec or spec`.
  - `ObjectMapper.construct` for `Data` containers, using `builder.matched_spec or self.spec`.

## What this delivers vs #1313

The motivating DynamicTable-column case (`SubjectsTable.date_of_birth: VectorData(dtype=isodatetime)`) now reads back as `date`/`datetime` values without any `spec_ext` plumbing. The position override flows through Phase 2's matcher writing `matched_spec`, then Phase 3's consumer honors it.

The orthogonal pieces of #1313 not covered here:
- `_isoformat` write helper (the write path already handles datetime → ISO via `__convert_string`, but Phase 5 will deduplicate that with the existing `__check_dset_spec` merge logic).
- `scalar_data` macro widening for datetime/date scalars in `get_class`-generated classes (independent of spec_ext; can ship as a small standalone PR).

## Test plan

- [x] `pytest tests/unit/build_tests/mapper_tests/test_read_datetime.py` (7 new tests, all passing): attribute datetime, attribute date-only, attribute non-datetime untouched, attribute already-parsed idempotent, def-site dtype on typed Data subclass, **inc-site dtype override** (the motivating case), and non-datetime dtype untouched.
- [x] `pytest tests/unit/` (1859 passed, no regressions)
- [x] CI green

## Notes for reviewers

- The base of this PR is `refactor/resolved-spec-read-matcher` (Phase 2). After #1447 and #1449 land, GitHub will rebase this onto `dev`.
- TODO comment on `__parse_datetime_on_read`: implement a lazy decoder for large datasets (currently eagerly materializes). Filed as a follow-up under "Out of scope" in the epic.

### DIFF
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -5,7 +5,6 @@ import warnings
 from collections import OrderedDict
 from copy import copy
 
-import h5py
 import numpy as np
 
 from .builders import DatasetBuilder, GroupBuilder, LinkBuilder, Builder, ReferenceBuilder, BaseBuilder
@@ -24,7 +23,7 @@ from ..utils import _is_collection, _get_length, _unwrap_scalar
 from ..query import ReferenceResolver
 from ..spec import Spec, AttributeSpec, DatasetSpec, GroupSpec, LinkSpec, RefSpec
 from ..spec.spec import BaseStorageSpec
-from ..utils import docval, getargs, ExtenderMeta, get_docval, get_data_shape, is_zarr_array, StrDataset
+from ..utils import docval, getargs, ExtenderMeta, get_docval, get_data_shape, is_array_like, is_zarr_array, StrDataset
 
 
 _const_arg = '__constructor_arg'
@@ -1396,7 +1395,7 @@ class ObjectMapper(metaclass=ExtenderMeta):
             elif isinstance(attr_val, ReferenceBuilder):
                 ret[attr_spec] = manager.construct(attr_val.builder)
             else:
-                ret[attr_spec] = self.__parse_datetime_on_read(attr_val, attr_spec)
+                ret[attr_spec] = self.__parse_if_datetime(attr_val, attr_spec)
         if isinstance(spec, GroupSpec):
             if not isinstance(builder, GroupBuilder):  # pragma: no cover
                 raise ValueError("__get_subspec_values - must pass GroupBuilder with GroupSpec")
@@ -1443,7 +1442,7 @@ class ObjectMapper(metaclass=ExtenderMeta):
             # matched_spec is set by the parent matcher when this builder was paired to a subspec
             # position; fall back to spec for top-level / direct-construct callers that bypass it.
             data_spec = builder.matched_spec or spec
-            ret[spec] = self.__parse_datetime_on_read(self.__check_ref_resolver(builder.data), data_spec)
+            ret[spec] = self.__parse_if_datetime(self.__check_ref_resolver(builder.data), data_spec)
         return ret
 
     @staticmethod
@@ -1456,14 +1455,16 @@ class ObjectMapper(metaclass=ExtenderMeta):
         return data
 
     @staticmethod
-    def __parse_datetime_on_read(value, spec):
+    def __parse_if_datetime(value, spec):
         """For specs with isodatetime/datetime dtype, parse stored ISO str/bytes back to datetime/date.
 
-        Handles scalars, lists/tuples, numpy arrays, and h5py datasets (eagerly materialized).
-        Idempotent: values whose elements are already parsed (or whose container type is
-        unknown, e.g., a DataIO wrapper) pass through unchanged.
+        Returns ``value`` unchanged when ``spec`` is not an attribute or dataset spec, or when its
+        dtype is not isodatetime/datetime — so callers can invoke this unconditionally on every
+        attribute/dataset value without first inspecting the dtype.
 
-        TODO: implement lazy decoder for large datasets.
+        Handles scalars, lists/tuples, numpy arrays, h5py datasets, and zarr arrays (eagerly
+        materialized via the numpy array protocol). Idempotent: values whose elements are already
+        parsed (or whose container type is unknown, e.g., a DataIO wrapper) pass through unchanged.
         """
         if not isinstance(spec, (AttributeSpec, DatasetSpec)):
             return value
@@ -1473,9 +1474,12 @@ class ObjectMapper(metaclass=ExtenderMeta):
             return _parse_isoformat(value)
         if isinstance(value, (list, tuple)):
             return type(value)(_parse_isoformat(v) for v in value)
-        if isinstance(value, np.ndarray) or isinstance(value, h5py.Dataset) or is_zarr_array(value):
+        if is_array_like(value):
             # Use the numpy array protocol so this works uniformly across numpy ndarrays,
             # h5py datasets, and zarr v2/v3 arrays without per-backend slicing.
+            # TODO: wrap the dataset to parse values on read instead of materializing the whole
+            # dataset in memory here. This becomes important for large isodatetime VectorData
+            # columns.
             materialized = np.asarray(value).tolist()
             if isinstance(materialized, list):
                 return [_parse_isoformat(v) for v in materialized]
@@ -1557,7 +1561,7 @@ class ObjectMapper(metaclass=ExtenderMeta):
             # matched_spec is set by the parent matcher when this builder was paired to a subspec
             # position; fall back to self.spec for top-level / direct-construct callers that bypass it.
             data_spec = builder.matched_spec or self.spec
-            const_args['data'] = self.__parse_datetime_on_read(
+            const_args['data'] = self.__parse_if_datetime(
                 self.__check_ref_resolver(builder.data), data_spec
             )
         for subspec, value in subspecs.items():
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -1,9 +1,11 @@
+import datetime
 import logging
 import re
 import warnings
 from collections import OrderedDict
 from copy import copy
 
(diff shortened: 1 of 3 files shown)


src/hdmf/build/objectmapper.py
```diff
@@ -94,6 +96,22 @@ def _ascii(s):
         raise ValueError("Expected unicode or ascii string, got %s" % type(s))
 
 
+def _parse_isoformat(s):
```
oruebel: - Add type hint for `s` since it can be either str or bytes or datetime
- Consider renaming `s` to something more meaningul, e.g., date_str
- Add parameter definition in the docstring

src/hdmf/build/objectmapper.py
```diff
@@ -1372,7 +1390,7 @@ def __get_subspec_values(self, builder, spec, manager):
             elif isinstance(attr_val, ReferenceBuilder):
                 ret[attr_spec] = manager.construct(attr_val.builder)
             else:
-                ret[attr_spec] = attr_val
+                ret[attr_spec] = self.__parse_datetime_on_read(attr_val, attr_spec)
```
oruebel: This may require an inline comment and/or renaming of  `self.__parse_datetime_on_read`. It is unclear here why I would always call to parse datetime here when in most cases the `attr_val` we encounter is actually not a datetime. In some sense, it seems like this function would handle any custom parsing of attributes, and datetime is the current main use-case where the function is taking actual action.

src/hdmf/build/objectmapper.py
oruebel, line 1389: Why do a Dataset need to use the matched_spec via `data_spec = builder.matched_spec or spec` but attributes do not?

src/hdmf/build/objectmapper.py
```diff
@@ -1428,6 +1449,34 @@ def __check_ref_resolver(data):
             return data.invert()
         return data
 
+    @staticmethod
+    def __parse_datetime_on_read(value, spec):
```
oruebel: How about renaming to `__parse_custom_dtype`? 

Also, the name mentions `_on_read`. What makes this function specific to read?

src/hdmf/build/objectmapper.py
rly: Addressed in 486a807c:

- Added type hint: `value: str | bytes | datetime.datetime | datetime.date`.
- Renamed `s` → `value`. I went with `value` rather than `date_str` because the parameter is intentionally polymorphic — non-str/bytes inputs (e.g., a `datetime` already produced by an earlier call) pass through unchanged so the helper is idempotent, so `date_str` would misrepresent that.
- Added `:param value:` and `:return:` sections to the docstring spelling out the bytes-decode, str-parse, and passthrough branches.

src/hdmf/build/objectmapper.py
```diff
@@ -1428,6 +1449,34 @@ def __check_ref_resolver(data):
...
+        """
+        if not isinstance(spec, (AttributeSpec, DatasetSpec)):
+            return value
+        if getattr(spec, 'dtype', None) not in ('isodatetime', 'datetime'):
+            return value
+        if isinstance(value, (str, bytes)):
+            return _parse_isoformat(value)
+        if isinstance(value, (list, tuple)):
+            return type(value)(_parse_isoformat(v) for v in value)
+        if isinstance(value, np.ndarray) or isinstance(value, h5py.Dataset) or is_zarr_array(value):
```
oruebel: This if condition creeps backend-specific logic into the ObjectMapper because it checks explicitly for h5py.Dataaset and zarr array, so this would need to be updated if one wanted to implement another backend. We should either have a generic `is_array` method in utils for this and ideally to not distribute this to deeply into HDMF, and ideally avoid checking for backend-specific types. Would it maybe be sufficient to do check for `shape` and then guard the np.asarray call?:


```suggestion
        if has_attr(value, 'shape')L
            try:
               np.asarray(value)
                   materialized = np.asarray(value).tolist() 
```

src/hdmf/build/objectmapper.py
```diff
@@ -1428,6 +1449,34 @@ def __check_ref_resolver(data):
...
+            # Use the numpy array protocol so this works uniformly across numpy ndarrays,
+            # h5py datasets, and zarr v2/v3 arrays without per-backend slicing.
+            materialized = np.asarray(value).tolist()
+            if isinstance(materialized, list):
+                return [_parse_isoformat(v) for v in materialized]
```
oruebel, line 1485: I would suggest adding a TODO item here. In the future, this would ideally be handled by wrapping the the dataset to parse values on read, rather than loading the whole dataset when opening the file. For now, this is not a big issue since the majority of cases are right now attributes, but if I created a VectorData of isodatetime this would likely cause performance issues.

src/hdmf/build/objectmapper.py
rly: Since this is a private method, the only use case we have right now is datetime, and the function itself is not simple, I think it's fine to keep the name `self.__parse_datetime` (yeah, it makes sense to drop `_on_read`). If we want to support other custom dtypes later, we can add `__parse_custom_dtype` later and have that call `__parse_datetime` and any other custom parser. What do you think?

oruebel (changes requested):
Looks good overall, just a few minor things that I think could use clarification

src/hdmf/build/objectmapper.py
rly: See my response to the comment below regarding the naming. 

> why I would always call to parse datetime here when in most cases the attr_val we encounter is actually not a datetime

This actually already happens a fair amount throughout the object mapper, e.g., `attr_value = self.__check_ref_resolver(attr_value)` checks if a dataset is a reference resolver and inverts it if so. `attr_val = self.__convert_string(attr_val, spec)` converts string types to the specified dtype and does nothing to non-string inputs. I agree, the names could be better. 

How about `__decode_datetime`?

src/hdmf/build/objectmapper.py
rly: We already have `is_zarr_array` used in `convert_dtype` because Zarr stores strings as objects. I agree, we should avoid checking for backend-specific types. All of numpy arrays, h5py datasets, and zarr arrays support the array API interface and have `hasattr(value, '__array__')` so I'll use that instead.

src/hdmf/build/objectmapper.py
oruebel: Sounds good

src/hdmf/build/objectmapper.py
rly: `hasattr(value, '__array__')` is pretty broad and might catch things that we don't want to catch though. What about in `utils.py`, add `is_array_like` that returns `isinstance(value, np.ndarray) or isinstance(value, h5py.Dataset) or is_zarr_array(value)`?

oruebel: > What about in `utils.py`, add `is_array_like` that returns `isinstance(value, np.ndarray) or isinstance(value, h5py.Dataset) or is_zarr_array(value)`?

I think that's fine. At least that puts the logic in a more central place

src/hdmf/build/objectmapper.py
rly: Hm. "decode" implies an encoding scheme like `bytes.decode()` which is not exactly this. How about just `__parse_if_datetime`? A little unconventional but clear.

src/hdmf/build/objectmapper.py
rly: Done

src/hdmf/build/objectmapper.py
rly: done

src/hdmf/build/objectmapper.py
oruebel: > How about just `__parse_if_datetime`? A little unconventional but clear.

That sounds like a good name for it

src/hdmf/build/objectmapper.py
rly: There is an edge case we don't currently exercise on the read path, where attribute dtypes (or shapes) are refined at the inc-site. The symmetric fix would be to pull `attr_spec` from `(builder.matched_spec or spec).attributes` rather than `spec.attributes`, matching what the dataset branch does. However, `self.__spec2attr` is keyed by the def-site AttributeSpec objects, so an inc-site-refined `AttributeSpec` has a different Python identity, the lookup misses, and the parsed value is silently dropped. Fixing this requires the `__spec2attr` identity-coupling fix tracked in #1462, which I've now updated to describe the read-side symptom alongside the existing write-side one.

oruebel: Thanks for the helpful clarification

oruebel (approved):

PR #1452, opened 2026-04-26
Refactor spec_ext phase 5: route write-time override via builder.resolved_spec

rly:
Stacked on #1451. Part of epic #1448.

## Summary

`ObjectMapper.build` now sets `builder.resolved_spec = spec_ext` as soon as the builder exists, and post-creation consumers read the override from the builder rather than the kwarg.

* **Group branch**: `resolved_spec` is set immediately after the builder is created (or used, when passed in).
* **Data branch**: `resolved_spec` is set after each of the three sub-paths (RefSpec, compound, regular dtype) has created the dataset builder.
* **Extra-attribute lookup** (line ~889): reads from `builder.resolved_spec` instead of `spec_ext`. The two carry the same value here, so behavior is unchanged.
* **`__add_containers`**: the redundant `__set_resolved_spec(new_builder, spec)` calls added in #1451 are removed. The inner `ObjectMapper.build` now sets the slot before returning.

## What this PR does *not* do

The epic originally framed Phase 5 as "drop the `spec_ext` kwarg." Looking at the code, that's not cleanly possible without further restructuring:

The Data branch needs the override info to compute `dtype` / `shape` / `dims` *before* the dataset builder is created (`__check_dset_spec` runs at the top of the branch, well before the builder exists). So the override has to flow in via a kwarg of some kind. After this PR, the kwarg stays, but its only role is feeding the pre-creation merge. Everything *after* builder creation reads `builder.resolved_spec`.

A future restructure could drop the kwarg by either (a) moving the dtype-merge step out of `ObjectMapper.build` so the caller pre-creates the builder, or (b) splitting `ObjectMapper.build` into a "compute build params" + "do the build" pair. Both are larger refactors than this phase warrants. I'll add a follow-up note to the epic if you want.

## Test plan

- [x] `pytest tests/unit/` (1864 passed, no regressions). The Phase 4 tests already lock in the contract that build-path-produced builders have `resolved_spec` set; this PR migrates *where* the set happens without changing observable behavior, so those tests continue to pass.
- [ ] CI green

## Notes for reviewers

- The base of this PR is `refactor/resolved-spec-write-matcher` (Phase 4). After #1447, #1449, #1450, and #1451 land, GitHub will rebase this onto `dev`.
- I removed an outdated TODO ("Check the BuildManager as refinements should probably be resolved rather than be passed in via spec_ext"). The remaining attribute-merge TODO about validation stays.

### DIFF
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -797,6 +797,8 @@ class ObjectMapper(metaclass=ExtenderMeta):
                               % (container.__class__.__name__, container.name, repr(source)))
             if builder is None:
                 builder = GroupBuilder(name, parent=parent, source=source)
+            if spec_ext is not None:
+                self.__set_resolved_spec(builder, spec_ext)
             self.__add_datasets(builder, self.__spec.datasets, container, manager, source)
             self.__add_groups(builder, self.__spec.groups, container, manager, source)
             self.__add_links(builder, self.__spec.links, container, manager, source)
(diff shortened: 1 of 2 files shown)


rly:
Folded into #1451. The remaining work after moving the ObjectMapper.build setter into Phase 4 was a single-line consumer migration plus a stale-comment cleanup, not enough to justify a separate PR.

PR #1453, opened 2026-04-26
Refactor spec_ext phase 5: docs and CHANGELOG cleanup

rly:
Stacked on #1451. Closes the cleanup phase of epic #1448.

## Summary

- **Docstrings**: `ObjectMapper.build`, `BuildManager.build`, and `TypeMap.build` now describe `spec_ext` precisely (it's the position-resolved subspec, used to compute dtype/shape before the dataset builder exists, and stored on the builder as `resolved_spec` post-creation), instead of the previous one-liner "a spec extension".
- **Misleading comment removed**: "use spec_dtype from self.spec when spec_ext does not specify dtype" in the Data branch. The actual behavior is the merge in `__check_dset_spec`, not a fallback.
- **CHANGELOG entry** under Unreleased / Enhancements describing the `resolved_spec` slot and the read-time `isodatetime` / `datetime` parsing introduced across the stack.

## What this PR does *not* do

The original Phase 5 task framed the kwarg drop as the main deliverable. After implementing Phases 1-4 it became clear that the kwarg drop is structurally a larger refactor:

- The Data branch in `ObjectMapper.build` needs the override info to compute `dtype` / `shape` *before* the dataset builder is created. `DatasetBuilder.__init__` takes `dtype` as a constructor arg, so the merge can't be deferred to post-creation without changing `DatasetBuilder`'s API.
- Realistic options for dropping the kwarg: (a) caller pre-creates the builder (relocates the merge problem to the caller); (b) out-of-band channel through `BuildManager` (relocates, doesn't really drop); (c) split `ObjectMapper.build` into "compute params" + "do the build" (real refactor, touches a lot of call sites).

Each is roughly the size of one of the existing phases. Left as future work; the doc and CHANGELOG side of the cleanup is closed here.

## Test plan

- [x] `pytest tests/unit/` (1864 passed, no regressions). This PR is doc/comment-only on the source side, so no new tests.
- [ ] CI green.

## Notes for reviewers

- The base of this PR is `refactor/resolved-spec-write-matcher` (Phase 4). After #1447, #1449, #1450, and #1451 land, GitHub will rebase this onto `dev`.
- The CHANGELOG entry references epic issue #1448 rather than any single PR, since the user-visible behavior is the result of the four-PR stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Enhancements
 - Set sensible default chunk sizes (~4 MB, in the recommended 2-16 MB range for cloud-hosted files) when `chunks=True`, replacing h5py's much smaller defaults. Added public `HDF5IO.compute_default_chunk_shape()` so users can inspect or override the chunk shape that would be used. @bendichter [#1440](https://github.com/hdmf-dev/hdmf/pull/1440)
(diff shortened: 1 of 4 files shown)


rly:
Folded into #1451. The remaining structural work (drop the  kwarg via build-split) is now tracked as a separate plan in epic #1448.

PR #1457, opened 2026-04-27
Fix building typed datasets with dtype isodatetime/datetime

rly:
Closes #1311. Stacked on the matched-spec helper-cleanups branch (#1455).

## Bug

Building a typed dataset whose inc-site declares `dtype: isodatetime` (e.g., a `VectorData` column refined as a date column) failed with:
```
Exception: could not resolve dtype for VectorData 'date_of_birth':
Expected unicode or ascii string, got <class 'datetime.date'>
```

The typed-dataset path (`__add_containers` → `build_manager.build` → inner mapper's `build` → `convert_dtype`) skips the `__convert_string` preprocessing that the untyped-dataset path applies. `convert_dtype` mapped `isodatetime` and `datetime` dtypes to `_ascii`, which only accepts `str` / `bytes`. So raw `datetime` / `date` values reached `_ascii(value)` and raised.

## Fix

- Added an `_isoformat(s)` module helper that converts `datetime` / `date` to ASCII-encoded ISO 8601 bytes (`s.isoformat().encode('ascii')`) and passes `str` / `bytes` through unchanged.
- Routed `isodatetime` and `datetime` in `ObjectMapper.__dtypes` to `_isoformat` instead of `_ascii`.
- Updated the four branches in `convert_dtype` that compared `spec_dtype_type is _ascii` to also accept `_isoformat`. Both produce ASCII output, so the `ret_dtype = 'ascii'` bookkeeping is identical.

The untyped-dataset path is unaffected: `__convert_string` still preprocesses datetime → ISO string before `convert_dtype` runs, and the new `_isoformat` accepts the resulting string just as `_ascii` did.

## Test plan

- [x] New `TestBuildTypedDatasetDatetime` class in `test_build_datetime.py` covers the regression from #1311 with three cases:
  - Typed column with `datetime` data and `dtype='isodatetime'`.
  - Typed column with `date` data and `dtype='isodatetime'`.
  - Typed column with `datetime` data and the `dtype='datetime'` alias.
- [x] Existing `TestBuildDatasetDateTime` tests (untyped-dataset path) continue to pass.
- [x] `pytest tests/unit/` (1863 passed, 3 new tests).
- [x] CI green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1459, opened 2026-04-27
Support scalar datetime in get_class and base Data

rly:
Stacked on #1457. Final piece of the matched_spec refactor stack (epic #1448, staging in #1458).

Ports the get_class / scalar-datetime support from #1313, on top of the matched_spec refactor. The architectural pieces of #1313 (spec_ext plumbing replacement, read-time datetime parsing) are already covered earlier in the stack; this PR covers the orthogonal pieces: scalar datetime acceptance in generated classes, the docval macro fix, and the test-helper adjustment.

## Summary

- **`hdmf.utils.scalar_data` macro** now includes `datetime.datetime` and `datetime.date`. This unblocks `Data(name=..., data=<datetime>)` for the base `Data` class (which uses the macro in its docval) and `get_class`-generated dtype-less Data subclasses.
- **`CustomClassGenerator.post_process` and `MCIClassGenerator.post_process`** take a new `type_map` argument (also threaded into the `DynamicTableMap` override). `CustomClassGenerator` adds an `_update_data_docval_arg` step that updates the inherited `data` docval arg's `type`, `doc`, and `shape` from the spec while preserving the parent class's `default`. So a refined `VectorData` subtype keeps `data=[]` as the default rather than becoming a required arg.
- **`process_field_spec`** reuses the already-computed `dtype` for the docval arg's `type` instead of recomputing via `_get_type`. Pure refactor.
- **`_assert_data_equal`** (test helper) skips the length comparison when both sides wrap scalar values; a `Data` container holding a single `datetime` has no `len()`.

## Tests

- Three new tests in `test_classgenerator.py`:
  - `test_get_class_include_scalar_datetime_attribute`
  - `test_get_class_include_scalar_datetime_dataset`
  - `test_get_class_dataset_preserves_parent_data_default`
- New `tests/unit/test_hdf5_roundtrip_extensions.py` with end-to-end roundtrip coverage for scalar and array datetime datasets and attributes through HDF5.
- Updated `test_docval.py::test_macro` for the widened tuple.
- Updated existing `post_process` test calls to pass `type_map`.

## Test plan

- [x] `pytest tests/unit/` (1868 passed, 5 new tests).
- [x] CI green.

## Notes

- No CHANGELOG entry on this PR — the consolidated entry lives on the staging branch (#1458).
- After this PR is merged into `fix/datetime-build` and the rest of the stack lands in staging, #1313 can be closed as superseded.

### DIFF
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -1,7 +1,6 @@
 from copy import deepcopy
 from datetime import datetime, date
 from collections.abc import Callable
-import warnings
 
 import numpy as np
 
@@ -316,22 +315,14 @@ class CustomClassGenerator:
     def _update_data_docval_arg(cls, docval_args, spec):
         """Update the inherited 'data' docval arg in place to reflect the dataset spec.
 
-        Updates only `type`, `doc`, and `shape` (when the spec declares one). Other keys
-        like `default` are preserved from the parent class's docval, so subclasses don't
-        accidentally lose, e.g., VectorData's empty-list default.
-        """
-        # fixed and default values on dataset specs are not yet applied to generated classes
-        if spec.value is not None:
-            warnings.warn(
-                f"Ignoring fixed value on dataset spec for type '{spec.data_type_def}' when generating class: "
-                f"fixed values are not yet applied to generated classes."
-            )
-        if spec.default_value is not None:
-            warnings.warn(
-                f"Ignoring default value on dataset spec for type '{spec.data_type_def}' when generating class: "
-                f"default values are not yet applied to generated classes."
-            )
+        Updates `type`, `doc`, `shape` (when the spec declares one), and `default` (when
+        the spec declares a `default_value`). Other keys are preserved from the parent
+        class's docval, so subclasses don't accidentally lose, e.g., VectorData's
+        empty-list default when the spec doesn't override it.
 
+        Fixed values (`value` on the spec) on dataset types are not yet applied to
+        generated classes.
+        """
         if spec.shape is None and spec.dims is None:
             if spec.dtype is not None:
                 dtype = cls._get_type_from_spec_dtype(spec.dtype)
@@ -346,10 +337,14 @@ class CustomClassGenerator:
             existing_data_arg['doc'] = spec.doc
             if spec.shape is not None:
                 existing_data_arg['shape'] = spec.shape
+            if spec.default_value is not None:
+                existing_data_arg['default'] = spec.default_value
         else:
             new_arg = dict(name='data', doc=spec.doc, type=dtype)
             if spec.shape is not None:
                 new_arg['shape'] = spec.shape
+            if spec.default_value is not None:
+                new_arg['default'] = spec.default_value
             docval_args.append(new_arg)
 
     @classmethod
--- a/src/hdmf/build/classgenerator.py
+++ b/src/hdmf/build/classgenerator.py
@@ -1,11 +1,12 @@
 from copy import deepcopy
 from datetime import datetime, date
 from collections.abc import Callable
+import warnings
 
 import numpy as np
 
 from ..container import Container, Data, MultiContainerInterface
-from ..spec import AttributeSpec, LinkSpec, RefSpec, GroupSpec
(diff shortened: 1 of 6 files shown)


src/hdmf/build/classgenerator.py
```diff
@@ -308,6 +309,49 @@ def post_process(cls, classdict, bases, docval_args, spec):
...
+        if isinstance(spec, DatasetSpec):
+            cls._update_data_docval_arg(docval_args, spec)
+
+    @classmethod
+    def _update_data_docval_arg(cls, docval_args, spec):
+        """Update the inherited 'data' docval arg in place to reflect the dataset spec.
+
+        Updates only `type`, `doc`, and `shape` (when the spec declares one). Other keys
+        like `default` are preserved from the parent class's docval, so subclasses don't
+        accidentally lose, e.g., VectorData's empty-list default.
```
oruebel: Please add doc for the function parameters.

src/hdmf/build/classgenerator.py
```diff
@@ -308,6 +309,49 @@ def post_process(cls, classdict, bases, docval_args, spec):
         # set default name in docval args if provided
         cls._set_default_name(docval_args, spec.default_name)
 
+        if isinstance(spec, DatasetSpec):
+            cls._update_data_docval_arg(docval_args, spec)
```
oruebel, line 312: Is `spec` in this case already the location-dependent `matched_spec`? If so, that would be useful to clarify in the docstring of the function or rename the parameter to clarify when the raw `spec` vs the `matched_spec` is being used.

src/hdmf/build/classgenerator.py
```diff
@@ -308,6 +309,49 @@ def post_process(cls, classdict, bases, docval_args, spec):
...
+        if spec.value is not None:
+            warnings.warn(
+                f"Ignoring fixed value on dataset spec for type '{spec.data_type_def}' when generating class: "
+                f"fixed values are not yet applied to generated classes."
+            )
+        if spec.default_value is not None:
+            warnings.warn(
+                f"Ignoring default value on dataset spec for type '{spec.data_type_def}' when generating class: "
+                f"default values are not yet applied to generated classes."
+            )
```
oruebel: Why are those not simply added to the `docval` as the default value?

Also, I'm wondering whether this will result in a lot of warnings for users. It seems like these warnings are mainly useful for developers of extensions.

src/hdmf/build/classgenerator.py
oruebel: Also, similar question, is `spec` the `matched_spec` here?

src/hdmf/build/classgenerator.py
rly: Good point - I added the default value to the docval. Fixed value is trickier but I cannot imagine the use case. I opened https://github.com/hdmf-dev/hdmf/issues/1466 for that.

src/hdmf/build/classgenerator.py
rly: This is the `spec` from the namespace catalog that defines the data type. For class generation, it only makes sense to use the `data_type_def` spec. Inc-site refinements (stored as `matched_spec` on builders) apply only during build/construct. For example, if `SubjectsTable` declares a column with `data_type_inc: VectorData` named "date_of_birth" and refines that slot to `dtype: isodatetime` with a  1D shape, those refinements can't be expressed on the generated `SubjectsTable.__init__`. The docval `type` for the "date_of_birth" arg is just `VectorData`, and docval does not support validating the constraint "must be a `VectorData` of isodatetime, shape (None,)". Instead, the refinements are stored in `builder.matched_spec` and get applied/checked during build and construct.

src/hdmf/build/classgenerator.py
rly: I added doc for the function parameters. `spec` is not the `matched_spec`. See my comment in the other thread.

src/hdmf/build/classgenerator.py
oruebel: Makes sense. Thanks for the clarification.

oruebel (approved):

PR #1476, opened 2026-05-15
Fix DatasetBuilder rejecting numpy scalar types with numpy 2.0

### LINKED ISSUE #1475: DatasetBuilder rejects numpy scalar types (numpy 2.0 removed __iter__ from scalars)
## Description

`DatasetBuilder.__init__` rejects numpy scalar values (e.g. `numpy.uint64`, `numpy.uint32`) passed as `data`, causing a `TypeError` when building containers with scalar-typed datasets.

## Root cause

The `scalar_data` docval macro contains only Python built-in types (`str`, `int`, `float`, `bytes`, `bool`, `datetime`, `date`). numpy scalars are not listed, and the `Iterable` fallback in the `DatasetBuilder.data` type list previously allowed them through because numpy 1.x scalars had `__iter__`. numpy 2.0 removed `__iter__` from scalars ([release notes](https://numpy.org/doc/stable/release/2.0.0-notes.html)), so `isinstance(np.uint64(5), Iterable)` now returns `False` and numpy scalars are rejected entirely.

## How it surfaces

`ObjectMapper.convert_dtype` returns numpy scalar types when converting scalar values — for example, a Python `int` value for a spec with `dtype: uint32` is converted to `numpy.uint32`, and an `int64` value for the same spec is upcasted to `numpy.uint64`. These numpy scalars are then passed as `data` to `DatasetBuilder.__init__`, which rejects them.

See downstream failure in pynwb: NeurodataWithoutBorders/pynwb#2193

## Fix

Add `numpy.generic` (the base class for all numpy scalar types) to the `scalar_data` docval macro in `src/hdmf/utils.py`.

rly:
## Summary

- Adds `numpy.generic` to the `scalar_data` docval macro so that numpy scalar types (e.g. `numpy.uint64`) are accepted as scalar dataset data values.
- Fixes compatibility with numpy 2.0, which removed `__iter__` from scalars, causing them to be silently rejected by `DatasetBuilder`.

Fixes #1475. See also NeurodataWithoutBorders/pynwb#2193.

## Test plan

- [x] New regression test in `TestDatasetBuilder.test_constructor_numpy_scalar_data` covers all common numpy scalar dtypes.
- [x] Updated `TestMacro.test_macro` to assert `numpy.generic` is in `scalar_data`.
- [x] Full test suite passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 6.0.2 (Unreleased)
+
+### Fixed
+- Added `numpy.generic` to the `scalar_data` docval macro so that numpy scalar types (e.g. `numpy.uint64`) are accepted as scalar data values. This fixes a compatibility issue with numpy 2.0, which removed `__iter__` from numpy scalars and caused them to be rejected by `DatasetBuilder`. @rly [#1474](https://github.com/hdmf-dev/hdmf/pull/1474)
+
 ## HDMF 6.0.1 (May 5, 2026)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## HDMF 6.0.2 (Unreleased)
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


rly:
Confirmed locally that this fixes https://github.com/NeurodataWithoutBorders/pynwb/issues/2193

oruebel (approved):
Looks good

Issue #1479, opened 2026-05-23
Refactor validator return type to support warnings as well as errors


rly:
## What would you like to see added to HDMF?

The HDMF validator currently returns a flat list of `Error` objects from `ValidatorMap.validate()`. Downstream applications such as PyNWB and NWB Inspector treat any non-empty return list as "this file is invalid".

This shape blocks #624 (and likely future severity levels such as info/hint). We cannot start emitting validation warnings, because mixing warning-level findings into the existing error list would silently flip previously-valid files to invalid the moment any warning fires. Routing warnings through Python's `warnings.warn()` machinery is also unattractive, since it breaks the validator's "collect and return, let the caller decide" model and forces downstream tools into `warnings.catch_warnings()` gymnastics just to render warnings alongside errors in the same report.

We need a way for the validator to return both errors and warnings as structured objects, without breaking the large body of existing downstream code that consumes `validate()`.

## What solution would you like?

Introduce a small `ValidationResult` wrapper that carries both lists and behaves like the existing errors list for the dominant usage patterns:

```python
class ValidationResult:
    def __init__(self, errors, warnings):
        self.errors = errors
        self.warnings = warnings
    def __iter__(self):       return iter(self.errors)
    def __len__(self):        return len(self.errors)
    def __bool__(self):       return bool(self.errors)
    def __getitem__(self, i): return self.errors[i]
```

Change `ValidatorMap.validate()` (and the per-type validators it delegates to) to return `ValidationResult` instead of a bare list. Add `ValidationWarning` as a sibling of `Error` (same `.name` / `.reason` / `.location` shape, not a `UserWarning` subclass) so warning objects can be rendered uniformly with errors.

Migration properties:
* `errors = vmap.validate(builder)` keeps working.
* `if errors:` is still True only when there are real errors. Warnings do not flip the verdict.
* `for e in errors:` and `len(errors)` and `errors[0]` still refer to errors only.
* New callers opt in to warnings explicitly via `result.warnings`.
* Only `isinstance(x, list)` breaks, which is rare.

No coordinated migration is required. PyNWB and NWB Inspector keep working as-is and can opt into surfacing warnings on their own schedule.

### Suggested PR breakdown

1. **Plumbing only, no new warnings yet.** Add `ValidationWarning`, add `ValidationResult`, change the validators to return `ValidationResult` with `warnings` always empty. Verify the existing validator test suite passes unchanged.
[...]

PR #1480, opened 2026-05-27
Refactor validator return type to support warnings via ValidationResult wrapper (#1479)

sejalpunwatkar:
## Motivation
Addresses #1479 

This PR implements Phase 1 (the foundational plumbing) to introduce `ValidationWarning` infrastructure without breaking the existing error reporting model. This change ensures that downstream packages (like PyNWB and NWB Inspector) can eventually consume warning-level messages without modifying their immediate logic or incorrectly marking files as invalid.

## How to test the behavior?
1. **Added Core Types**: Introduced `ValidationWarning` class as a sibling to `Error` inside `errors.py`, preserving the matching internal shape (`.name`, `.reason`, `.location`).
2. **Built the Smart Container**: Implemented the `ValidationResult` wrapper class in `errors.py` with custom Python overrides (`__bool__`, `__len__`, `__iter__`, `__getitem__`) pointing explicitly to the `errors` bucket to ensure backward compatibility.
3. **Wired the Plumbing**: Updated `ValidatorMap.validate()` inside `validator.py` to capture the `errors_list`, wrap it into the new `ValidationResult` container with an empty warnings container (`warnings=[]`), and adjusted its strict `@docval` return type check to `rtype=(list, ValidationResult)`.

## Checklist


src/hdmf/validate/errors.py
```diff
@@ -85,6 +85,78 @@ def __eq__(self, other):
...
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
+
+
+class ValidationResult:
+
+    def __init__(self, errors = None, warnings = None):
+        self.errors = list(errors) if errors is not None else []
+        self.warnings = list(warnings) if errors is not None else []
```
rly:
```suggestion
        self.warnings = list(warnings) if warnings is not None else []
```

src/hdmf/validate/errors.py
rly, line 18: Please add the new classes to `__all__`

src/hdmf/validate/validator.py
```diff
@@ -325,7 +270,7 @@ def get_validator(self, **kwargs):
             raise ValueError(msg)
 
     @docval({'name': 'builder', 'type': BaseBuilder, 'doc': 'the builder to validate'},
-            returns="a list of errors found", rtype=list)
+            returns="a list of errors found", rtype=(list, ValidationResult))
```
```suggestion
            returns="A ValidationResult containing the errors and warnings found", rtype=ValidationResult)
```

rly: Please also update the docstring of `validate(...)` in `src/hdmf/common/__init__.py:210`

rly:
Thanks! Looks good overall. Could you please add some basic tests of `ValidationResult`:
  - `validate()` returns a ValidationResult with `.warnings == []`.
  - Backward compatibility: `bool`, `len`, iteration, indexing all reflect errors only.
  - Construct `ValidationResult(warnings=[w])` and assert the warning survives.

Could you please also add a changelog entry that points to this PR?

src/hdmf/validate/errors.py
```diff
@@ -85,6 +85,78 @@ def __eq__(self, other):
...
+        return iter(self.errors)
+
+    def __len__(self):
+        return len(self.errors)
+
+    def __bool__(self):
+        return bool(self.errors)
+
+    def __getitem__(self, i):
+        return self.errors[i]
```
rly: Below this, could you add a simple `__repr__` so that users can print this object (like they did the list before) and get a human-readable listing of the errors and warnings instead of the object address?

rly:
@sejalpunwatkar Thanks for your work on this. Let me know when it is ready for re-review.

sejalpunwatkar:
Hi @rly, the PR is now ready for your re-review! 

I have fully addressed your feedback from the initial code review:
1. Bug Fix: Fixed the initialization logic typo in `ValidationResult.__init__` (`warnings` vs `errors`).
2. Ecosystem Safety: Added the new classes to `__all__` in `errors.py`.
3. Type-Checking & Docs: Tightened `@docval` in `validator.py` to strictly enforce returning `ValidationResult`, and updated the corresponding docstrings and imports in `src/hdmf/common/__init__.py`.
4. Enhanced Diagnostics: Added a human-readable `__repr__` method to `ValidationResult`.
5. Testing & Coverage: Added comprehensive unit tests at the bottom of `test_validate.py` covering the wrapper construction, magic methods, and empty state evaluations. 100% of the test suite is passing cleanly.
6. Changelog: Added a release tracking entry under the `## HDMF 5.0.0 (Upcoming)` section in `CHANGELOG.md`.

src/hdmf/common/__init__.py
```diff
@@ -207,10 +207,13 @@ def get_manager(**kwargs):
          'doc': 'the namespace to validate against', 'default': CORE_NAMESPACE},
         {'name': 'experimental', 'type': bool,
          'doc': 'data type is an experimental data type', 'default': False},
-        returns="errors in the file", rtype=list,
+        returns="errors and warnings in the file", rtype=ValidationResult,
```
rly:
```suggestion
        returns="A ValidationResult object containing the errors and warnings found.", rtype=ValidationResult,
```

src/hdmf/common/__init__.py
```diff
@@ -207,10 +207,13 @@ def get_manager(**kwargs):
...
         is_method=False)
 def validate(**kwargs):
-    """Validate an file against a namespace"""
+    """Validate an file against a namespace.
+    Returns:
+        ValidationResult: A ValidationResult object containing the errors and warnings found.
```
rly: It's unconventional, but the return docstring is encoded in the `returns` argument in the `@docval` above, so this should be removed here.
```suggestion
```

CHANGELOG.md
```diff
@@ -60,9 +60,11 @@
 - Deprecated `BaseStorageSpec.add_attribute`, `GroupSpec.add_group`, `GroupSpec.add_dataset`, and `GroupSpec.add_link`. Use `set_attribute`, `set_group`, `set_dataset`, and `set_link` instead. @rly [#1333](https://github.com/hdmf-dev/hdmf/pull/1333)
 - Deprecated unused `BaseStorageSpec.get_data_type_spec` and `BaseStorageSpec.get_namespace_spec`. @rly [#1333](https://github.com/hdmf-dev/hdmf/pull/1333)
 - Moved `test`, `docs`, and `min-reqs` from `[project.optional-dependencies]` to `[dependency-groups]` (PEP 735). `min-reqs` was renamed to `test-min-deps`. @rly [#1395](https://github.com/hdmf-dev/hdmf/pull/1395)
+- Refactored validator return type to `ValidationResult` to support upcoming validation warnings. @sejalpunwatkar [#1480](https://github.com/hdmf-dev/hdmf/pull/1480)
```
rly: The changelog has been updated since you started this work. Please add this entry to a new section at the top of the file
```
## HDMF 6.1.0 (Upcoming)
```

tests/unit/validator_tests/test_validate.py
```diff
@@ -1765,6 +1765,38 @@ def test_utf8_for_ascii_hdf5(self):
         self.assertIsInstance(results[0], DtypeError)
         self.assertEqual("Foo/data (my_foo/data): incorrect type - expected 'bytes', got 'utf'", str(results[0]))
 
+
```
rly: This test class was inserted in the middle of `TestVlenStringData` so the tests after your additions here have been added to `TestValidationResultWrapper` which is not correct.

tests/unit/validator_tests/test_validate.py
```diff
@@ -1765,6 +1765,38 @@ def test_utf8_for_ascii_hdf5(self):
...
+class TestValidationResultWrapper:
```
```suggestion
class TestValidationResultWrapper(TestCase):
```
By convention, all test classes should subclass `hdmf.testing.TestCase`

tests/unit/validator_tests/test_validate.py
```diff
@@ -1765,6 +1765,38 @@ def test_utf8_for_ascii_hdf5(self):
...
+        assert list(result) == [err]
+
+    def test_validation_result_empty_behavior(self):
+        empty_result = ValidationResult()
+        assert len(empty_result) == 0
+        assert bool(empty_result) is False
+        assert empty_result.warnings == []
+
+    def test_validate_method_returns_empty_warnings(self):
+        pass
```
rly: This should be implemented

tests/unit/validator_tests/test_validate.py
```diff
@@ -12,10 +12,10 @@
 from hdmf.testing import TestCase, remove_test_file
 from hdmf.validate import ValidatorMap
 from hdmf.validate.errors import (DtypeError, MissingError, ExpectedArrayError, MissingDataType,
-                                  IncorrectQuantityError, IllegalLinkError, ShapeError, IncorrectDataType)
+                                  IncorrectQuantityError, IllegalLinkError, ShapeError,IncorrectDataType,
```
```suggestion
                                  IncorrectQuantityError, IllegalLinkError, ShapeError, IncorrectDataType,
```

src/hdmf/validate/errors.py
```diff
@@ -85,6 +87,81 @@ def __eq__(self, other):
         return hash(self) == hash(other)
 
 
+class ValidationWarning:
```
rly: Since `ValidationWarning` is almost entirely identical to `Error`, it would be better to create a shared base class `ValidationIssue` that both extend from and that contains all the shared code. Downstream code will still be able to differentiate between `Error` and `ValidationWarning` because they are different classes.

One minor thing is that because `__eq__` is defined based on the hash and I believe the hash of a `ValidationWarning` and the hash of an identical `Error` object will be the same, then `Error("name", "reason") == ValidationWarning("name", "reason")` will be True. This might become an issue if/when errors and warnings are placed in a dict together. 

To address this, I would change `__eq__` to:
```python
    def __eq__(self, other):
          return type(self) is type(other) and hash(self) == hash(other)
```

rly:
Thanks @sejalpunwatkar . This is looking much better. I have a few more comments and suggestions above

sejalpunwatkar:
Hi @rly, I've addressed all your feedback: fixed the changelog section, updated the docstrings, corrected the test indentation, implemented the missing test, and updated the ValidationWarning.__eq__ check. Ready for review!

src/hdmf/validate/errors.py
rly: Thanks for the below change to `ValidationWarning.__eq__`. 

Since `ValidationWarning` and `Error` share significant code, could you also create a shared base class `ValidationIssue` that both extend from and that contains all the shared code? That would also make `Error.__eq__` match `ValidationWarning.__eq__`, which it should.

rly: With these changes, could you also add a quick test that an error and warning with the same name are not equal, so that these changed lines are tested?

src/hdmf/common/__init__.py
```diff
@@ -207,10 +207,11 @@ def get_manager(**kwargs):
          'doc': 'the namespace to validate against', 'default': CORE_NAMESPACE},
         {'name': 'experimental', 'type': bool,
          'doc': 'data type is an experimental data type', 'default': False},
-        returns="errors in the file", rtype=list,
+        returns="errors and warnings in the file", rtype=ValidationResult,
         is_method=False)
 def validate(**kwargs):
-    """Validate an file against a namespace"""
+    """Validate an file against a namespace."""
```
rly:
```suggestion
    """Validate a file against a namespace."""
```

sejalpunwatkar:
Thanks for catching that, @rly ! Fixed now.

rly:
Thanks for your patience and contributions @sejalpunwatkar ! I made a few minor adjustments and this is now good to go.

rly (approved):

sejalpunwatkar:
Thank you so much for the guidance and the thorough reviews throughout this process, @rly ! I’m incredibly glad to see this foundation land in the core framework. Looking forward to collaborating on the next stages of the validation warnings ecosystem!

Issue #1486, opened 2026-06-13
Epic: Read and write LinkML schema for HDMF (CSRMatrix / base.yaml)


rly:
## Background

HDMF defines data standards using a custom schema language (the HDMF Schema Language, or HDMFSL) developed roughly 10 years ago, when no equivalent tooling existed. That custom language and its supporting machinery (schema handling, validation, documentation) carry a real maintenance cost: they create a barrier to entry for new contributors, make it expensive to add schema features that users request, and isolate HDMF-based standards from the broader data-modeling and scientific data ecosystem.

Since then, [LinkML](https://linkml.io) has emerged as a modern, widely adopted data modeling language. It is human-readable and machine-processable, integrates with ontologies, supports validation of complex data structures, and interoperates with other schema languages (JSON Schema) and data formats (JSON, YAML, RDF, OWL, SQL). It is used across biomedical projects, such as the NCATS Biomedical Data Translator and the Alliance of Genome Resources, and it is being adopted by the DANDI Archive, BICAN, and openMINDS, with the Allen Institute for Neural Dynamics and OME-Zarr considering adoption. The LinkML Arrays Working Group has added a metamodel for labeled multi-dimensional arrays, which closes the main historical gap between LinkML and HDMF's array-centric modeling.

This epic is the first step in a multi-epic effort to add LinkML support to HDMF. This work is funded by an NIH National Library of Medicine grant [1R03LM014996-01](https://reporter.nih.gov/search/6QerHIl9F0qiYpmkxsU3DA/project-details/11240650) (PI: Ruebel) and is part of Aim 1 (enhancing interoperability of HDMF data models) of that grant.

This epic makes a small, self-contained type-set (`CSRMatrix` and the base types) fully bidirectional between HDMFSL and LinkML. Later epics extend the conventions, reader, and writer to the remaining HDMFSL constructs (links, references, compound dtypes, the `DynamicTable` family, inheritance roll-down) so the rest of the HDMF Common and NWB types are covered, and eventually adopt LinkML as the primary schema language.

## Goal

[...]

PR #1492, opened 2026-06-15
Add HDMF ↔ LinkML mapping conventions and the base.yaml / CSRMatrix fixtures

### LINKED ISSUE #1487: Define the initial HDMF → LinkML mapping conventions
Part of the ["Read and write LinkML schema for HDMF (CSRMatrix / base.yaml)" epic](https://github.com/hdmf-dev/hdmf/issues/1486)

Decide and document how a limited subset of HDMF Schema Language (HDMFSL) constructs map to LinkML. The deliverable is a **mapping-conventions document**; translating the schemas and building the reader are later issues that depend on this one. The conventions are defined against the specific constructs used by `base.yaml` and `CSRMatrix` (`sparse.yaml`).

The document defines a LinkML representation, with rationale, for:

- **`data_type_def` / `data_type_inc`** — defined types and inheritance, and how the reader tells a `GroupSpec` from a `DatasetSpec`.
- **The group / dataset / attribute distinction** — must survive in the LinkML so the reader can rebuild the right `Spec` subclass.
- **Object naming and identity** — fixed `name`, `default_name`, and unnamed types.
- **dtypes** — HDMFSL dtypes to LinkML ranges without losing precision, plus the no-dtype case.
- **Arrays (`dims` / `shape`)** — fixed, unspecified (`null`), and multiple-allowed shapes via the arrays metamodel; dimension naming.
- **`quantity`** — `1` / `?` / `*` / `+` to `required` / `multivalued`.
- **Namespace level** — how an HDMFSL `namespace.yaml` (its name, version, metadata, and list of schema files) maps to LinkML, including how the per-file schemas relate to the namespace. Cross-namespace imports are out of scope here.

## Acceptance criteria

- [ ] Mapping-conventions document committed to `docs/source/linkml-mapping.rst`, covering the constructs above with rationale and noting that this is not a complete mapping of all HDMFSL constructs.
- [ ] Discuss with the team how the conventions support faithful reconstruction of the `Spec` objects HDMF loads from the test namespace (`base.yaml` and `sparse.yaml`) via the namespace/catalog path.

### LINKED ISSUE #1488: Translate base.yaml and CSRMatrix to LinkML
Part of the ["Read and write LinkML schema for HDMF (CSRMatrix / base.yaml)" epic](https://github.com/hdmf-dev/hdmf/issues/1486). Depends on the [mapping-conventions issue](https://github.com/hdmf-dev/hdmf/issues/1487).

Hand-author the LinkML translation of a minimal test namespace: a `namespace.yaml` that references only `base.yaml` (`Data`, `Container`, `SimpleMultiContainer`) and `sparse.yaml` (`CSRMatrix`), plus those two schema files, following the agreed mapping conventions. These become the fixtures the reader and writer are tested against, so they must be correct and follow the conventions exactly. Include any companion files the conventions call for (for example, the dtype type definitions). The namespace references only these two files so the out-of-scope `table.yaml` and `resources.yaml` are excluded.

The HDMFSL sources are `src/hdmf/common/hdmf-common-schema/common/base.yaml` and `src/hdmf/common/hdmf-common-schema/common/sparse.yaml`; the test `namespace.yaml` is a new, minimal namespace modeled on `src/hdmf/common/hdmf-common-schema/common/namespace.yaml` but referencing only those two files.

The examples should be described in full in the new `docs/source/linkml-mapping.rst` file from #1487, as well as added as YAML to `tests/unit/linkml_tests/fixtures/`.

## Acceptance criteria

- [ ] LinkML files for the test namespace, the base types, and `CSRMatrix` committed to `tests/unit/linkml_tests/fixtures/`.
- [ ] The LinkML follows the agreed conventions and loads without error under `linkml-runtime`'s `SchemaView`.
- [ ] Every construct in the two HDMFSL schemas is represented; nothing the reader will need is dropped.
- [ ] `linkml-runtime` is added as an optional `linkml` dependency group; the core install does not require it.
- [ ] A unit test that shows that the translated LinkML schema is valid LinkML, using `linkml-runtime`'s `SchemaView`.

rly:
**Fixes #1487**
**Fixes #1488**

## Motivation

This adds the initial **HDMF ↔ LinkML mapping conventions** document together with the hand-authored LinkML translation of a minimal test namespace that applies them, the first two deliverables of the "Read and write LinkML schema for HDMF (CSRMatrix / base.yaml)" epic (#1486). Before any reader or writer is built, we need an agreed, written contract for how HDMF Schema Language (HDMFSL) constructs map to LinkML, and fixtures that prove the contract is real.

The two were originally separate issues. They are combined here because a conventions document that nothing has ever parsed is not a verifiable deliverable, and because running the conventions through `linkml-runtime` changed several of them. Fixing the doc before it merges is cheaper than amending it right after.

The scope is the constructs used by `base.yaml` (`Data`, `Container`, `SimpleMultiContainer`) and `sparse.yaml` (`CSRMatrix`), packaged as a minimal test namespace. What is out of scope is listed explicitly so the doc is not mistaken for a complete mapping.

## What is here

- `docs/source/linkml_mapping.rst`, added to the "Resources" toctree.
- LinkML fixtures in `tests/unit/linkml_tests/fixtures/`: `namespace.yaml`, `base.yaml`, `sparse.yaml`, and the companion `hdmf-linkml-types.yaml` defining the HDMFSL dtypes.
- A new `linkml` optional dependency group (`linkml-runtime`), wired into `all`, `tox.ini`, and the coverage workflow. The core install is unaffected and the tests skip when it is absent.
- `tests/unit/linkml_tests/test_fixtures.py`: the fixtures load under `SchemaView`, every range resolves, the dtype schema covers `DtypeHelper.valid_primary_dtypes`, and every field of the `Spec` objects HDMF loads natively from hdmf-common is present in the LinkML.

The coverage tests walk the natively loaded `Spec` objects rather than a hand-copied expectation, so a change to `base.yaml` or `sparse.yaml` that the fixtures do not follow fails here. The worked examples in the doc are `literalinclude`s of the fixture files, so the documented example cannot drift from what is tested.

## Conventions that changed once they were run through linkml-runtime

- The `name` identifier slot is declared once at each hierarchy root (`Data`, `Container`) and inherited via `is_a`. LinkML permits at most one identifier per class, so redeclaring it on every subclass was redundant.
- Every HDMFSL dtype string is its own named type, **synonyms included**. `Spec` stores whichever spelling the schema used, so collapsing `uint` onto `uint32` would rewrite the dtype on the way back and break the comparison. A synonym's `typeof` points at its primary, which records the synonym relationship in LinkML itself.
- `float`, `double`, `date`, and `datetime` are reused from `linkml:types` rather than redefined. They collide by name with LinkML built-ins whose semantics agree (`xsd:float` is 32 bit, `xsd:double` is 64 bit), and redefining them would shadow the built-ins for every schema in the import closure.
- `required` is always written explicitly on an attribute slot. HDMFSL defaults an attribute to required while LinkML defaults a slot to optional, so an omitted `required` would flip the meaning.
- `Spec` subclasses `dict` and inherits `dict` equality, so the comparison is over the keys actually present. The doc now states the two consequences for the reader: defaults stay absent (no `quantity: 1`, no explicit default `required`), and order within `attributes` / `datasets` / `groups` is preserved.
- Includes carry the HDMFSL `doc` as the slot `description`. The original example dropped it.
- Fixed `name` and `default_name` on a type map to `equals_string` and `ifabsent` on the identifier slot. Issue #1487 asked for these; neither is exercised by the types in scope, so both are stated as the convention.
- `AnyType` is called out as the one class range that does not mean a `data_type_inc`.

## How to test the behavior

```
pytest tests/unit/linkml_tests/
cd docs && sphinx-build -b html -W source _build/html
```

Verified: 14 tests pass (40 subtests), the full unit suite passes (1979 passed), the tests skip cleanly when `linkml-runtime` is not installed, ruff and codespell are clean, and the docs build with `-W` and 0 warnings.

## Checklist

- [x] Did you update `CHANGELOG.md`?
- [x] Have you reviewed the Contributing Guide?
- [x] Does the PR use "Fix #XXX" notation?

### DIFF
--- a/.github/workflows/run_coverage.yml
+++ b/.github/workflows/run_coverage.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Install package with optional dependencies
         if: ${{ matrix.opt_req }}
         run: |
-          python -m pip install --group test ".[tqdm,sparse,zarr,termset]"
+          python -m pip install --group test ".[tqdm,sparse,zarr,termset,linkml]"
 
       - name: Run tests and generate coverage report
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 6.0.3 (Upcoming)
+
+### Documentation
+- Added documentation of the conventions for mapping HDMF schema language constructs to LinkML, scoped to the `base.yaml` and `CSRMatrix` types. @rly [#1492](https://github.com/hdmf-dev/hdmf/pull/1492)
+
 ## HDMF 6.0.2 (May 15, 2026)
 
 ### Fixed
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -46,6 +46,7 @@ If you use HDMF in your research, please use the following citation:
    export
    validation
    spec_language_support
+   linkml_mapping
 
 .. toctree::
    :hidden:
--- a/docs/source/linkml_mapping.rst
+++ b/docs/source/linkml_mapping.rst
@@ -9,6 +9,12 @@ These conventions cover the constructs used by ``base.yaml`` (``Data``, ``Contai
 namespace. This is **not** a complete mapping of HDMFSL; constructs not yet covered are
 listed at the end.
 
+The LinkML files that apply these conventions to that test namespace live in
+``tests/unit/linkml_tests/fixtures/``. They are the reference the reader and writer are
+tested against, and ``tests/unit/linkml_tests/test_fixtures.py`` checks that they load
+under ``linkml-runtime``'s ``SchemaView`` and cover every field of the ``Spec`` objects HDMF
+loads natively from the HDMFSL sources.
+
 Goal and guiding principles
 ---------------------------
 
@@ -21,6 +27,21 @@ HDMF can read LinkML into its ``Spec`` objects (``GroupSpec``, ``DatasetSpec``,
    classes carry (``name``, ``doc``, ``dtype``, ``dims``, ``shape``, ``required``,
    ``quantity``, ``data_type_def`` / ``data_type_inc``, containment) must be recoverable
    from the LinkML.
+
+   ``Spec`` subclasses ``dict`` and inherits ``dict`` equality, so the comparison is over
+   the keys actually present. This puts two requirements on the reader:
+
+   - **Defaults stay absent.** HDMF omits a field left at its default: an
+     ``AttributeSpec`` with ``required`` at its default carries no ``required`` key, and a
+     dataset or subgroup at ``quantity`` ``1`` carries no ``quantity`` key. A reader that
+     writes the default value out explicitly produces a ``Spec`` that does not compare
+     equal.
+   - **Order within a construct list is preserved.** ``attributes``, ``datasets``, and
+     ``groups`` are lists, so their order is part of the comparison. LinkML holds all three
+     kinds in one ``attributes`` block on the class, so the reader partitions the slots by
+     ``spec_type`` and keeps their relative order within each partition. Declaring the
+     slots in HDMFSL order (attributes, then datasets, then groups) makes the LinkML read
+     the same way as the HDMFSL source.
 2. **Annotations carry HDMFSL provenance.** LinkML flattens groups, datasets, and
    attributes into classes and slots; HDMF needs to recover which was which. We record
    that with a small, explicit annotation vocabulary rather than inferring it.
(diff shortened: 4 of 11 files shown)


rly:
@jwbear This is an initial pass of a document that describes the mapping between HDMF Schema Language and LinkML. Please translate `sparse.yaml` from the HDMF Common namespace into LinkML following these conventions, and add the translated yaml into the doc. If anything is not clear, please ask, and then we should probably clarify the doc. Your fresh eyes looking at this doc will be invaluable is assessing whether it is clear and comprehensive. I suggest we tackle this together with pair programming this week after you are familiar with the task. We can coordinate over Slack.

jwbear:
@rly This is a nice reference, but in a CR isn't readily testable or verifiable. It might be helpful if you changed each of the actual requirements into a method stub or unit test that can be run to validate local criteria before a full compilation.

PR #1504, opened 2026-06-23
Fix HERD.get_object_entities(attribute=...) to mirror add_ref(attribute=...)

### LINKED ISSUE #1501: HERD.get_object_entities(attribute=...) is asymmetric with add_ref(attribute=...) and fails for non-DataType attributes
## What happened?

`HERD.add_ref(container=..., attribute=...)` and `HERD.get_object_entities(container=..., attribute=...)` resolve the `attribute` argument differently, so a reference added with an `attribute` cannot be read back with the same `attribute`.

- `add_ref` routes through `_validate_object` (`src/hdmf/common/resources.py:547-591`), which handles three cases. For a **non-DataType attribute** (a scalar attribute such as `DynamicTable.description`), it stores the *parent container* with a computed `relative_path` (e.g. `relative_path='description'`).
- `get_object_entities` (`src/hdmf/common/resources.py:883-887`) instead does `container[attribute]` and passes the default `relative_path=''`.

Because `container[attribute]` uses `__getitem__` (column/row lookup on `DynamicTable`), a scalar attribute name raises `KeyError`, and for containers that do not implement `__getitem__` it raises `TypeError`. The `attribute` path in `get_object_entities` therefore only works for the narrow case where `container[attribute]` happens to return the same object that was stored (e.g. a `DynamicTable` column).

## Steps to reproduce

```python
from hdmf.common.resources import HERD
from hdmf import Container, HERDManager

class FileC(Container, HERDManager):
    __fields__ = ({'name': 'external_resources', 'child': True, 'required_name': 'external_resources'},)

table = DynamicTable(name='table', description='table')
table.add_column(name='col1', description='column')
table.add_row(id=0, col1='data')
file = FileC(name='file')

er = HERD()
er.add_ref(file=file, container=table, attribute='description',
           key='key1', entity_id='entity_0', entity_uri='entity_0_uri')
print(er.objects.data[0])
# (0, '<uuid>', 'DynamicTable', 'description', '')   <- stored with relative_path='description'

er.get_object_entities(file=file, container=table, attribute='description')
# KeyError: 'description'

# Workaround that does work, showing the data is retrievable:
er.get_object_entities(file=file, container=table, relative_path='description')
# [{'entity_id': 'entity_0', 'entity_uri': 'entity_0_uri'}]
```

## Expected

`get_object_entities(attribute=...)` should resolve the object the same way `add_ref(attribute=...)` does (i.e. reuse the `_validate_object` logic, computing the same `relative_path`), so that a reference added with a given `attribute` can be retrieved with the same `attribute`. At minimum the two methods should be symmetric for every attribute case `add_ref` supports.

## Notes

- The existing `test_get_obj_entities_attribute` passes only because it uses a `DynamicTable` column (`attribute='col1'`), where both `getattr` and `__getitem__` resolve to the same column object (the DataType-attribute case). It does not cover the non-DataType attribute case exercised by `test_add_ref_nested`.

## Code of Conduct

- [x] I agree to follow this project's Code of Conduct

rly:
## Motivation

Fixes #1501.

`HERD.add_ref(container=..., attribute=...)` and `HERD.get_object_entities(container=..., attribute=...)` resolved the `attribute` argument differently, so a reference added with an `attribute` could not be read back with the same `attribute`:

- `add_ref` routes through `_validate_object`, which for a non-DataType attribute (e.g. `DynamicTable.description`) stores the parent container under a computed `relative_path`.
- `get_object_entities` instead did `container[attribute]` with the default `relative_path=''`, which raises `KeyError` (or `TypeError` for containers without `__getitem__`) for non-DataType attributes.

## Solution

`_validate_object` now takes a `create` flag and threads it to its `_check_object_field` calls. `get_object_entities` reuses `_validate_object(..., create=False)` for the attribute case, so both methods resolve the attribute identically.

The `attribute=None` path in `get_object_entities` is unchanged, so directly passing `relative_path` still works.

## How to test the behavior

Added `test_get_obj_entities_non_datatype_attribute`: adds a ref with `attribute='description'` and retrieves it with the same attribute. It fails on `dev` (`KeyError: 'description'`) and passes with this change. The existing `test_get_obj_entities_attribute` (DataType-attribute / column case) continues to pass.

## Checklist

- [x] Is your contribution compliant with our coding style? This will be checked during CI.
- [x] Have you included the relevant issue number using "Fix #XXX" notation where XXX is the issue number?

### DIFF
--- a/src/hdmf/common/resources.py
+++ b/src/hdmf/common/resources.py
@@ -352,15 +352,20 @@ class HERD(Container):
              'default': ''},
             {'name': 'field', 'type': str, 'default': '',
              'doc': ('The field of the compound data type using an external resource.')},
-            {'name': 'create', 'type': bool, 'default': True})
+            {'name': 'create', 'type': bool, 'default': False,
+             'doc': ('If the object is missing, create and return a dict that add_ref can use to add it to the '
+                     'tables. The created dict misses the idx key since the entry does not yet exist in HERD.')})
     def _check_object_field(self, **kwargs):
         """
         Check if a container, relative path, and field have been added.
 
-        The container can be either an object_id string or an AbstractContainer.
+        The following cases may occur:
+          1. If a single, corresponding object is found, then return that object from ``self.objects.row``.
+          2. If the container, relative_path, and field have not been added yet and ``create`` is True, then
+             return a new dict that add_ref can use to create the object. The returned dict misses the idx key
+             since the object has not been created yet.
 
-        If the container, relative_path, and field have not been added, add them
-        and return the corresponding Object. Otherwise, just return the Object.
+        :raises ValueError: If the object is missing and ``create`` is False, or if multiple matching objects exist.
         """
         file = kwargs['file']
         container = kwargs['container']
@@ -547,7 +552,24 @@ class HERD(Container):
         if len(missing_terms)>0:
             return {"missing_terms": missing_terms}
 
-    def _validate_object(self, container, attribute, field, file, create=True):
+    def _validate_object(self, container, attribute, field, file, create=False):
+        """
+        Resolve the object that an external reference is attached to and return it via ``_check_object_field``.
+
+        The ``attribute`` is resolved to the container and relative_path that identify the referenced object:
+          - ``attribute`` is None: the reference is on the container itself (relative_path '').
+          - ``attribute`` names a DataType (an AbstractContainer): the reference is on that sub-container.
+          - ``attribute`` names a non-DataType attribute (e.g. ``DynamicTable.description``): the reference is on the
+            nearest container ancestor of the attribute spec, with the relative_path computed from the spec path.
+
+        :param container: The Container/Data object that the reference is attached to.
+        :param attribute: The name of the attribute on the container, or None for the container itself.
+        :param field: The field of a compound data type using an external resource, or '' if not applicable.
+        :param file: The HERDManager file associated with the container.
+        :param create: Passed to ``_check_object_field``. If True and the object is missing, return a dict that
+                       add_ref can use to add it; if False, a missing object raises a ValueError.
+        :returns: The matching object from ``self.objects.row``, or (when ``create`` is True) a dict for add_ref.
+        """
         if attribute is None:  # Trivial Case
             relative_path = ''
             object_field = self._check_object_field(file=file,
@@ -698,7 +720,7 @@ class HERD(Container):
                        % (entity_uri, entity.entity_uri, entity_id))
                 warn(msg, stacklevel=3)
 
-        object_field = self._validate_object(container, attribute, field, file)
+        object_field = self._validate_object(container, attribute, field, file, create=True)
 
         #######################################
         # Validate Parameters and Populate HERD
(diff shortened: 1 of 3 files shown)


src/hdmf/common/resources.py
```diff
@@ -544,21 +544,23 @@ def add_ref_termset(self, **kwargs):
         if len(missing_terms)>0:
             return {"missing_terms": missing_terms}
 
-    def _validate_object(self, container, attribute, field, file):
+    def _validate_object(self, container, attribute, field, file, create=True):
```
oruebel: It would be useful to add a docstring to `_validate_object` to clarify what it does and what the different arguments mean.

src/hdmf/common/resources.py
```suggestion
            {'name': 'create', 'type': bool, 'default': True,
             'doc': ('If the fields are missing, create a dict that can be used to add it to the tables. 
                      The created dict will miss the idx key since the entry does not yet exist in HERD.')})
```

oruebel, line 837:
```suggestion
object_field = self._check_object_field(file=file,
                                        container=container,
                                        relative_path=relative_path,
                                        field=field,
                                        create=False)
```

The `get_key` is a retrieval function so we should not create entries here. Also, the subsequent for loop will fail if an `object_field` is being created by `self._check_object_field` because the loop accesses `object_field.idx` which is missing when `._check_object_field` creates a new dict.

oruebel, line 369:
```suggestion
def _check_object_field(self, **kwargs):
       """
       Check if a container, relative path, and field have been added.
       
       :returns: The following cases may occur:
         1. If a single, corresponding object is found then return that object 
            from `self.objects.rows`
         2. If the container, relative_path, and field have not been added yet
            and the create argument is set to True, then return a new dict that 
            add_ref can use to create the object. The returned dict 
            will miss the idx key since it has not been created yet. 
         :raises ValueError: A ValueError is raised if the object is missing OR if 
                             multiple matching objects exist.
       """
```

Removed `The container can be either an object_id string or an AbstractContainer.` because docval indicates that it must be an `AbstractContainter`.

Also `if the container, relative_path, and field have not been added, add them` is misleading as the function does not seem to actually add anything, but it only creates a dict that can be added.

src/hdmf/common/resources.py
oruebel: For both `_validate_object` and `_check_object_field`, I would change the default of the `create` argument to `false`. The `create=false` is the more common use-case for these functions to do look-ups. The `create=true` case looks like a special case and convenience for `add_ref`. As such, I think having `add_ref` set `create=True` explicitly and having all the other functions omit the create parameter by having `create=False` as default is more intuitive.

oruebel (changes requested):
I added some suggestions to help clarify the usage of `_check_object_field` and fix what I believe is a bug in `get_key`

src/hdmf/common/resources.py
rly: fixed in [2f2a319](https://github.com/hdmf-dev/hdmf/pull/1504/commits/2f2a3197daff63c2338d7b62bad8d6f8a75ddffc)

src/hdmf/common/resources.py
rly: fixed in [2f2a319](https://github.com/hdmf-dev/hdmf/pull/1504/commits/2f2a3197daff63c2338d7b62bad8d6f8a75ddffc)

rly: Good catch. fixed in [2f2a319](https://github.com/hdmf-dev/hdmf/pull/1504/commits/2f2a3197daff63c2338d7b62bad8d6f8a75ddffc)

rly: fixed in [2f2a319](https://github.com/hdmf-dev/hdmf/pull/1504/commits/2f2a3197daff63c2338d7b62bad8d6f8a75ddffc)

src/hdmf/common/resources.py
rly: I agree. fixed in [2f2a319](https://github.com/hdmf-dev/hdmf/pull/1504/commits/2f2a3197daff63c2338d7b62bad8d6f8a75ddffc)

oruebel (approved):

PR #1512, opened 2026-06-23
Remove the file argument from HERD.add_ref; resolve the file automatically

bendichter:
## Motivation

Follow-up to the HERD tutorial review in pynwb. From the CN+LBNL NWB sync (June 23) and the inline review comments on NeurodataWithoutBorders/pynwb#2200:

> I don't understand why it is necessary to provide the file arg here. Can't you resolve the file from the container directly?

— https://github.com/NeurodataWithoutBorders/pynwb/pull/2200#discussion_r3460642596

> we want to populate this automatically and error out when file not present

— https://github.com/NeurodataWithoutBorders/pynwb/pull/2200#discussion_r3461354051

Group consensus: a reference should only be addable to a container that is already in a file, the file should be resolved automatically from the container, and `add_ref` should error clearly when no file is available.

## Changes

**Breaking:**
- Removed the `file` argument from `HERD.add_ref`. The file is now always resolved from the container's parent hierarchy via `_get_file_from_container`.
- Removed the `file` argument from `HERD.add_ref_termset` as well — it was now purely vestigial (it only forwarded the file to `add_ref`, which no longer accepts it). `get_key`, `get_object_entities`, and `_check_object_field` still take `file` since they use it for lookups, not forwarding.

**Fix (carried from the first commit on this branch):**
- `_get_file_from_container` previously only raised when the container had *no parent at all*; a container whose parent chain contained no file silently returned `None`. It now walks the full ancestry and raises a clear, actionable error naming the container in both missing-file cases.

**Docs:**
- Updated the `plot_external_resources` gallery tutorial to the new API and revised the surrounding prose ("How add_ref resolves the file").

**Tests:**
- Updated the HERD unit tests and HERD IO tests to parent each container to its file before calling `add_ref`. Where a single container was previously associated with multiple anonymous files via the old `file=` arg (no longer possible, since a container has one parent), the tests now use a single shared file or distinct containers as appropriate.

## Verification

- `tests/unit/common/test_resources.py`: 84 tests pass
- `tests/unit/back_compat_tests/test_herd_1_8_0.py`: passes
- `tests/unit/test_io_hdf5_h5tools.py::TestHERDIO`: passes
- The `plot_external_resources` gallery script runs end-to-end
- `ruff` clean

Note: `src/hdmf/testing/make_test_files.py` intentionally still uses `add_ref(file=...)` — it is pinned to run under `hdmf==4.3.1` to regenerate legacy back-compat files, so it must keep the old API.


src/hdmf/common/resources.py
```diff
@@ -405,20 +405,21 @@ def _get_file_from_container(self, **kwargs):
...
+        while parent is not None:
+            if isinstance(parent, HERDManager):
+                return parent
+            parent = parent.parent
+
+        # No file was found in the container's ancestry. This happens when the
+        # container has not been added to a file, either because it has no parent
+        # or because none of its ancestors is a file.
+        msg = ("Could not find the file associated with container '%s'. Please add the container "
+               "to the file before adding an external reference." % getattr(container, 'name', container))
```
oruebel: This is really just a very nit-picky comment to avoid possible confusion, rather than suggesting a specific change. 

The code here is looking for the closest parent that is a `HERDManager`. In the case of NWB this is always the root-level file (and probably in most real use cases). However, in the more general cases, this may not always be `the file associated with container`, but whatever the closest parent is that inherits from `HERDManager`. I.e., I think it is possible to define a schema where a file would look like this:

```
 /root
    |-- group1 <HERDManager>
    |        |-- dataset1
    |--group2 <HERDManager>
            |--dataset2
```

In this setup, `dataset1` would resolve to `group1` and `dataset2` to group2, and no object would ever resolve to the root file because it is itself not a HERDManager. I.e., technically, the term `file` is not correct here, but a more accurate phrasing maybe the  `parent herd manager`. However, I understand that: 1) currently the "file" is always the "parent herd manager" and 2) saying "file" is more intuitive for users.

Again, I don't think it is worth-while confusing the user with this detail. buy maybe adjusting the inline comments here to clarify could be helpful.


 


```suggestion
        # Walk up the parent chain looking for the parent HERDManager.
        # In most practical cases, this will be the root file, however, it is
        # possible to construct a file that stores multiple Container objects
        # that each act as separate HERDManager for their child objects
        parent = container.parent
        while parent is not None:
            if isinstance(parent, HERDManager):
                return parent
            parent = parent.parent

        # No HERDManager was found in the container's ancestry. This happens when the
        # container has no parent (e.g., if it is has not been added to a file yet)  or because
        # none of its ancestors are a HERDManager
        msg = ("Could not find the file associated with container '%s'. Please add the container "
               "to the file before adding an external reference." % getattr(container, 'name', container))
```

docs/gallery/plot_external_resources.py
```diff
@@ -186,11 +183,14 @@ def __init__(self, **kwargs):
...
 
 ###############################################################################
-# Using the add_ref method without the file parameter.
+# How add_ref resolves the file
 # ------------------------------------------------------
-# Even though :py:class:`~hdmf.common.resources.File` is required to create/add a new reference,
-# the user can omit the file parameter if the :py:class:`~hdmf.common.resources.Object` has a file
-# in its parent hierarchy.
+# A reference can only be added to a container that has already been added to a file.
+# :py:func:`~hdmf.common.resources.HERD.add_ref` automatically resolves the file by walking
```
oruebel:
```suggestion
# A reference can only be added to a container that has already been added to a file
# (or more accurately to a :py:class:`~hdmf.container.Container` that is a 
# :py:class:`hdmf.container.HERDManager`, which in most practical cases is the file).
# :py:func:`~hdmf.common.resources.HERD.add_ref` automatically resolves the file by walking
```

Not sure if adding this comment here is more confusing than helpful. Feel free to reject the suggestion in case you think this is too confusing.

oruebel (commented):
Looks good to me. I'll let @rly do final approvals since Ryan is working on a few other HERD PRs

rly (approved):

rly:
For historical tracking purposes, I want to note that this PR introduces a breaking change to the HERD interface, but since HDMF 6.0 was introduced not long ago and there are no known consumers, especially for HERD, I think it's fine not to bump to the next major version.

rly:
Looks good to me. Thanks @bendichter

PR #1513, opened 2026-06-23
HERD.add_ref: only warn on entity_uri mismatch, not on re-passing the same URI

bendichter:
## Motivation

From the HERD tutorial review in pynwb (NeurodataWithoutBorders/pynwb#2200):

> let's check to see if these cases are handled. We want to ensure that the tables are always normalized, so adding an identical entry should not duplicate a row

— https://github.com/NeurodataWithoutBorders/pynwb/pull/2200#discussion_r3461368830

I verified the dedup behavior empirically: when `add_ref` is called with an `entity_id` that already exists, `get_entity` finds it and the existing entity row is reused — **no duplicate is ever inserted**, so the tables are normalized as intended.

The remaining wrinkle is ergonomic. Previously `add_ref` warned —

> `This entity already exists. Ignoring new entity uri`

— **whenever** an `entity_uri` was passed for an existing `entity_id`, *even when the URI was identical to the stored one*. Re-passing the same `entity_uri` is the common case when annotating many objects/files with the same entity (e.g. tagging the species of every file in a dandiset), so this produced spurious warnings and pushed that streaming tutorial to special-case it:

```python
entity = herd.get_entity(entity_id="NCBI_TAXON:10090")
entity_uri = None if entity is not None else "https://..."   # only to dodge the warning
herd.add_ref(..., entity_id="NCBI_TAXON:10090", entity_uri=entity_uri)
```

## Change

`add_ref` now warns **only when a different `entity_uri` is provided** for an existing `entity_id`. The existing URI is always kept (unchanged behavior); re-passing the same URI is silent. The mismatch warning is also more informative (it names both URIs and the entity_id).

This lets callers drop the `entity is not None` workaround and simply always pass `entity_uri=...`.

## Tests

- `test_entity_uri_warning_on_mismatch` — a *different* URI for an existing entity warns (asserts the exact message) and keeps the stored URI; the entity is not duplicated.
- `test_entity_uri_no_warning_when_same` — re-passing the *same* URI raises no warning (`warnings.simplefilter("error")`) and does not duplicate the entity row.

All 80 `test_resources` tests pass; `ruff` clean.

Note: this PR is independent of #1512 (the `file`-argument removal) — both touch `add_ref` but in different sections and should merge cleanly in either order.


oruebel (commented):
Looks good to me. I'll let @rly approve since Ryan is working on HERD things right now.

rly (approved):
LGTM!

PR #1515, opened 2026-06-24
Tidy HERD resources.py: drop polymorphic object lookup, de-duplicate blocks

rly:
## Motivation

A survey of `src/hdmf/common/resources.py` while reviewing #1504 turned up internal cruft worth cleaning. It branches off #1504 because it builds on the `create`-flag work there.

## Solution

Three independent cleanups. No public method signatures change.

### Object resolution
`_check_object_field` / `_validate_object` were replaced by three focused helpers:

- `_resolve_object_target(container, attribute) -> (container, relative_path)`
- `_find_object(file, container, relative_path, field) -> Row` (raises if absent)
- `_find_or_add_object(file, ...) -> Row` (adds the object, and its file entry, when absent)

This removes the old helper's flag-dependent return type (an `Object` row on a hit, a `dict` on a miss-with-`create`), which is what let `get_key` fail on a missing object's `idx`.

It also fixes the object lookup's identity model. An object is now identified by its **file** together with its `object_id`, `relative_path`, and `field`. Because an `object_id` is not unique across files (a file can be copied and modified without changing object_ids), the same `object_id` can appear under different files; the old lookup ignored the file and would collide. Restricting matches to the given file resolves each to the correct object, and makes `file` a genuinely-used parameter rather than a passed-but-ignored one.

A container that is not in a file still raises its own distinct error (`Could not find the file associated with container ...`), separate from `Object not in Object Table.` for a container that is in a file but unreferenced.

### `assert_external_resources_equal`
Six copy-pasted try/except blocks become a loop over the table attribute names. Same behavior: every table is compared and all mismatches collect into one `AssertionError`.

### `from_zip`
The six near-identical per-table parse blocks become a dispatch over a `{table name: table class}` mapping, and the five copy-pasted index-range loops become a shared `_assert_idx_in_range` helper. The reduced branching lets the `# noqa: C901` waiver be dropped.

## Behavior change

The only behavior change is the lookup identity fix above: a shared `object_id` across files now resolves correctly instead of colliding. No public method signatures change. (HERD is experimental.)

## How to test the behavior

Existing tests cover the refactor; the `_check_object_field` unit tests are updated for the new helpers, and tests were added for file-based disambiguation (`test_find_object_disambiguates_by_file`), object reuse, and the distinct container-not-in-a-file error. Full `tests/unit` passes (1809 passed, 117 skipped).

## Checklist

- [x] Is your contribution compliant with our coding style? This will be checked during CI.
- [x] Have you checked to ensure that there aren't other open Pull Requests for the same change?


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
...
 
+### Changed
+- Refactored `HERD` object resolution into focused helpers (`_resolve_object_target`, `_find_object`, `_find_or_add_object`), replacing a single helper whose return type (an `Object` row or a dict) depended on a flag. The object lookup now identifies an object by its file together with its object_id, relative_path, and field, so a shared object_id across files (e.g. a copied, modified file) resolves to the correct object instead of colliding. A container that is not in a file continues to raise its own distinct error, separate from a container that is in a file but has no entry in the object table. Also de-duplicated the repetitive table-comparison and `from_zip` parsing/validation blocks. @rly [#1515](https://github.com/hdmf-dev/hdmf/pull/1515)
```
```suggestion
- Refactored `HERD` internals (no public API or behavior change). Renamed `_check_object_field` to `_find_object` object and updated return type to always return an `Object` row instead of returning a row or a `dict` depending on the `create` flag. Also cleaned up code duplication of code for table-comparisons in `assert_external_resources_equal` and  for table parsing/validation in  `from_zip`. @rly [#1515](https://github.com/hdmf-dev/hdmf/pull/1515)
```

rly:
#1504 should be merged first, and this refactor second.

src/hdmf/common/resources.py
```diff
@@ -342,68 +313,68 @@ def _add_entity_key(self, **kwargs):
...
+
+        if len(matches) == 1:
+            return self.objects.row[matches[0]]
+        elif len(matches) > 1:  # pragma: no cover
+            # It isn't possible for this to happen unless the user used _add_object directly.
+            raise ValueError("Found multiple instances of the same object id, relative path, "
+                             "and field in objects table.")
+        return None
+
+    def _find_object(self, file, container, relative_path, field):
```
oruebel: Just a minor comment. The `_find_object` function seems a bit superfluous since it only turns the None return value of `_object_row` into a `ValueError`. I think it could be removed and `object_row` be renamed to `_find_object`. Up to you.

oruebel (approved):

src/hdmf/common/resources.py
rly: Good point.
```
if object_field is None:                                                            
    raise ValueError("Object not in Object Table.")
```
gets copied in two places but I think removing a layer of function calls would be less. Fixed.

PR #1517, opened 2026-06-25
Use PyPI trusted publishing and gh CLI for releases

rly:
## Motivation

Modernize the release publishing path in `deploy_release.yml` (suggestions 1 + 4 from the DevOps infra review):

1. **PyPI trusted publishing (OIDC).** The PyPI upload used `twine upload -u <user> -p <password>` with a long-lived password stored as a secret and passed on the command line (visible in the process list). It now uses `pypa/gh-action-pypi-publish` with OIDC trusted publishing: short-lived, no stored credentials. The job declares `id-token: write` and runs in a dedicated `pypi` environment.
4. **Drop stale release tooling.** The GitHub release was created with `githubrelease` (a lightly maintained third-party CLI installed at release time). It now uses the `gh` CLI that ships on the runner, authenticated with the built-in `GITHUB_TOKEN` (`contents: write`) instead of `BOT_GITHUB_TOKEN`, and the release body is generated from the changelog (see Notes).

## ⚠️ Required one-time setup before the next release

**edit: Done**

PyPI trusted publishing will not work until a trusted publisher is configured **on PyPI** (this cannot be done from the repo):

- PyPI → `hdmf` project → Settings → Publishing → Add a new GitHub publisher:
  - Owner: `hdmf-dev`, Repository: `hdmf`
  - Workflow: `deploy_release.yml`
  - Environment: `pypi`

After this, the `BOT_PYPI_USER` / `BOT_PYPI_PASSWORD` secrets can be removed.

## Notes / things to confirm

- The GitHub release is now authored by `github-actions[bot]` rather than the bot PAT account. No workflow triggers on `release:` events, so losing the PAT's ability to trigger downstream workflows is not a concern here.
- The GitHub release notes are built automatically from this version's `CHANGELOG.md` section (the block under `## HDMF <tag>`), so the post-release "paste the changelog into the release" step is no longer needed. It falls back to `--generate-notes` if no matching section is found.
- The `deploy-dev` pre-release job in `run_tests.yml` still uses `scikit-ci-addons`; its rolling pre-release logic is more involved and is left for a separate change.

## How to test the behavior?
```
Cannot be fully exercised without a tag push + the PyPI trusted publisher configured.
Verified: YAML parses; the publish/release steps are standard, documented usages of
pypa/gh-action-pypi-publish and the gh CLI.
```

## Checklist

- [ ] Did you update `CHANGELOG.md` with your changes? (CI-only change; no user-facing entry)
- [x] Have you reviewed our Contributing Guide?
- [ ] Does the PR use "Fix #XXX" notation? (no associated issue)

### DIFF
--- a/.github/workflows/deploy_release.yml
+++ b/.github/workflows/deploy_release.yml
@@ -16,13 +16,13 @@ jobs:
       contents: write  # create the GitHub release
     steps:
       - name: Checkout repo with submodules
-        uses: actions/checkout@v7
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
         with:
           submodules: 'recursive'
           fetch-depth: 0  # tags are required for versioneer to determine the version
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6.3.0
         with:
           python-version: '3.14'
 
@@ -53,7 +53,7 @@ jobs:
         # Publishes to PyPI via OIDC trusted publishing. The hdmf project on PyPI must have a trusted
         # publisher configured for the hdmf-dev/hdmf repository, the deploy_release.yml workflow, and the
         # pypi environment.
-        uses: pypa/gh-action-pypi-publish@v1.14.0
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
           skip-existing: true
 
--- a/.github/workflows/deploy_release.yml
+++ b/.github/workflows/deploy_release.yml
@@ -18,6 +18,7 @@ jobs:
       - name: Checkout repo with submodules
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
         with:
+          persist-credentials: false
           submodules: 'recursive'
           fetch-depth: 0  # tags are required for versioneer to determine the version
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_release.yml
```diff
@@ -8,15 +8,22 @@ jobs:
...
+    environment:
+      name: pypi
+      url: https://pypi.org/p/hdmf
+    permissions:
+      id-token: write  # mint the OIDC token for PyPI trusted publishing
+      contents: write  # create the GitHub release
     steps:
       - name: Checkout repo with submodules
-        uses: actions/checkout@v7
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
```
oruebel, line 19: Could you clarify why all the actions are tied to a particular sha rather than a release version, i.e., `actions/checkout@v7` vs `actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0` and similarly for the other actions, e.g., setup-python? Indicating the actual version seems more readable and will keep the action updated within the version, rather than having to update sha every time.

oruebel (approved):

.github/workflows/deploy_release.yml
rly: Last year, a popular GitHub Action `tj-actions/changed-files` was compromised and most release tags were updated to point to a new commit with code that leaked secrets. https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised. (I mentioned this and another attack on Slack.)

To prevent these attacks, GitHub now supports ["immutable releases"](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases), but adoption has been slow across the actions community. Surprisingly, not even the github-maintained core "actions/checkout" action is immutable: https://github.com/actions/checkout/issues/2316 . The ruff action has started to make [immutable releases](https://github.com/astral-sh/ruff-action/releases/tag/v4.0.0), and so I think pinning to 4.0.0 there is fine (note that they stopped publishing "floating tags" (v4, v4.0) for similar reasons). Until actions are immutable, pinning to a specific release commit gives the highest security possible, at the costs that you mention. I think in a year, actions will be immutable, and we can return to tag-based pointers.

oruebel: That makes sense. Thanks for the additional background. It looks like the GitHub dependency bot also checks the SHAs automatically so updating to newer version may actually be fairly straight-forward in practice that way.

Issue #423, opened 2020-08-28
Add support for other `quantity` values in schema


rly:
This is low priority but worth documenting.

HDMF supports only these values for the `quantity` key: '*', '?', '+', and 1.

The schema language allows values other values: `'zero_or_many'`, `'one_or_many'`, `'zero_or_one'`, and integers larger than 1.
https://hdmf-schema-language.readthedocs.io/en/latest/specification_language_description.html#quantity

We could change the schema language to remove those options, or add support for them in HDMF. I think the latter would be better.

oruebel:
> ... or add support for them in HDMF

I agree, that is the preferred option

PR #1521, opened 2026-06-26
Validate quantity in GroupSpec and DatasetSpec

### LINKED ISSUE #531: Disallow or explicitly allow spec quantity to be specified with a numeric string
The codebase seems to assume that a spec quantity is either a string from a reserved set of special characters or named wildcard quantities or an integer value. However, the spec schema documentation does not seem to exclude numeric strings and DatasetSpec accepts a numeric string quantity.

```
>>> from hdmf.spec import DatasetSpec
>>> ds = DatasetSpec('A Dataset', data_type_inc='Foo', quantity=1)
>>> ds.is_many()
False
>>> ds = DatasetSpec('A Dataset', data_type_inc='Foo', quantity='1')
>>> ds.is_many()
True
```

I'm not sure if this is a problem that's actually been encountered, but it seems likely to come up at some point.

This came up while testing #526 and trying to determine what edge cases needed to be tested.

Options:
* disallow numeric strings in the schema documentation and in `BaseStorageSpec`
* explicitly allow it in the documentation and convert numeric strings to integer value in `BaseStorageSpec.__init__`

**Describe alternatives you've considered**
Treat both integer and numeric strings in all cases where quantity is used.

**Additional context**
None

## Checklist

jwbear:
ISSUE #531

## Motivation

Fix #531

## How to test the behavior?
```
```
Test FLAGS not in dictionary, integers, strings, and integers entered as strings eg "2"
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
+- Fix #531 Disallow or explicitly allow spec quantity to be specified with a numeric string @jwbear [#1521]
 - Fixed iterative HDF5 writes (e.g. from a `DataChunkIterator`) storing data as `float32` when no explicit dtype was set on the builder. The data's own dtype is now used, so e.g. integer data is stored as integers rather than silently downcast, and an `H5pyDeprecationWarning` is no longer emitted. @rly [#1519](https://github.com/hdmf-dev/hdmf/pull/1519)
 - Removed the broken `str` (object_id) option from the `container` argument of `HERD.add_ref`, `HERD.add_ref_termset`, `HERD.get_key`, and `HERD.get_object_entities`; these methods now require an `AbstractContainer` and reject a string with a clear type error. @rly [#1514](https://github.com/hdmf-dev/hdmf/pull/1514)
 - Fixed `HERD.get_object_entities(attribute=...)` not finding references added with `HERD.add_ref(attribute=...)`, which raised `KeyError`/`TypeError` for non-DataType attributes. Both methods now resolve the attribute the same way. @rly [#1504](https://github.com/hdmf-dev/hdmf/pull/1504)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
-- Fix #531 Disallow or explicitly allow spec quantity to be specified with a numeric string @jwbear [#1521]
+- Fix #531 Validate quantity specified with an integor or numeric string @jwbear [#1521]
 - Fixed iterative HDF5 writes (e.g. from a `DataChunkIterator`) storing data as `float32` when no explicit dtype was set on the builder. The data's own dtype is now used, so e.g. integer data is stored as integers rather than silently downcast, and an `H5pyDeprecationWarning` is no longer emitted. @rly [#1519](https://github.com/hdmf-dev/hdmf/pull/1519)
 - Removed the broken `str` (object_id) option from the `container` argument of `HERD.add_ref`, `HERD.add_ref_termset`, `HERD.get_key`, and `HERD.get_object_entities`; these methods now require an `AbstractContainer` and reject a string with a clear type error. @rly [#1514](https://github.com/hdmf-dev/hdmf/pull/1514)
 - Fixed `HERD.get_object_entities(attribute=...)` not finding references added with `HERD.add_ref(attribute=...)`, which raised `KeyError`/`TypeError` for non-DataType attributes. Both methods now resolve the attribute the same way. @rly [#1504](https://github.com/hdmf-dev/hdmf/pull/1504)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
-- Fix #531 Validate quantity specified with an integor or numeric string @jwbear [#1521]
+- Fix #531 Validate quantity specified with an integor or numeric string @jwbear [#1521] (https://github.com/hdmf-dev/hdmf/pull/1521)
 - Fixed iterative HDF5 writes (e.g. from a `DataChunkIterator`) storing data as `float32` when no explicit dtype was set on the builder. The data's own dtype is now used, so e.g. integer data is stored as integers rather than silently downcast, and an `H5pyDeprecationWarning` is no longer emitted. @rly [#1519](https://github.com/hdmf-dev/hdmf/pull/1519)
 - Removed the broken `str` (object_id) option from the `container` argument of `HERD.add_ref`, `HERD.add_ref_termset`, `HERD.get_key`, and `HERD.get_object_entities`; these methods now require an `AbstractContainer` and reject a string with a clear type error. @rly [#1514](https://github.com/hdmf-dev/hdmf/pull/1514)
 - Fixed `HERD.get_object_entities(attribute=...)` not finding references added with `HERD.add_ref(attribute=...)`, which raised `KeyError`/`TypeError` for non-DataType attributes. Both methods now resolve the attribute the same way. @rly [#1504](https://github.com/hdmf-dev/hdmf/pull/1504)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
-- Fix #531 Validate quantity specified with an integor or numeric string @jwbear [#1521] (https://github.com/hdmf-dev/hdmf/pull/1521)
+- Fixed #531 Validate quantity specified with an integer or numeric string @jwbear [#1521](https://github.com/hdmf-dev/hdmf/pull/1521)
 - Fixed iterative HDF5 writes (e.g. from a `DataChunkIterator`) storing data as `float32` when no explicit dtype was set on the builder. The data's own dtype is now used, so e.g. integer data is stored as integers rather than silently downcast, and an `H5pyDeprecationWarning` is no longer emitted. @rly [#1519](https://github.com/hdmf-dev/hdmf/pull/1519)
 - Removed the broken `str` (object_id) option from the `container` argument of `HERD.add_ref`, `HERD.add_ref_termset`, `HERD.get_key`, and `HERD.get_object_entities`; these methods now require an `AbstractContainer` and reject a string with a clear type error. @rly [#1514](https://github.com/hdmf-dev/hdmf/pull/1514)
 - Fixed `HERD.get_object_entities(attribute=...)` not finding references added with `HERD.add_ref(attribute=...)`, which raised `KeyError`/`TypeError` for non-DataType attributes. Both methods now resolve the attribute the same way. @rly [#1504](https://github.com/hdmf-dev/hdmf/pull/1504)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
-- Fixed #531 Validate quantity specified with an integer or numeric string @jwbear [#1521](https://github.com/hdmf-dev/hdmf/pull/1521)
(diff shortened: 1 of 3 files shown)


jwbear:
Added validate quantity and tests for validate_quantity, updated change log, and converted to private static method. Added name as parameter as outside scope for validate_quantity.

src/hdmf/spec/spec.py
```diff
@@ -513,6 +516,49 @@ def __init__(self, **kwargs):
...
+        Args:
+            name: name string.
+            qty: quantity string or number.
+
+        Returns:
+            A validated quantity string.
+
+        Raises:
+            ValueError: If quantity is not in FLAGS or is not >=1.
+            TypeError: If quantity is not an integer.
```
oruebel: HDMF currently uses Sphinx-style docstrings. Could you please update the docstring accordingly.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
+- Fix #531 Disallow or explicitly allow spec quantity to be specified with a numeric string @jwbear [#1521]
```
oruebel: `Disallow or explicitly allow` may be confusing to readers because it does not communicates the issue, rather then the behavior that was implemented. In the changelog it will be useful to state the the actual change/behavior that was implemented. 

Also, if the approach implemented here constitutes a change in behavior of the API, then I would move the item the "Changed" section instead

src/hdmf/spec/spec.py
oruebel, line 448: Since the behavior for `quantity` has changed, it may be useful to clarify this also here in the `doc` for the parameter

src/hdmf/spec/spec.py
jwbear: Done.

CHANGELOG.md
jwbear: Okay, this is the title of the issue changed to added better descriptor

oruebel:
The code looks overall good to me. I added a few suggestions mostly related to documentation and asked Copilot to review. I'll let @rly do final review and approvals.

src/hdmf/spec/spec.py
jwbear: This still seems correct as we simply expanded the handling inputs and not the required values.

src/hdmf/spec/spec.py
```diff
@@ -513,6 +516,49 @@ def __init__(self, **kwargs):
...
+        )
+        valid_flags = ", ".join(FLAGS.values())
+        invalid_int = f"Invalid quantity '{qty}': must be greater than or equal to 1 or in '[{valid_flags}]'"
+        # Check FLAGS
+        if isinstance(qty, str) and ((qty in FLAGS.values()) or (qty in FLAGS)):
+            if qty in (ONE_OR_MANY, ZERO_OR_MANY):
+                if name is not None:
+                    raise ValueError(invalid_name)
+            return qty
+
```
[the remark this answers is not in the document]
jwbear: these values are not modified or returned here and only used in string msg and comparison.

src/hdmf/spec/spec.py
```diff
@@ -513,6 +516,49 @@ def __init__(self, **kwargs):
...
+            name: name string.
+            qty: quantity string or number.
+
+        Returns:
+            A validated quantity string.
+
+        Raises:
+            ValueError: If quantity is not in FLAGS or is not >=1.
+            TypeError: If quantity is not an integer.
+        """
```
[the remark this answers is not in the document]
jwbear: True removed type error

tests/unit/spec_tests/test_group_spec.py
```diff
@@ -302,6 +302,36 @@ def test_name_with_multiple(self):
         with self.assertRaisesWith(ValueError, msg):
             GroupSpec(doc='A test group', name='MyGroup', quantity='*')
 
+    def test_quantity_with_string_float(self):
+        msg = ("Invalid quantity '3.5': must be greater than or equal to 1 or in '[?, *, +]'")
+        with self.assertRaisesWith(ValueError, msg):
+            GroupSpec(doc='A test group', name='MyGroup', quantity='3.5')
+
```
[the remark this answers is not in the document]
jwbear: Agreed, no valid testing code section for quantity

CHANGELOG.md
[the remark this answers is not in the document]
jwbear: updated

jwbear:
@oruebel Thanks, I updated everything and will push again based on the suggestions. Except I ddint see positive tests. copilot: "These tests cover invalid quantity strings, but they don't assert the new intended behavior from #531: numeric strings like '2' should be accepted and treated as integers (so '1' does not imply 'many'). Adding a positive-case regression test would prevent the original bug from reappearing."

src/hdmf/spec/spec.py
rly: Since we're updating how "quantity" is handled, this is a good time to make this docstring more informative. Could you update it to say something like `f"the required number of instances. Must be a positive integer, a string representation of a positive integer, or one of the flag keys {list(FLAGS.keys())} or values {list(FLAGS.values())}."`

src/hdmf/spec/spec.py
```diff
@@ -513,6 +516,49 @@ def __init__(self, **kwargs):
...
+        :raises ValueError: If quantity is not in FLAGS or is not >= 1.
+        :raises TypeError: If quantity is not an integer.
+        """
+        invalid_name = (
+            f"Cannot give specific name to something that can exist multiple times: name='{name}', quantity='{qty}'"
+        )
+        valid_flags = ", ".join(FLAGS.values())
+        invalid_int = f"Invalid quantity '{qty}': must be greater than or equal to 1 or in '[{valid_flags}]'"
+        # Check FLAGS
+        if isinstance(qty, str) and ((qty in FLAGS.values()) or (qty in FLAGS)):
```
rly, line 540: I think `or (qty in FLAGS)` should not be included _yet_. While quantity values like `"zero_or_one"` were allowed before, it resulted in a spec that was not processed correctly downstream in `required()/is_many()` as Copilot mentions below. This is a separate issue https://github.com/hdmf-dev/hdmf/issues/423 that should be addressed separately. Until that is properly supported, I think HDMF should complain with an error if a user provides a FLAGS key like `quantity="zero_or_one"`.

src/hdmf/spec/spec.py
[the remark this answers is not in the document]
rly: see above comment for different approach

tests/unit/spec_tests/test_group_spec.py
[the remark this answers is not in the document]
rly: A positive test like what is described would be good to add.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
+- Fix #531 Validate quantity specified with an integor or numeric string @jwbear [#1521]
```
rly: Please add the URL to the end of this line so that `#1521` renders as a Markdown link in the changelog

rly: Also please make this past tense, and correct spelling of "integer". "Quantity" should be clarified to be GroupSpec and DatasetSpec quantity. We typically don't mention the issue number since the PR links back to the issue. Either way is fine. So something like:

```
- Fixed issue #531 where invalid values were accepted for `GroupSpec.quantity` and `DatasetSpec.quantity`. Now only positive integers, string representations of positive integers, and '?', '*', and '+' are allowed. @jwbear [#1521](https://github.com/hdmf-dev/hdmf/pull/1521)
```

src/hdmf/spec/spec.py
```diff
@@ -513,6 +516,49 @@ def __init__(self, **kwargs):
...
+            if qty in (ONE_OR_MANY, ZERO_OR_MANY):
+                if name is not None:
+                    raise ValueError(invalid_name)
+            return qty
+
+        # Convert numeric strings
+        if isinstance(qty, str):
+            try:
+                qty = int(qty)
+            except (ValueError):
```
rly: It's more conventional to use `except ValueError` for single errors.

Separately, re-raising the ValueError from within the `except` block will result in the traceback:
```
    File "<stdin>", line 6, in validate
  ValueError: invalid literal for int() with base 10: 'foo'

    File "<stdin>", line 11, in <module>
    File "<stdin>", line 8, in validate
  ValueError: Invalid quantity 'foo': must be >= 1 or in '[?, *, +]'
```

I think that would be confusing for users, and it would be better if only your new, clearer error message was shown. To do that, add `from None` after `raise ValueError(invalid_int)`

src/hdmf/spec/spec.py
rly: On further review, I found that HDMF already contains tests that check the FLAGS keys like "zero_or_one". Dropping `or (qty in FLAGS)` would break the test and be a breaking change, even if the behavior wasn't 100% working before. Probably a better approach would be to keep supporting it here, but prioritize fixing `required()/is_many()` which I can work on after this is merged. So never mind. No change needed for this line.

src/hdmf/spec/spec.py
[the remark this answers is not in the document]
rly: The new docstring still says TypeError

src/hdmf/spec/spec.py
jwbear: If I dont allow it then tests fail where the key is being passed in: FAILED tests/unit/spec_tests/test_dataset_spec.py::DatasetSpecTests::test_name_with_compatible_quantity - ValueError: Invalid quantity 'zero_or_one': must be greater than or equal to 1 or in '[?, *, +]'

CHANGELOG.md
jwbear: Done.

src/hdmf/spec/spec.py
jwbear: Okay. Done.

tests/unit/spec_tests/test_group_spec.py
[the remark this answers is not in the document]
rly: You can add it to `test_dataset_spec.py` where the other positive tests live currently

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Hardened the GitHub Actions CI: added least-privilege `permissions` blocks, pinned actions to commit SHAs (or immutable release tags), deduplicated the test setup into a composite action, passed untrusted inputs through environment variables, and added a zizmor security audit of the workflows. @rly [#1518](https://github.com/hdmf-dev/hdmf/pull/1518)
 
 ### Fixed
+- Fix #531 Validate quantity specified with an integor or numeric string @jwbear [#1521] (https://github.com/hdmf-dev/hdmf/pull/1521)
```
```suggestion
- Fixed #531 Validate quantity specified with an integer or numeric string @jwbear [#1521](https://github.com/hdmf-dev/hdmf/pull/1521)
```

Fixed spelling and removed extra space

CHANGELOG.md
rly: Please update text to be more descriptive of the change to users. Something like the above

src/hdmf/spec/spec.py
jwbear: Okay, that sounds good, but this is scope creep and should be in a new issue. I would also point out we are not really changing how qty is handled. Instead we are fixing the validation. It is a very normal test case to validate when ints are entered as strings and vice versa. We are also only supporting explicit dictionary calls because prior work explicitly supports it.

src/hdmf/spec/spec.py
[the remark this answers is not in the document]
jwbear: Done but as this does not effect functionality or development, and is an enhancement over the existing standard of document the Nit tag may be appropriate.

CHANGELOG.md
jwbear: Done.

jwbear:
Hi
To help us move faster and keep our code reviews stress-free, I want to try a more collaborative approach for our CR rounds. 

I am going to create a separate branch specifically for additional positive unit tests, since that falls outside the scope of this original PR. Moving forward, let's try to list any updates, structural enhancements, or cosmetic changes that don't affect core functionality as separate  issues. This will help us limit the number of back-and-forth CR rounds and get our code shipped much faster!

We can also start using the `[nit]` tag for general personal preferences or style ideas that aren't strictly necessary to change. That way, we can share cool ideas without blocking the merge.

Here is a list of well-known code review anti-patterns and some friendly quick-replies we can use. Maybe we can work on adopting these together to speed up our development? Let me know what you think!

### 🔍 Code Review Anti-Patterns & Friendly Solutions

| Anti-Pattern | Description | 💬 Collaborative Git Comment Solution |
| :--- | :--- | :--- |
| **Bikeshedding** | Spending massive time debating trivial details like variable naming or formatting. | `[nit] Looks good! Let's defer this specific style point to our automated formatter so we can unblock the build.` |
| **LGTM Stampede** | Blindly approving massive, unreadable PRs without looking closely. | `[blocking] This looks awesome, but it's quite large (>300 lines) to review safely. Let's split it up so we don't miss anything!` |
| **Moving the Goalposts** | Adding new structural demands on subsequent review passes after initial feedback. | `[suggestion] Really great catch here! Let's log this new architectural improvement as a separate follow-up ticket so we can merge this piece.` |
| **Scope Creep** | Forcing authors to refactor old, unrelated code within a functional PR. | `[nit] Let's keep this PR scoped strictly to the current task. I'll gladly open a quick technical debt ticket to track this refactor next!` |
| **Ghost Reviewer** | Reviewers who ignore notifications and stall deployment pipelines. | `ping @reviewer - Checking in to see if you have cycles for a quick look today, or if I should loop in a backup to help share the load?` |

src/hdmf/spec/spec.py
rly: OK, that's fine. We'll update this docstring as a separate issue.

src/hdmf/spec/spec.py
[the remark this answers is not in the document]
rly: Thanks @jwbear. I agree that this is nit-level. One small thing though - the function only raises `ValueError`, so in your most recent commit, the `TypeError` line should have been removed, not the `ValueError`. Since it was a two-line change I just pushed it directly to save us the round trip.

rly:
Sounds good. A separate PR for the positive tests works. Thanks for offering to do that. Thanks also for bearing with the long review. Going forward, I'll flag nits as non-blocking from the start to help streamline reviews.

Nothing else blocking from me. Looks good and congrats on your first contribution to HDMF!

rly (approved):

tests/unit/spec_tests/test_group_spec.py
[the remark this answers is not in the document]
rly: Deferred to separate PR as discussed below.

PR #1533, opened 2026-07-07
Fix rank-3 maxshape when writing a reference column targeting a DynamicTable

### LINKED ISSUE #1532: Object-reference VectorData column targeting a DynamicTable fails to write (rank-3 maxshape from get_data_shape recursing into the table)
While building [ndx-guppy](https://github.com/catalystneuro/ndx-guppy) (an NWB extension for the GuPPy fiber-photometry processing tool), I was migrating its event references to the pyNWB 4.0 `EventsTable` model. I have a small registry table (`GuppyEventsTable`, one row per behavioral event) and I wanted each row to hold an **object reference to the `EventsTable`** that stores that event's onset timestamps — i.e. "point this row at that whole table." The spec column is a `VectorData` with `dtype: {target_type: EventsTable, reftype: object}`.

It constructs fine but fails on write with:

```
ValueError: 'maxshape' must have same rank as dataset shape
```

**Root cause.** When the HDF5 backend writes the reference dataset, it derives the `maxshape` from `hdmf.utils.get_data_shape` on the column data. Because a `DynamicTable` is `len()`-able and indexable, `get_data_shape` recurses *into* the referenced table as if it were nested array data and returns a rank-3 shape for what is logically a 1-D array of references:

```python
from hdmf.utils import get_data_shape
t = DynamicTable(name="target", description="t"); t.add_column(name="x", description="c", data=[]); t.add_row(x=1.0)
get_data_shape([t])   # -> (1, 1, 1)   # expected (1,): one reference
```

The write path then stamps that rank-3 shape onto the `(1,)` reference dataset as its `maxshape`, and h5py rejects the rank mismatch (`require_dataset(..., shape=(1,), maxshape=(None, None, None))`).

**Why it's specific to table targets.** An object-reference column targeting a non-sized container (e.g. `TimeSeries`) writes fine — `get_data_shape([time_series]) == (1,)` because a `TimeSeries` has no `__len__`. And whole-table references stored as **attributes** (e.g. `DynamicTableRegion.table`) are unaffected, since attributes aren't shape-inferred. The bug only bites an object-reference *dataset column* whose targets are themselves table-shaped.

**Self-contained repro** (pynwb 4.0.0 / hdmf 6.1.0 / h5py 3.16.0):

```python
import tempfile, os
from pynwb import get_class, load_namespaces, NWBHDF5IO
from pynwb.spec import NWBNamespaceBuilder, NWBGroupSpec, NWBDatasetSpec, NWBRefSpec, export_spec

ns = "ex-tableref"
b = NWBNamespaceBuilder(name=ns, version="0.1.0", doc="repro", author=["x"], contact=["x@x"])
b.include_namespace("core")
spec = NWBGroupSpec(
    neurodata_type_def="RefTable", neurodata_type_inc="DynamicTable", doc="table with an object-ref column",
    datasets=[NWBDatasetSpec(name="ref", neurodata_type_inc="VectorData",
                             doc="object reference to a whole DynamicTable",
                             dtype=NWBRefSpec(target_type="DynamicTable", reftype="object"))])
d = tempfile.mkdtemp(); export_spec(b, [spec], output_dir=d)
load_namespaces(os.path.join(d, f"{ns}.namespace.yaml")); RefTable = get_class("RefTable", ns)

nf = mock_NWBFile(); pm = nf.create_processing_module(name="m", description="m")
tgt = DynamicTable(name="target", description="t"); tgt.add_column(name="x", description="c", data=[]); tgt.add_row(x=1.0)
pm.add(tgt)
rt = RefTable(name="reftable", description="r"); rt.add_row(ref=tgt); pm.add(rt)
with NWBHDF5IO(os.path.join(d, "repro.nwb"), "w") as io:
    io.write(nf)   # -> ValueError: 'maxshape' must have same rank as dataset shape
```

### Environment
- hdmf 6.1.0
- pynwb 4.0.0
- h5py 3.16.0
- Python 3.13, macOS

pauladkisson:
Fixes #1532.

## What changed

`ObjectMapper.__get_spec_info` now treats a reference dataset like a compound dataset when computing its shape — it takes the length of the reference array (1D) instead of calling `get_data_shape`, which recurses into `DynamicTable` reference targets (they are `len()`/index-able) and reports a spurious higher-rank shape. That rank was matched against `VectorData`'s multi-dimensional allowed shapes and used as the `maxshape`, so the write raised `ValueError: 'maxshape' must have same rank as dataset shape` even though the backend writes the reference array 1D.

The check covers both references declared via `RefSpec` and untyped columns whose data are references.

## Testing

Added `TestDefaultExpandableWithTableReferences` (roundtrip + export) — a `VectorData` reference column whose targets are `DynamicTable`s; it fails before the change and passes after. The existing build / common / HDF5 suites pass (1146 passed, 0 regressions).

### DIFF
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -1007,7 +1007,13 @@ class ObjectMapper(metaclass=ExtenderMeta):
         if spec_shape is None:
             return None, None
         else:
-            if isinstance(spec.dtype, list):
+            if isinstance(spec.dtype, (list, RefSpec)) or self.__is_reftype(data):
+                # A compound (list) or reference dataset is written 1D with one element per entry, so
+                # treat it as 1D of length len(data). This covers references declared in the spec
+                # (RefSpec) as well as an untyped column whose data are references. Using
+                # get_data_shape here would recurse into reference targets (e.g. a DynamicTable,
+                # which is len()/index-able) and report a spurious higher-rank shape, yielding a
+                # maxshape whose rank does not match the 1D dataset the backend actually writes.
                 data_shape = (_get_length(data),)
             else:
                 data_shape = get_data_shape(data)
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -988,7 +988,12 @@ class ObjectMapper(metaclass=ExtenderMeta):
         spec
             The DatasetSpec or AttributeSpec providing ``shape``, ``dims``, and ``dtype`` to
             match against. When ``spec.dtype`` is a list (compound dtype), the data is treated
-            as 1D with length equal to ``len(data)`` for the purpose of shape matching.
+       spec
+           The DatasetSpec or AttributeSpec providing ``shape``, ``dims``, and ``dtype`` to
+           match against. When ``spec.dtype`` is a list (compound dtype), or a``RefSpec``, or
+           when ``data`` holds references, the data is treated as 1D with length equal to
+           ``len(data)`` for the purpose of shape matching, since the backend writes such
+           datasets 1D with one element per entry.
 
         Returns
         -------
--- a/src/hdmf/build/objectmapper.py
+++ b/src/hdmf/build/objectmapper.py
@@ -987,13 +987,10 @@ class ObjectMapper(metaclass=ExtenderMeta):
             any array-like object supported by :func:`get_data_shape`.
         spec
             The DatasetSpec or AttributeSpec providing ``shape``, ``dims``, and ``dtype`` to
-            match against. When ``spec.dtype`` is a list (compound dtype), the data is treated
-       spec
-           The DatasetSpec or AttributeSpec providing ``shape``, ``dims``, and ``dtype`` to
-           match against. When ``spec.dtype`` is a list (compound dtype), or a``RefSpec``, or
-           when ``data`` holds references, the data is treated as 1D with length equal to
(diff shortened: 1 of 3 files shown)


rly (approved):

rly:
The `__get_spec_info` docstring still describes only the compound-dtype case, but this PR broadens the 1D handling to `RefSpec` dtypes and reference-holding data. ~Could you update the `spec` parameter description (around lines 989–991) so the docs match the new behavior?~ I was able to make the change.

````markdown
```suggestion
        spec
            The DatasetSpec or AttributeSpec providing ``shape``, ``dims``, and ``dtype`` to
            match against. When ``spec.dtype`` is a list (compound dtype) or a ``RefSpec``, or
            when ``data`` holds references, the data is treated as 1D with length equal to
            ``len(data)`` for the purpose of shape matching, since the backend writes such
            datasets 1D with one element per entry.
```
````

File: src/hdmf/build/objectmapper.py
Comment (rly) [inline src/hdmf/build/objectmapper.py]:
```suggestion
       spec
           The DatasetSpec or AttributeSpec providing ``shape``, ``dims``, and ``dtype`` to
           match against. When ``spec.dtype`` is a list (compound dtype), or a``RefSpec``, or
           when ``data`` holds references, the data is treated as 1D with length equal to
           ``len(data)`` for the purpose of shape matching, since the backend writes such
           datasets 1D with one element per entry.
```

rly (approved):

rly:
Thanks @pauladkisson !

PR #1539, opened 2026-07-12
Fix _parse_isoformat to read the ISO 8601 'Z' (UTC) designator (Fix #1524)

### LINKED ISSUE #1524: _parse_isoformat cannot read the ISO 8601 'Z' (UTC) designator on Python < 3.11
## Summary

HDMF parses `isodatetime` strings with `datetime.datetime.fromisoformat()` (in `_parse_isoformat`, `src/hdmf/build/objectmapper.py`). In Python 3.7–3.10, `fromisoformat` cannot parse the ISO 8601 `Z` (UTC) designator, so reading an otherwise-valid `isodatetime` value that ends in `Z` raises:

```
ValueError: Invalid isoformat string: '2026-07-02T08:27:27.216166Z'
```

`Z`-suffix support was only added to `fromisoformat` in Python 3.11, so this affects any HDMF/PyNWB install running on Python 3.10.

## Schema vs. implementation

This is a genuine schema-vs-implementation gap for `isodatetime`:

- **The schema recommends `Z` for UTC.** The NWB schema documents it explicitly (e.g. `session_start_time`: *"Dates stored in UTC end in 'Z' with no timezone offset"*), and `Z` is valid ISO 8601.
- **HDMF writes `+00:00`, not `Z`.** `_isoformat` serializes via `datetime.isoformat()`, which emits a numeric offset (`+00:00`) for UTC. So HDMF neither emits the schema-recommended form nor reads it on older Python.
- **Consequence:** files written by other tools that follow the schema and emit `Z` for UTC (e.g. MatNWB, on several MATLAB releases) cannot be read by HDMF/PyNWB on Python 3.10 — even though `Z` is valid ISO 8601 and the schema-preferred form.

## Impact

On Python 3.10, HDMF/PyNWB fails to read `isodatetime` fields that use the `Z` suffix (`session_start_time`, `file_create_date`, `timestamps_reference_time`, etc.).

## Suggested fix

In `_parse_isoformat`, normalize a trailing `Z` to `+00:00` (or use a parser that accepts `Z` across all supported Python versions) before calling `fromisoformat`, so HDMF reads both `Z` and `+00:00` regardless of Python version. This also resolves the asymmetry where HDMF writes `+00:00` but cannot read `Z`.

## Notes

- Reproduced with a MatNWB-written file read by PyNWB on Python 3.10.
- MatNWB is separately being updated to emit `+00:00` (matching HDMF's own output) as a workaround, but the underlying HDMF reader limitation affects any `Z`-form file from any source.

Leonard013:
## Motivation

`ObjectMapper._parse_isoformat` (`src/hdmf/build/objectmapper.py`) reads a stored ISO
8601 timestamp by passing it straight to `datetime.datetime.fromisoformat(value)` with
no normalization. On **Python < 3.11**, `fromisoformat` does **not** accept a trailing
`Z` (the UTC "Zulu" designator) — CPython only added `Z` support in 3.11 (bpo-35829 /
gh-80010). So reading a `...Z` timestamp raises:

```
ValueError: Invalid isoformat string: '2026-07-02T08:27:27.216166Z'
```

This matters because hdmf's `requires-python` is `>=3.10`, so **Python 3.10 is an
officially supported interpreter** — "just upgrade Python" is not a valid answer.

There is also a **reader/writer asymmetry inside hdmf itself**: the writer
(`_isoformat`) serializes a UTC-aware datetime via `datetime.isoformat()`, which emits
the `+00:00` offset — never `Z`. So hdmf can always read what hdmf wrote, but a `Z`
timestamp produced by a peer tool (NWB files and many other producers commonly write
`Z`) denotes the *same instant* as the `+00:00` hdmf emits, yet is unreadable on 3.10.

## What this changes

A one-branch normalization confined to `_parse_isoformat`: before handing a datetime
string to `fromisoformat`, a trailing `Z`/`z` is rewritten to the equivalent `+00:00`
offset.

```python
if 'T' in value or ' ' in value:
    if value.endswith(('Z', 'z')):
        value = value[:-1] + '+00:00'
    return datetime.datetime.fromisoformat(value)
```

Properties:
- **Version-independent and safe.** On 3.11+ it is a harmless normalization (`Z` and
  `+00:00` already parse identically); on 3.10 it makes `Z` parse. No `sys.version`
  check needed.
- **Only the final designator** is touched, and only in the datetime branch. Real
  offsets (`+00:00`, `-05:00`), naive strings (which stay naive), date-only strings,
  and already-parsed `datetime`/`date` passthroughs are untouched.
- Also normalizes lowercase `z` (see reviewer note below).

## How to test

```python
from hdmf.build.objectmapper import _parse_isoformat
# On Python 3.10 this raises ValueError before the fix; after the fix it equals the
# '+00:00' form on every supported Python:
assert _parse_isoformat("2026-07-02T08:27:27.216166Z") == \
       _parse_isoformat("2026-07-02T08:27:27.216166+00:00")
```

New tests in `tests/unit/build_tests/mapper_tests/test_read_datetime.py`:
- `TestParseIsoformatZDesignator` — directly unit-tests `_parse_isoformat`: `Z` equals
  the `+00:00` form and is tz-aware UTC; lowercase `z`; `bytes`; and no-regression
  guards for explicit offsets and naive strings. Asserting `Z == +00:00` is a
  **cross-version guard** — it fails on Python < 3.11 without the fix and passes on all
  versions with it.
- `TestReadAttributeDatetime.test_attribute_isodatetime_z_designator` — exercises the
  full construct/read path with a stored `b"...Z"` attribute (mirrors the existing
  `+00:00` test above it).

Evidence (isolated conda envs):
- **Python 3.10.20** (the `<3.11` floor): reverting only the source fix, the 5
  `Z`-designator tests FAIL with `ValueError`; the 2 offset/naive no-regression tests
  still PASS. With the fix: **14 passed**; wider `tests/unit/build_tests/` → **379
  passed, 7 skipped, 1 xfailed**.
- **Python 3.12.13**: **14 passed** (native `Z` support; the fix is a no-op there).
- `ruff check` clean on both changed files.

## Reviewer notes / open questions

- **Lowercase `z`.** RFC 3339 §5.6 permits a lowercase `z`; stdlib `fromisoformat`
  accepts *neither* lowercase `z` (any version) *nor* uppercase `Z` before 3.11.
  Normalizing both makes hdmf slightly more lenient than stdlib (Postel's law) at zero
  regression risk (no naive/offset/date string ends in `z`). If you'd rather match
  stdlib exactly and accept only uppercase `Z`, change `('Z', 'z')` → `'Z'` and drop
  `test_lowercase_z` — happy to do so.
- **Test placement.** Tests are in the dedicated `mapper_tests/test_read_datetime.py`
  alongside the existing `+00:00` tests (their natural home). Easy to move.
- **Scope.** The change is confined to `_parse_isoformat`; the `date.fromisoformat`
  branch is untouched, and all datetime read call-sites route through this one helper.
- CHANGELOG entry added under `## HDMF 6.2.0 (Upcoming)` → `### Fixed`.

## Checklist

- [x] Updated `CHANGELOG.md`
- [x] PR describes the problem and the solution
- [x] Reviewed the Contributing Guide
- [x] Uses "Fix #1524" notation

---

*Prepared with AI assistance (Claude Code): the reproduction (on an isolated Python
3.10 environment), fix, and cross-version tests were AI-drafted and independently
reviewed. Reasoning and verification are documented above.*


rly (approved):

rly:
Thanks @Leonard013 ! Looks good to me.

PR #1546, opened 2026-07-22
Fix. Render scalar numpy bool/int attributes inline in HTML repr

h-mayorquin:
Scalar numpy `bool`/`int` attributes (`np.bool_`, `np.int64`) were misrouted to the array-rendering path of the Jupyter HTML repr instead of rendering inline like `float`/`str` scalars (`np.float64` slips through because it subclasses Python `float`). Depending on how the container was obtained this either raised inside the array renderer (`'memoryview' object has no attribute 'dtype'`, shown as `unable to render`) for an in-memory container, or produced a misleading expandable block reporting `Shape: ()` for one read from a file. I fixed it in `_unwrap_scalar` so every numpy scalar (`np.generic`) is unwrapped to a Python scalar and reaches the inline branch, rather than broadening the isinstance check in `container.py`, since that one helper is where the other scalar checks already funnel through.

Before:
image

After
image


rly (approved):
LGTM! Thanks @h-mayorquin !

PR #1549, opened 2026-07-28
Fix indexing an empty DynamicTableRegion selection in memory

h-mayorquin:
## Motivation

Selecting a zero-length region from a `DynamicTableRegion` raised `IndexError: arrays used as indices must be of integer (or boolean) type` when the target table's columns held their data as numpy arrays. This happens for an empty slice and, more commonly, for a ragged region row that references no target rows, which is the shape produced when rows are appended to a table whose `DynamicTableRegion` column has no targets for them. `DynamicTableRegion.get` takes `np.unique` of the selected row indices, and `np.unique` of an empty selection returns a `float64` array, which numpy rejects as an index. A column whose data is a list takes a different branch in `Data.get`, indexing elementwise, and tolerates the float array, which is why the existing `TestDynamicTableRegion` suite, whose fixture is built from lists, stayed green.

The same selection works for files on disk, because the indices come back from HDF5 already integer-typed, so the container behaves differently in memory than it did after a round trip. I think it should work for both cases. I cast the unique row indices to an integer dtype rather than returning early on an empty selection, because region data are row indices: the cast is a no-op when the selection is non-empty, and it fixes the class of problem (a non-integer index array) rather than the one input that produces it.

I found this downstream in AllenNeuralDynamics/aind-ephys-pipeline#121, reported by @alejoe91  @CodyCBakerPhD, where units from a second probe are appended to an NWB units table that already has an `electrodes` region column. Those units are written with an empty region, the file writes and reads back correctly, and only inspecting the in-memory `NWBFile` before writing raised.

## How to test the behavior?
```python
import numpy as np
from hdmf.common import DynamicTable, VectorData

target = DynamicTable(
    name="target", description="t",
    columns=[VectorData(name="x", description="x", data=np.array([1, 2, 3]))],
)
table = DynamicTable(name="table", description="t")
table.add_column(name="region", description="r", index=True, table=target)
table.add_row(region=[0])
table.add_row(region=[])   # a row that references no target rows

table["region"][0]   # one-row DataFrame, before and after
table["region"][1]   # IndexError before, empty DataFrame after
```

## Checklist


rly:
Thanks @h-mayorquin ! I added an extra check to cover the case where the DynamicTableRegion points to a ragged column and this error was raised: `IndexError: index 0 is out of bounds for axis 0 with size 0`

rly (approved):

rly (approved):

Issue #1559, opened 2026-08-13
[Bug]: `DynamicTable.add_row` is quadratic in the number of rows


h-mayorquin:
## What happened?

Writing a 21,330-row `EventsTable` took 124 s, of which about 123.5 s was `is_ragged`. I found it writing that table from neuroconv, where the source is a complete twelve-minute pyPhotometry recording, a 364 KB file whose digital line pulses at 30 Hz. Real sessions run for hours, so this gets worse quickly.

https://github.com/catalystneuro/neuroconv/pull/1920

The reason is a (most likely unintended) check on `is_ragged`.

`add_row` re-derives from the whole table, on every insert, a fact that only the element just appended can change. It calls `is_ragged` on the entire column after appending a single value, and every caller pays it as `check_ragged` defaults to `True`:
https://github.com/hdmf-dev/hdmf/blob/6.1.0/src/hdmf/common/table.py#L847-L852

`enforce_unique_id` has the same shape, though it defaults to `False` and so is only paid by callers who ask for it: `row_id in self.id` iterates the whole `ElementIdentifiers` column as it defines no `__contains__`:
https://github.com/hdmf-dev/hdmf/blob/6.1.0/src/hdmf/common/table.py#L835-L837

So each check costs O(n) at row n and the script below fills a one column table in O(n^2). Note that its `fill` leaves both off unless asked, where hdmf's own default has `check_ragged=True`:

```
  rows  check_ragged  enforce_unique_id  both off
  1000         0.11s              0.18s     0.02s
  2000         0.42s              0.68s     0.03s
  4000         1.63s              2.91s     0.07s
  8000         6.19s             10.63s     0.11s
```

Doubling the rows quadruples the time in the first two columns. However, the last column is flat at about 14 microseconds per row.
So `add_row` itself is linear and the cost is these two checks. This will be even worse with more columns as each one is scanned on each insert.

I think both could be fixed by keeping an incremental check as the previous entries were already scanned:
* For raggedness, holding the column's common element length and an already-ragged flag answers from the appended element alone. The new element still has to be recursed into for nested raggedness, but that costs its own size rather than the table's.
* For the ids, a set answers in O(1), and the work is keeping it true when ids arrive by another route: passed at construction, appended directly through `table.id.append`, or read from a file where `id.data` is an h5py dataset and building the set would materialize the whole column. Caching only when the data is an in-memory list would be the conservative version.

[...]

PR #1561, opened 2026-08-19
Make the ragged check in `DynamicTable.add_row` incremental

h-mayorquin:
## Motivation

In the review of #1066 @rly pointed out that the ragged check iterates over every element and that he did not see a way around it, which is why `check_ragged` was added as an opt-out. This PR is an answer to that question, and to the first half of #1559. The idea is to test raggedness on-line, comparing the new element against the ones already added instead of rescanning the column. This means only asking whether the new element is the one that makes the column ragged. Once a column is marked as ragged the warning is emitted once, as that can only happen once, and the check is skipped for that column from then on. The column is rescanned once if it grew by something other than that single append, so a row added with `check_ragged=False` is still counted.

With this improvement, the benchmark for filling a one-column table with `check_ragged` at its default of `True`, so this is what every caller of `add_row` pays:

| rows | 6.1.0 (dev) | this PR | speedup |
| --- | --- | --- | --- |
| 1000 | 0.109s | 0.022s | 5x |
| 2000 | 0.380s | 0.038s | 10x |
| 4000 | 1.452s | 0.068s | 21x |
| 8000 | 5.647s | 0.132s | 43x |
| 16000 | 22.317s | 0.295s | 76x |

The speedup doubles every time the table doubles, as the check went from linear per row to constant.

The biggest change is that we do not emit the warning on every row any more, which was redundant, but I think that is fine. The message is identical every time and a ragged column never becomes unragged, and the default warning filter was already collapsing the repeats by code location. What do you think? Is there something I am missing?

The `enforce_unique_id` half of the issue is a separate change with no behavior attached to it, so I am sending it as its own PR.

## How to test the behavior?

The timing, which is the script from #1559 with `check_ragged` left at its default:

```python
import time

def fill(num_rows):
    table = DynamicTable(name="bench", description="")
    table.add_column(name="a", description="")
    start = time.perf_counter()
    for _ in range(num_rows):
        table.add_row(a=1.0)
    return time.perf_counter() - start


for num_rows in (1_000, 2_000, 4_000, 8_000, 16_000):
    print(f"{num_rows:6d}  {fill(num_rows):7.3f}s")
```

And the warning, where the third row warns on `dev` and is silent here:

```python

table = DynamicTable(name="table", description="")
table.add_column(name="qux", description="")
table.add_row(qux=[1])
table.add_row(qux=[1, 2])  # warns, this is the row that makes the column ragged
table.add_row(qux=[3, 4])  # warned on dev too, silent now
```

## Checklist

On the last one, the PR references #1559 but on purpose without the `Fix` keyword, as the issue covers two checks and this PR fixes one of them. It should stay open until the `enforce_unique_id` PR lands.


rly:
This looks great. I made a couple minor updates to docs. Thank you @h-mayorquin !

rly (approved):

PR #1567, opened 2026-08-29
Return an empty DataFrame from `HERD.to_dataframe` when there are no references

rly:
## Motivation

`HERD.to_dataframe()` raises `ValueError: No objects to concatenate` on a HERD that holds no references:

```python
from hdmf.common.resources import HERD
HERD().to_dataframe()
# ValueError: No objects to concatenate
```

Step 3 of `to_dataframe` builds one merged frame per row of the `object_keys` table and hands the list to `pd.concat`. With no rows, that list is empty and `pd.concat` raises.

An empty HERD is not an exotic state. It is what you get before adding any reference, and it is also what `HERD.add_ref_termset` leaves behind when none of the terms are found in the `TermSet`: the terms come back in the returned `missing_terms` dict, nothing is added, and the natural next step of inspecting the HERD then raises. The repr methods already worked around this with an explicit emptiness check in `__flattened_dataframe_or_none`.

## How to test the behavior?

`to_dataframe` now returns an empty `DataFrame` carrying the same columns as a populated one, in the same order, with the four index columns still typed `uint32`. `use_categories=True` still produces the two-level `MultiIndex`.

Two tests in `tests/unit/common/test_resources.py`, `test_to_dataframe_empty` and `test_to_dataframe_empty_use_categories`, compare an empty HERD's frame against one built from a populated HERD, so they fail if the columns drift apart. Both fail with `ValueError: No objects to concatenate` without the fix.

The emptiness check in `__flattened_dataframe_or_none` is gone, since `to_dataframe` handles the case itself. The helper keeps its `try`/`except` for a HERD whose backing file is closed.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 6.2.1 (Upcoming)
+
+### Fixed
+- Fixed `HERD.to_dataframe` raising `ValueError: No objects to concatenate` on a `HERD` that holds no references. It returns an empty `DataFrame` with the usual columns. A `HERD` is in this state before any reference is added, and also after `HERD.add_ref_termset` finds none of its terms in the `TermSet` and returns them all as `missing_terms`. @rly [#1567](https://github.com/hdmf-dev/hdmf/pull/1567)
+
 ## HDMF 6.2.0 (August 19, 2026)
 
 ### Documentation and tutorial enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## HDMF 6.2.1 (Upcoming)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # HDMF Changelog
 
+## HDMF 6.2.1 (Upcoming)
+
+### Fixed
+- Fixed `HERD.to_dataframe` raising `ValueError: No objects to concatenate` on a `HERD` that holds no references. It returns an empty `DataFrame` with the usual columns. A `HERD` is in this state before any reference is added, and also after `HERD.add_ref_termset` finds none of its terms in the `TermSet` and returns them all as `missing_terms`. @rly [#1567](https://github.com/hdmf-dev/hdmf/pull/1567)
```
```suggestion
- Fixed `HERD.to_dataframe` raising `ValueError` on a `HERD` that holds no references. It returns an empty `DataFrame` with the usual columns. @rly [#1567](https://github.com/hdmf-dev/hdmf/pull/1567)
```

oruebel (approved):

PR #1576, opened 2026-09-02
Support numpy variable-length strings (`StringDType`) in `convert_dtype`

h-mayorquin:
This comes out of the Zarr v3 migration in hdmf-zarr, [#325](https://github.com/hdmf-dev/hdmf-zarr/pull/325). Zarr v3 has a real `string` data type, so every string column read from a v3 store comes back as a numpy array of kind `T`, and hdmf does not recognize that anywhere in the dtype machinery. `convert_dtype` throws a `TypeError` for a `text` spec, or silently infers `str` when the spec has no dtype and the write then fails in the backend. 

hdmf-zarr works around it in [#366](https://github.com/hdmf-dev/hdmf-zarr/pull/366) by overriding `dtype` on a `zarr.Array` subclass so the array reports `object` while it is really holding a `StringDType`, which means `array.dtype` stops answering what the array is. This PR treats kind `T` as `utf8` and hands the array back uncoerced, the same as an object array of strings, so the override is not needed anymore.


rly:
Thanks @h-mayorquin ! I confirmed the bug and made two small follow-ups.

Array-valued attributes were still broken. `convert_dtype` is used for `AttributeSpec` too, and it was handing h5py something it can't take, since `h5tools.py:917` converted only `kind == 'U'` arrays to `H5_TEXT`. I updated that to `kind in ('U', 'T')`.

I also dropped `assertEqual(ret.dtype.kind, 'T')` from the two new `convert_dtype` tests. It followed `assertIs(ret, value)` on an array the test constructed with `StringDType()`, so it mirrored the setup line and couldn't fail.

**Relatedly: `ascii` / `bytes` / `isodatetime` specs still raise.** `astype('S')` at `objectmapper.py:304` and `:307` fails on `StringDType` the same way `astype('U')` did. hdmf-zarr with zarr v3 stores ascii datasets as `StringDType` arrays too.

It doesn't bite today, since `HDMFZarrArray` subclasses `zarr.Array` and `convert_dtype` takes the `is_zarr_array` branch, which passes `ascii` through untouched. But it bites once the array is materialized to numpy: an export, a re-write, or a `VectorData` built from a slice. Routing kind `'T'` into the `_ascii` and `_isoformat` branches is two lines; `np.char.encode(value, 'utf-8')` works where `astype('S')` doesn't. What do you think about that change?

h-mayorquin:
This makes sense.

Yes to `ascii` / `bytes` / `isodatetime`, but with `'ascii'` as the codec rather than `'utf-8'`. Otherwise things like "café" or "desiré" get written as UTF-8 bytes into a field the schema declares as `ascii`, so anything reading them back using that assumption fails. Using `'ascii'` gives a loud error at entry instead. I mean the following:

```python
ret = np.char.encode(value, 'ascii') if value.dtype.kind == 'T' else value.astype('S')
```

oruebel:
@h-mayorquin @rly I believe the following areas also would need to be updated

## 1. Update `validator._get_type_from_dtype_attr`

I have not tested this, but I believe when HDMF validates an array with `np.dtypes.StringDType()` (`kind == 'T'`), it falls through the type-check in `_get_type_from_dtype_attr` and returns the raw `StringDType()` class, which should cause schema validation against `'utf8'` or `'ascii'` to fail.

**Proposed Change in `src/hdmf/validate/validator.py`:**
```python
    # Handle object dtype (zarr style variable-length strings) or numpy StringDType
    if data.dtype.kind in ('O', 'T'):
        if _get_length(data) > 0:
            return get_type(data[0], builder_dtype)
        return "utf", None
```

## 2. Update `h5tools._check_str_dtype`

When reading datasets from HDF5, `DatasetBuilder._check_str_dtype` determines if the dataset should be wrapped in `StrDataset`. It currently explicitly filters for `kind == 'O'`. It should be updated to recognize `'T'` to handle variable-length string unpacking consistently.

**Proposed Change in `src/hdmf/backends/hdf5/h5tools.py`:**
```python
    def _check_str_dtype(self, h5obj):
        dtype = h5obj.dtype
        if dtype.kind in ('O', 'T'):
            if dtype.metadata and dtype.metadata.get('vlen') is str and H5PY_3:
                return StrDataset(h5obj, None)
        return h5obj
```

## 3. Verify Other `kind == 'O'` Checks in H5Tools

There are a few other places in `src/hdmf/backends/hdf5/h5tools.py` that explicitly check for `kind == 'O'`. We may want to review if they also need to support `'T'`:

* **In `__compound_dtype_to_list`:**
  ```python
      if h5obj.dtype.kind in ('O', 'T') and len(h5obj) > 0:
  ```
* **In `__h5_to_builder_dtype`:**
  ```python
      if dtype.kind in ('O', 'T'):
  ```

h-mayorquin:
@rly the `ascii` and `isodatetime` branches are done in the last commits with `'ascii'` as the codec.

@oruebel Thanks for going through the rest of the dtype machinery.

1) I tested it and it fails the way you describe, done in the last commit. It only fires for `zarr.Array` and `h5py.Dataset` though, numpy arrays take the `isinstance(data, np.ndarray)` branch of `get_type` and already resolve to `utf`. It does belong here as hdmf-zarr arrays report kind `T` to the validator once the `dtype` override is dropped (thanks for catching it, I had only looked at the build side!).

2) and 3) I think that these only make sense once the read path in #1578 is enabled:

* h5py doesn't return kind `T` unless we ask for it with `dataset.astype(np.dtypes.StringDType())` so right now no file reaches those branches.
* `_check_str_dtype` is only reached from the `kind == 'O'` gate at `h5tools.py:641` so widening it alone changes nothing. I couldn't find `__h5_to_builder_dtype`, I think you mean that same place.
* Once we do ask for `StringDType`, `DatasetBuilder` rejects the `AsTypeView` that h5py returns and an `ascii` dataset reports `utf` which fails validation.

I can take care of this and #1578 after this is merged.

oruebel:
> I can take care of this and #1578 after this is merged.

Sounds good. 

> 2\. and 3) I think that these only make sense once the read path in [Read string data as a string dtype instead of `object` #1578](https://github.com/hdmf-dev/hdmf/issues/1578) is enabled:

Agreed, since these are HDF5-read-specific, those items can be addressed there.

oruebel (approved):
Looks good to me. I'll let @rly handle final approval and merge.

rly:
I clarified some comments, added a simple test, and simplified the changelog entry. LGTM!

PR #1581, opened 2026-09-09
Materialize the elements array before mapping enum terms to their index

h-mayorquin:
This comes out of the zarr v3 migration in hdmf-zarr, https://github.com/hdmf-dev/hdmf-zarr/pull/325. Zarr v3 follows the array API and `array[0]` on a 1-d array returns a 0-d array instead of a scalar. `_map_elements` uses those elements as dict keys so any `EnumData` read from a zarr v3 store throws `unhashable type: 'numpy.ndarray'`. hdmf-zarr covers this right now by unwrapping 0-d results in `HDMFZarrArray.__getitem__` and this is one of the two places that need that shim, see https://github.com/hdmf-dev/hdmf-zarr/pull/366.

This PR slices the elements once with `elements[:]` and iterates the result. A full slice comes back as a numpy array with plain scalars on every backend, and the terms are read in one call instead of one per element.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - Fixed `HERD.to_dataframe` raising `ValueError` on a `HERD` that holds no references. It returns an empty `DataFrame` with the usual columns. @rly [#1567](https://github.com/hdmf-dev/hdmf/pull/1567)
 - Fixed the windows-python3.14-ros3 job failing on Windows by working around the HDF5 ros3 shutdown deadlock. @rly [#1572](https://github.com/hdmf-dev/hdmf/pull/1572)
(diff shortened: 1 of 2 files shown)


oruebel (approved):

rly:
Interesting: [NumPy deliberately](https://numpy.org/neps/nep-0056-array-api-main-namespace.html#parts-of-the-standard-that-are-not-adopted) chose to return array scalars rather than 0-D arrays in most cases where the array API standard, and other array libraries, return 0-D arrays, because it would be a large, costly effort with a big impact.

rly (approved):

rly:
LGTM. EnumData is also rarely (if ever) used, and the `elements` field is intended to be small, so materializing it all should be fine. We can address any performance issues if/when they come up.

